PR #22, opened 2025-07-19
jir-dummy-permissions

jirhiker:
WIP: Authentication/Permissions

### DIFF
--- a/api/thing.py
+++ b/api/thing.py
@@ -215,7 +215,7 @@ def create_thing_id_link(link_data: CreateThingIdLink, session: session_dependen
 def create_well(
     well_data: CreateWell,
     session: Session = Depends(get_db_session),
-    user = well_user_dependency
+    user=well_user_dependency,
 ) -> WellResponse:
     """
     Create a new well in the database.
--- a/api/thing.py
+++ b/api/thing.py
@@ -13,11 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
-from typing import List, Annotated, Union
-from shapely.geometry import mapping
-from shapely import wkb
-from fastapi import APIRouter, Depends, Query
+from fastapi import APIRouter, Depends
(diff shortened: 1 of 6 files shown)


api/thing.py
```diff
@@ -217,11 +221,14 @@ def create_thing_id_link(link_data: CreateThingIdLink, session: session_dependen
     status_code=status.HTTP_201_CREATED,
 )
 def create_well(
-    well_data: CreateWell, session: Session = Depends(get_db_session)
+    well_data: CreateWell,
+    session: Session = Depends(get_db_session),
+    # user = well_user_dependency
```
jacob-a-brown: How come the user doesn't need to be authenticated to create a new well at this POST endpoint?

api/thing.py
```diff
@@ -23,7 +23,11 @@
 from starlette import status
 
 from api.pagination import CustomPage
-from core.dependencies import session_dependency
+from core.dependencies import (
+    session_dependency,
+    well_user_function,
```
chasetmartin: Is well_user_function a necessary import? Is it needed because well_user_dependency depends on it when it's defined in dependencies.py?

chasetmartin (approved):
Looks good to me for starting to test auth

api/thing.py
jirhiker: commented out for testing. We need to protect most of the post endpoints with user dependencies

api/thing.py
jirhiker: not a necessary import.  will be removed

jacob-a-brown: The `black` linter can be used to find (and subsequently remove) unused imports. This can be set up in both GitHub Actions and pre-commit hooks. My dev branch has `.pre-commit-config.yaml` set up - it'll be available after I finish with Docker and open a PR.

api/thing.py
jacob-a-brown: Should this be left in for security and consistency? You are overriding the dependency in the `thing` tests so it should still run correctly.

PR #23, opened 2025-07-20
jir-contact-thing-service

jirhiker:
⚠️ Make sure to look into merging PR #28 

added patch endpoints for contact and thing.

```
thing/{thing_id}
thing/{thing_id}/location
thing/{thing_id}/well or thing/well/{well_id}

/contact/address/{address_id}
/contact/phone/{phone_id}
/contact/email/{email_id}
/contact/{contact_id}
```


api/thing.py
```diff
@@ -107,9 +120,12 @@ def get_things(
 
 @router.get("/well", summary="Get all wells")
 async def get_wells(
+    session: session_dependency,
     # api_id: str = None,
     # ose_pod_id: str = None,
-    session: session_dependency,
+    sort: str = None,
+    order: str = None,
+    filter_: Annotated[str, Field(alias="filter")] = None,
```
jacob-a-brown, line 136: Will this turn into an enums of possible filters? Or will that be dependent on the model(s) being filtered? Because their columns will change.

db/contact.py
```diff
@@ -80,8 +81,8 @@ class Address(Base, AutoBaseMixin):
     city = Column(String(100), nullable=False)
     state = Column(String(50), nullable=False)
     postal_code = Column(String(20), nullable=False)
-    country = Column(String(100), nullable=True)
-    address_type = lexicon_term(nullable=True)
+    country = Column(String(100), nullable=False, default="US")
```
jacob-a-brown: US or USA?

schemas_v2/contact.py
```diff
@@ -135,3 +138,104 @@ class CreateContact(BaseModel):
...
+
+class ContactResponse(ORMBaseModel):
+    """
+    Response schema for contact details.
+    """
+
+    things: List[ThingResponse] = []  # List of related things
+    id: int
+    name: str
+    role: str
```
jacob-a-brown, line 188: eventually this will be an enum?

jacob-a-brown: or keep that info in the database table?

schemas_v2/location.py
```diff
@@ -15,10 +15,40 @@
...
 
 
+# -------- CREATE ----------
+class CreateLocation(BaseModel):
+    """
+    Schema for creating a sample location.
+    """
+
+    notes: str | None = None
+    point: str = "POINT(0 0)"  # Default to a point at the origin
```
jacob-a-brown: should this be required?

schemas_v2/location.py
```diff
@@ -15,10 +15,40 @@
...
+    release_status: str | None = "draft"
+
+
+class CreateGroup(BaseModel):
+    """
+    Schema for creating a group.
+    """
+
+    name: str
+
```
jacob-a-brown, line 53: how about a `description` field?

jacob-a-brown (commented):
Should `POST` and `PATCH` endpoints have rollbacks for security? We can put these into a `try/except` statement like

```
try:
    <create object with info>
    <flush object>
    db.commit()
except Exception as e:
    db.rollback()
    raise HTTPException(status_code=500, detail=str(e))
finally:
    db.close()
```

In the case of an asset, where a file is being uploaded somewhere, it should be deleted in the `except` clause. Since this is a POC, though, if we move forward with FastAPI and SQL Alchemy we can create a ticket to backfill these implementations.

schemas_v2/contact.py
jirhiker: we need to have a discussion and develop an ADR about where our lexicon lives.

db/contact.py
jirhiker: I don't have a strong opinion. Probably should be a `lexicon_term` and be spelled out as `United States`

jirhiker:
> Should `POST` and `PATCH` endpoints have rollbacks for security? We can put these into a `try/except` statement like
> 
> ```
> try:
>     <create object with info>
>     <flush object>
>     db.commit()
> except Exception as e:
>     db.rollback()
>     raise HTTPException(status_code=500, detail=str(e))
> finally:
>     db.close()
> ```
> 
> In the case of an asset, where a file is being uploaded somewhere, it should be deleted in the `except` clause. Since this is a POC, though, if we move forward with FastAPI and SQL Alchemy we can create a ticket to backfill these implementations.

Im not sure this is necessary and potential can obsure other issues, but if we encounter a need for such a guard we can add it to the backlog.

api/thing.py
jirhiker: sort, order, and filter_ are to service the front end client's data table requirements.  

filter_ will be a dict like `filter={"field":"id","value":1,"operator":"eq"}`

schemas_v2/location.py
```diff
@@ -27,7 +27,7 @@ class CreateLocation(BaseModel):
     """
 
     notes: str | None = None
-    point: str = "POINT(0 0)"  # Default to a point at the origin
+    point: str  # point is required and should be in WKT format
```
jacob-a-brown, line 31: We can add `@field_validators` to ensure WKT. Google's AI suggested using shapely for this - something like

```
from shapely import wkt

@field_validator("point")
def validate_point_is_wkt(cls, point):
    try:
        geometry = wkt.loads(value)
        if not geometry.is_valid:
            raise ValueError("WKT geometry is not topologically valid")
        return value
    except Exception as e:
        raise ValueError(f"Invalid WKT geometry: {e}")
```

validators could also be used to ensure that the geometry is a point.

jirhiker: done

jacob-a-brown (approved):

PR #28, opened 2025-07-21
jir-schema-refactor

jirhiker:


chasetmartin:
@jirhiker My understanding is that this refactors the schema into single files for each model type (like contact, thing) that contains all associated schemas for response, create, update, etc?

If so that change of organization makes sense to me - I like having all pydantic schemas in one file for each model.

api/contact.py
```diff
@@ -28,15 +24,12 @@
 from core.dependencies import session_dependency
 from db import ThingContactAssociation, Thing
 from db.contact import Contact, Email, Phone, Address
-from db.engine import get_db_session
-from schemas.response.contact import (
+from schemas_v2.contact import (
+    CreateContact,
```
jacob-a-brown, line 32: If all response schemas are named `<Object>Response` (e.g. `ContactResponse`) should the create schemas follow a similar pattern for style repetition? Such as `<Object>Create` (e.g. `ContactCreate`). Or, is `Response` appended and `Create` prepended because that makes more sense when talking?

schemas_v2/observation.py
jacob-a-brown, line 1: We spoke about having each observation having their id/pk column called `observation_id`. In light of this, and inheritance, should it be

```
class ObservationResponse(BaseModel):
    observation_id: int
    observation_timestamp: datetime

class GroundWaterLevelObservationResponse(ObservationResponse):
    groundwater_level_observation_id: int
    ...


class GeothermalObservationResponse(ObservationResponse):
    geothermal_obseration_id: int
    ...
```
?

schemas_v2/thing.py
jacob-a-brown, line 1: `<Station>Response`s (e.g. `WellResponse`) should inherit from `ThingResponse` so that all attributes are also returned

jacob-a-brown: Similarly, there can be a `CreateThing` class from which all `Create<Station>` classes inherit to simplify code and its maintenance

jacob-a-brown:
I also concur with @chasetmartin's comment. I like having pydantic schemas organized by model

schemas_v2/thing.py
```diff
@@ -0,0 +1,220 @@
...
+    # ose_pod_id: str | None = None
+    # usgs_id: str | None = None
+    well_type: str | None = None
+    well_depth: float | None = None  # in feet
+    hole_depth: float | None = None  # in feet
+    construction_notes: str | None = None
+    # Additional fields can be added as needed
+
+
+class LocationWellResponse(LocationResponse):
```
jacob-a-brown, line 115: Would the location be part of everything `Thing`'s `geometry` field? If so, `ThingReponse` can have a `geometry` field and then inheritance can be used to streamline fields across all things/types. Or, if we want to proceed with listing things at locations, that can be abstracted into a `LocationThingResponse` and then inheritance can be used for all thing types.

api/contact.py
jirhiker: appended and prepended for ease of talking, but I've also been wondering  if <Object>Create, <Object>Response is better

schemas_v2/observation.py
jirhiker: I think every Model should have a pk called `id` not `modelname_id`

schemas_v2/thing.py
jirhiker: The WellResponse does not inherit from ThingResponse in the same way a WellThing does not inherit from a Thing.  A WellThing is associated with a Thing, and they have different attributes. 

the Thing could be returned with a WellThing e.g.

```
class ThingResponse(BaseModel):
      name: str
      location_id: int

class WellResponse(BaseModel):
     well_depth: float | None
     ...
      thing: ThingResponse
     ...

File: schemas_v2/thing.py (hunk 2 above)
Comment (jirhiker) [reply schemas_v2/thing.py:115]: Im thinking the latter. 

The location api, where LocationWellResponse is used needs cleanup/deprecation, some of its functionality I anticipate to be replaced by the geospatial router

File: schemas_v2/observation.py
Comment (jacob-a-brown) [reply schemas_v2/observation.py:1]: Having every pk be called `id` makes it easier to standardize across all models.

I also think that inheritance can be used for shared attributes in different types of observations. Perhaps they could inherit from the parent (`ObservationResponse`) and a Mixin of shared attributes like

```
ObservationResponseMixin(BaseModel):
     observation_id: int
     observation_timestamp: datetime
```

chasetmartin: I know we have diverging thoughts on this matter on our team, but I'll second the opinion that pk should be `id` with no qualifiers for every model.

ksmuczynski: I was on the fence about pk naming conventions, but I agree that pk should be `id` with no qualifiers. My initial hesitancy comes from my experience working directly with SQL, but moving forward with an ORM, `id` appears to be the standard and most accepted approach.

manage.py
jacob-a-brown, line 1: Are `manage.py` and `migration/migration2.py` for migrating from NM_Aquifer to the new system? The nomenclature is similar to alembic (that is, "migrations")

jirhiker: manage.py is supposed to be similar to django's manage.py,  a utility module to manage the project.  
migration2.py is for uploading staging data to the new system, its not intended to mimic alembic in a migration versioning sense.  I didn't like migration.py so I rewrote it as migration2.py. the data migration scripts will need cleanup and improvement

jacob-a-brown:
I can't approve it because there are some merge conflicts, but besides that it looks good to me

PR #29, opened 2025-07-21
BDMS-20: Update CD Action - UV Pip Freeze and GAE Delete Oldest Version

chasetmartin:
This PR:
 - Updates the staging deploy action to use uv pip freeze to generate the necessary requirements.txt file. I believe these steps will take care of this.
 - Adds a step for listing the app engine versions, finding the oldest one, then deleting it (after deployment of the new version)
 - Removes the --no-promote flag so traffic is immediately directed to deployed version


chasetmartin:
@jirhiker Alright I added in the cleanup oldest version step to this same PR

jirhiker (approved):
Ill approve this pull request but I have a question about creating and deleting the app.yaml file.

What about using environment variables in the app.yaml file and then adding an app.yaml file to the repo? (instead of creating it in the GitHub Action yaml file)

you still can use the github secrets, just write the secret to the env var in the first step of the job

lastly is there a need to delete the app.yaml file? Is that recommended practice for any files created on a github action worker VM?

chasetmartin:
@jirhiker Yeah I think we can do that, as long as we don't mind the app.yaml file living in the repo. I see nothing wrong with this and it will simplify the entire action.

PR #35, opened 2025-07-23
fix: update contact model with alembic

jacob-a-brown:


ksmuczynski (approved):

chasetmartin (approved):

jirhiker (approved):

jirhiker:
Should we consider an automated process for running the alembic migrations on the production database?

jacob-a-brown:
That sounds like a good idea to me. Would tests be one-offs for a migration? That could also be a good way to track schema changes. The tests would have to be performed against databases with actual data to see if anything break or any errors are raised (production data in a testing environment? by using Docker images we could quickly spin up a database and then populate it with the existing production data and run tests/migrations).

I had some test data in my database and when I applied the migration it ran into an error, which elucidated this testing need to me (easily taken care of locally by deleting and remaking the database locally, but that won't always be the case because we won't want to spin down a production database).

PR #36, opened 2025-07-23
jir-fix-thing-api

jirhiker:


jacob-a-brown (commented):
This pertains to discussions later, but if there's a one-to-one correspondence between a type and a thing (e.g. a well and a thing) why have the extra type in the path (e.g. `base` or `well`)? Why not just retrieve the `thing` and all of its associated properties by getting the `thing` with `thing_id`?

PR #37, opened 2025-07-24
JIR simplifying the thing concept.

jirhiker:


jacob-a-brown (approved):
This looks good to me. Something that crossed my mind last night was that we may not actually need documentation in the database, such as prefixes, to indicate which columns correspond with which thing type. Since we intend to interact with the database only through the API (whether through endpoints or admin panels), we can have that information reside in the schemas. When a user wants a well thing, the pydantic schema defines which columns are relevant and returns them (ignoring the others).

chasetmartin (approved):
This looks good and I'm excited to try it more. I am having trouble running this locally because of alembic migrations, so I will try to sort that out on my end.

PR #42, opened 2025-07-27
JIR: asset service

jirhiker:


api/group.py
```diff
@@ -59,11 +59,17 @@ def create_group_thing(
 
 # ============= Get =============================================
 @router.get("", summary="Get groups")
-async def get_groups(session: session_dependency) -> CustomPage[GroupResponse]:
+async def get_groups(
+    session: session_dependency,
+    filter_: str = Query(
+        ...,
```
chasetmartin: I'm getting 5 failed tests from the test_group and test_thing files, and it looks like they are failing because the tests do not pass a filter when I think this code makes the filter required? I was able to fix the failing tests by adding default=None:
```
filter_: str = Query(
        default=None,
        alias="filter",
    ),
```

api/thing.py
```diff
@@ -79,12 +79,14 @@ def get_things(
     query: str = None,
     sort: str = None,
     order: str = None,
-    filter_: Annotated[str, Field(alias="filter")] = None,
+    filter_: str = Query(
+        ...,
```
chasetmartin: Same thing here?
```
filter_: str = Query(
        default=None,
        alias="filter",
    ),
```

chasetmartin (changes requested):
I just have the two questions about whether the filters should be required. Otherwise I'm excited to try this to use GCP for asset storage.

chasetmartin (approved):
The filter default is working for me now.

alembic/versions/59763cc319cf_initial_migration.py
```diff
@@ -0,0 +1,45 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+from db import *  # Import your Base from models/__init__.py
+from db.engine import engine
+
+configure_mappers()
+
+Base.metadata.drop_all(engine)
+Base.metadata.create_all(engine)
```
jacob-a-brown: this pretty much takes care of creating and dropping the tables, but doing this hides what is happening behind the scenes. that is, this doesn't really show which tables are being created and how they are being created, so we aren't getting records/paper trails of how the database is initially being constructed. Should we proceed with this method?

jacob-a-brown: If we are going to proceed with this until we settle on a schema, we can put `drop_all` in downgrade and `create_all` in upgrade

jirhiker: I see your point about the paper trail. 

Lets add all the op.create_table commands to the script but comment them out

PR #48, opened 2025-07-28
BDMS-28:  Groundwater Level entry form

jirhiker:
This updates the api with a few API endpoints and query parameters. It is a relatively minor update.  

Although adding a gwl observation is now functional via DM, some discussion is probably required regarding "Series"


tests/__init__.py
```diff
@@ -33,15 +36,64 @@
...
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
+def sample():
+    with session_ctx() as session:
+        loc = Location(point="SRID=4326;POINT(0 0)")
+        session.add(loc)
+        session.commit()
+
+        thing = add_thing(
```
jacob-a-brown: instead of creating a `thing` here you can make the thing fixture an argument in your sample fixture. the thing fixture will then be run before the sample fixture, which also shows how sample is dependent on thing.

jacob-a-brown: the attributes of the `thing` created by the fixture can still be retrieved, such as when `thing_id` needs to be populated.

jacob-a-brown (commented):
these look good to me. i made some refactor/todo comments in my development branch about putting fixtures in one location, which is being done here in `tests/__init__.py`. I'll remove those comments.

an update i think that should occur is that records should be deleted in the fixtures so that tests are independent of each other.

in the sample fixture, for example, after `yield sample` and `session.close()` one can call

```
session.query(Sample).delete()
```

jacob-a-brown (commented):
Do you think it would be worth setting `autouse=True` so we don't have to worry about the order in which the tests are run? If I'm understanding it correctly, this will just instantiate the fixture(s) for the full session and then they can be invoked where needed.

In the future we could also implement `conftest.py` to instantiate and share those fixtures across the testing module and we wouldn't need to worry about imports (though do imports make it easier to understand?).

https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files

jirhiker:
Test improvements are good things for the backlog.  In general, I think any improvements to tests should be out of necessity and in service to adding more functionality to the API

jacob-a-brown (approved):

PR #49, opened 2025-07-31
BDMS 22+: implement sample model | some pytest cleanup | pre-commit hooks

jacob-a-brown:
### **Why**
- It implements the current version of the sample model, with some anticipated changes coming down the road
- Updates pytest session fixtures to be in `tests/conftest.py` for repeated use and retrieval throughout the testing suite.
- Configures pre-commit hooks

### **How**
- get, post, and patch endpoints were made for the `/sample` router
- tests were made for all endpoints, as well as retrieving a single sample by its id
- pytest fixtures were moved to `tests/conftest.py` for use throughout the testing suites
- pre-commit hooks were configured in .pre-commit-config.yaml

### **Notes**
- The sample model is still being developed, but the work done here is applicable to future development so I wanted to get it into `pre-production` (the fixtures in particular)
- Notes about alembic migrations were taken in the initial migration script for when alembic and its migrations are adopted
- refactoring notes were taken and are found by searching for `REFACTOR NOTE/TODO` for future ideas/work
- Some work was done to make tests independent of each other, but that should be a task for another PR. It was only implemented a bit here.


jirhiker:
Alembic migrations do not need to be used for the tests. This seems to be a continuing source of complexity and is not necessary. These unit tests are not for testing database migrations, they are for testing that the API functions correctly.

jacob-a-brown:
Once they are fully implemented, though, shouldn't they be a part of the tests (or there would be a separate testing suite)? To ensure they are working as we expect because if the migrations don't work then the database won't be correctly configured.

api/sample.py
```diff
@@ -69,6 +68,33 @@ def add_sample(sample_data: CreateSample, session: Session = Depends(get_db_sess
 #     return adder(session, GeothermalSample, sample_data)
 
 
+# ============= Update =============================================
+@router.patch("/{sample_id}", summary="Update Sample")
+def update_sample(
+    sample_id: int,
+    sample_data: UpdateSample,
+    session: Session = Depends(get_db_session),
```
jirhiker, line 76: use the existing `session_dependency` dependency

jacob-a-brown:
That makes sense that these are API tests. Would we have a database testing suite, though, to ensure that it's all built and working as expected?

jirhiker (changes requested):
The sample api looks good and the use of fixtures in the tests is an improvement

But I do not like the use of alembic in the tests. It seems unnecessary and adds complexity that is a distraction from the true purpose of the tests

jacob-a-brown:
I'm now using `Base.metadata.drop_all(engine)` and `Base.metadata.create_all(engine)` for the tests since these are just API, not database, tests.

api/sample.py
jacob-a-brown: `session_dependency` is now used

jirhiker (approved):

PR #53, opened 2025-08-01
fix: fixed init_lexicon

jirhiker:
### **Why**
This PR fixes initialization of the lexicon from the lexicon.json. There was an issue when adding a term to multiple categories. 

### **How**
- fixed `init_lexicon` to either add or use and existing term. If the term exists it still will try to add to a category

### **Notes**
N/A


chasetmartin (approved):
It's working for me! I'll push my address_type changes to data manager.

PR #59, opened 2025-08-04
NO TICKET: organize API endpoints and documentation

jacob-a-brown:
### **Why**
- To make the OpenAPI documentation easier to understand the routers should be in their own well-defined sections.
- It should also be easy to scroll through the documentation and find the router/collection of endpoints needed

### **How**
- `lexicon` and `sample` were missing tags that have now been added
- Order the routers alphabetically for easier access

### DIFF
--- a/main.py
+++ b/main.py
@@ -34,29 +34,24 @@ from api.asset import router as asset_router
 from api.search import router as search_router
 from api.geospatial import router as geospatial_router
 
-
-app.include_router(contact_router)
-app.include_router(group_router)
-
-app.include_router(location_router)
-app.include_router(thing_router)
-app.include_router(sensor_router)
-# app.include_router(series_router)
-app.include_router(sample_router)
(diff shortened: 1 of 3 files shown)


jacob-a-brown:
I noticed that tags were missing from `lexicon` and `sample` as I was reviewing the sample endpoints/router and looked at the `/docs` of the hosted API (locally). This is a short and quick fix.

main.py
```diff
@@ -34,29 +34,24 @@
...
-app.include_router(asset_router)
+app.include_router(sample_router)
+app.include_router(sensor_router)
+app.include_router(search_router)
+app.include_router(thing_router)
 
-from admin.user import *
-from admin.base import *
+from admin.user import *  # noqa: F401, F403
+from admin.base import *  # noqa: F401, F403
```
jirhiker, line 54:
```

jirhiker (approved):

PR #61, opened 2025-08-04
fix: added optional thing_id to the BaseAsset schema. fixed sample tests to use singleton asserts instead of asserting equal dictionaries

jirhiker:
### **Why**
- Associating an asset with a thing
- Sample tests

### **How**
- updated the Asset schema to have optional thing_id field
- changed sample tests to use singleton asserts

### **Notes**
- Use bullet points here

### DIFF
--- a/db/sample.py
+++ b/db/sample.py
@@ -45,7 +45,7 @@ class Sample(Base, AutoBaseMixin, ReleaseMixin):
     )
     sensor_id: Mapped[Optional[int]] = mapped_column(
         ForeignKey("sensor.id"),
-        unique=True,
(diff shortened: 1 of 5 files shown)


chasetmartin (approved):
That's a much more graceful way to handle the create and update asset schemas.

jacob-a-brown:
This looks good. @ksmuczynski and I worked on updating the sample database models, schemas, and tests as well. After this gets merged into pre-production I'll resolve merge conflicts and open a PR for the sample endpoint.

jacob-a-brown (approved):

PR #62, opened 2025-08-05
BDMS 22: Implement the sample model in SQLAlchemy POC so that a set of staging data can be displayed in data manager

jacob-a-brown:
### **Why**
- A new sample model was implemented, and then updated after the initial implementation.

### **How**
- Updates to the sample model, mostly for non-nullable fields
- Updates to schemas to include the new fields, restrictions, and validations
- Updates to the sample fixture and tests for the new model and schemas

### **Notes**
- Since PATCH and POST endpoints require the same field validators, a base validator pydantic class is created from which CreateSample and UpdateSample inherit.
- There are many shared fields throughout, but some of them are required for CreateSample and are optional for UpdateSample. Should we try and find some clever to inherit and set fields to optional or not?
- There are some development notes in here for future selves
- There are some questions that have yet to be answered and decided upon. These are noted in REFACTOR TODO notes throughout the code base
- per PR #61 changed sample tests to use singleton asserts


schemas_v2/sample.py
```diff
@@ -16,80 +16,195 @@
...
+"""
+
+
+# -------- VALIDATE ----------
+class ValidateSample(BaseModel):
+    """
+    Validator for Sample data for Create and Update schemas.
+    """
+
+    @field_validator("field_sample_id", check_fields=False)
```
jirhiker: I'm wondering if there is a need for these type of data validations. The data model ensures the uniqueness of this field

db/sample.py
```diff
@@ -45,28 +43,32 @@ class Sample(Base, AutoBaseMixin, ReleaseMixin):
...
     # Sample Attributes
     sample_date: Mapped[datetime.datetime] = mapped_column(
-        DateTime, comment="Date and time of sample collection."
+        DateTime, nullable=False, comment="Date and time of sample collection."
     )
+    # REFACTOR TODO: update with enum/restricted values
     sample_matrix: Mapped[Optional[str]] = mapped_column(
         comment="The material of the sample (e.g., 'gw', 'soil')."
     )
+    # REFACTOR TODO: update with enum/restricted values
```
jirhiker, line 57: `lexicon_term` good for this purpose

schemas_v2/sample.py
```diff
@@ -16,80 +16,195 @@
...
+                    .filter_by(field_sample_id=field_sample_id)
+                    .first()
+                )
+                if existing_sample:
+                    raise ValueError(
+                        f"Field sample ID {field_sample_id} already exists."
+                    )
+        return field_sample_id
+
+    @field_validator("sensor_id", check_fields=False)
```
jirhiker: Same here. Foreign key ensures sensor_id exists. Why test when you can handle a database exception

schemas_v2/sample.py
```diff
@@ -16,80 +16,195 @@
...
+    qc_sample: str = "Original"
 
     sensor_id: int | None = None
-    sample_date: datetime | None = None
-    sample_matrix: str | None = None
-    sample_method: str | None = None
-    sampler_name: str | None = None
-    qc_sample: str | None = None
-    duplicate_sample_number: int | None = None
+    sample_matrix: str | None = (
```
jirhiker, line 88: If the lexicon is used then the db will throw an error if an invalid term is used. There are some data integrity issues to think about but is this a low impact "high" effort task?

tests/test_sample.py
```diff
@@ -36,85 +36,64 @@ def test_add_sample(thing):
...
             "sampler": "Test Sampler A",
             "field_sample_id": "FS-12345",
+            "sampler_name": "Test Add Sampler",
         },
     )
     data = response.json()
-    assert response.status_code == 201
-    assert data["id"] is not None
-    assert data["thing_id"] == thing.id
+    assert data["id"] == data["id"]
```
jirhiker: These asserts don't test anything

tests/test_sample.py
```diff
@@ -36,85 +36,64 @@ def test_add_sample(thing):
...
-
-
 #  ============= Patch tests for samples =============================================
 def test_patch_sample(sample):
     """
     Test updating a sample in the collaborative network.
     """
-    original_method_patch = sample.sample_method
-    original_timestamp_patch = sample.sample_date
+    original_sample = deepcopy(sample)
```
jirhiker: I find there is almost never a need to make deep (or even shallow) copies

tests/test_sample.py
```diff
@@ -36,85 +36,64 @@ def test_add_sample(thing):
...
+    assert data["sample_date"] == sample_date_patch.split("+")[0]
+    assert data["sample_method"] == sample_method_patch
+    assert data["thing_id"] == sample.thing_id
 
     # cleanup after patching the sample
     with session_ctx() as session:
         updated_sample = session.query(Sample).filter(Sample.id == sample.id).one()
-        updated_sample.sample_method = original_method_patch
-        updated_sample.sample_date = original_timestamp_patch
+        updated_sample.sample_method = original_sample.sample_method
```
jirhiker: Is there really a need to rollback a patched sample?

tests/test_sample.py
```diff
@@ -147,15 +126,14 @@ def test_patch_sample_422_thing_id_not_found(sample):
...
-        }
-    ]
+    assert isinstance(data["detail"], list)
+    assert len(data["detail"]) == 1
+    error = data["detail"][0]
+    assert error["type"] == "value_error"
+    assert error["loc"] == ["body", "thing_id"]
+    assert error["msg"] == f"Value error, Thing with ID {bad_thing_id} does not exist."
+    assert error["input"] == bad_thing_id
+    assert error["ctx"] == {"error": {}}
```
jirhiker: Is there value in asserting the ctx?

tests/test_sample.py
```diff
@@ -178,48 +156,277 @@ def test_patch_sample_422_invalid_timestamp(sample):
...
-    assert response.status_code == 200
-    data = response.json()
-    assert "items" in data
-    assert len(data["items"]) > 0
+Both POST and PATCH endpoints use the same custom validators, which is why 
+they are being tested together here. For true unit testing, though, these can be separated
+into different POST and PATCH validation tests.
+"""
+
+
```
jirhiker: You don't need to call the api to test the validations. If the purpose of the unittest is to test the validation use the validator directly, not via calling the api

jirhiker (changes requested):

schemas_v2/sample.py
jacob-a-brown: We talked about "smart pipes/dumb container" for the API/database, making me think that we should validate incoming data as strictly and often as we can in the API. This may also make it easier to build custom error messages for display in the frontend. While these errors should and will be caught by the database, should we build in as many safe-guards as possible? Through the API (and frontend) when feasible. 

Or, would that make it overly complicated and difficult to maintain?

schemas_v2/sample.py
jacob-a-brown: To have as much heavy lifting be done by the API and have as many safeguards for data integrity and validity as we can.

tests/test_sample.py
jacob-a-brown: That's a silly validation... I think that was an autocomplete by AI, but I should have monitored its updates more closely. This will be amended in the patch to use the payload data.

tests/test_sample.py
jacob-a-brown: I suppose not... I was thinking that patching it changes the data and therefore subsequent tests. However, I'm just comparing the response to the data in the testing database and it doesn't really matter what is there (as long as it is valid).

tests/test_sample.py
jacob-a-brown: When Pydantic raises `ValueError`s FastAPI (or maybe all frameworks?) "Value error, " gets prepended to the message (e.g. `Value error, Thing with ID 999999 does not exist`). Do you think that we should use the ctx to craft human-readable errors to be used by the frontend?

tests/test_sample.py
jacob-a-brown: This will be present in the update push

schemas_v2/sample.py
```diff
@@ -16,80 +16,195 @@
...
+    #             raise ValueError(
+    #                 "Sample bottom cannot be greater than sample top."
+    #             )
+    #     return sample_bottom
+
+    # REFACTOR TODO: fields are evaluated in the order in which they are defined.
+    # are sample top/bottom really working as expected?
+
+    @field_validator("sample_top", check_fields=False)
+    def validate_sample_top(cls, sample_top: float | None, values) -> float | None:
```
chasetmartin: Instead of using field validators for sample top / sample bottom, could you use a @model_validator that runs once on multiple fields after pydantic does its other validations, like:
```
class CreateSample():
    ...
    
    @model_validator(mode='after')
    def validate_sample_topbottom(self) -> 'CreateSample':
        """
        validate sample_top and sample_bottom are provided together
        """
        if (self.sample_top is not None and self.sample_bottom is None) or \
           (self.sample_bottom is not None and self.sample_top is None):
            raise ValueError("Both sample_top and sample_bottom must be defined together")
        
        return self
```

Would this help with logic duplication between top and bottom field validators?

schemas_v2/sample.py
jirhiker: I understand the argument about custom error messages, but I think it's a little excessive at this point. 

I like the philosophy when programming "better to plead for forgiveness than ask for permission" i.e. assume valid input but handle exceptions when they occur.  A chief benefit is performance as the exception handling only happens when an invalid input is provided. With the current prior data validation model the check (e.g. call to the database to see if there is a conflict) always executes.

tests/test_sample.py
jirhiker: My point is, why do we need to test if ```assert error["ctx"] == {"error": {}}```. What benefit is there? 

Do we have a business case that requires the API to return error messages with a ctx field? Just because it's part of the response doesn't mean we need to assert its correctness

schemas_v2/sample.py
jacob-a-brown: `@model_validator` is probably the way to go. One of the things that I like about `@field_validator` is that it indicates in the response which fields were invalid and why, so if you want to display the error message beneath (or near) the invalid field it's easy for you.

In the response from `@field_validator`, for example, the invalid field is pointed to by `"loc": ["body", "sample_bottom"]`

Nevertheless, from reading forums and chatting with CoPilot `@model_validator` seems to be the best way to perform cross-field validations. And, if you want to proceed with error messages next to/near invalid fields, we can figure out how to do that efficiently/correctly with `@model_validator`

tests/test_sample.py
jacob-a-brown: My assumption was wrong. I just removed the rollback for the patched sample and in subsequent get tests when I invoke the `sample` fixture it still thinks that the patched fields are their original values. To prevent this cross-pollination from tests I'll keep the rollback functionality.

jacob-a-brown: `2025-01-01` is the original value and `2025-01-02` is the updated value.

image

tests/test_sample.py
jacob-a-brown: Fair point. I also removed these and am just testing the validators directly, not through endpoints.

schemas_v2/sample.py
jacob-a-brown: That's a good point. I also just asked CoPilot and Google's AI (and read some forums) and they agree with your assessment. I'll move this error handling to the endpoint with try/except clauses.

schemas_v2/sample.py
jacob-a-brown: database exceptions are now handled at the API patch and post endpoints and 409 errors are raised with informative datails

schemas_v2/sample.py
jacob-a-brown: I think it's a high impact task because it ensures the integrity of our database. It can, though, be handled in a future ticket when we address further restrictions of data. (or should that be handled now? there are a number of areas where we can begin to restrict data values)

jirhiker: in theory, i agree it adds data integrity, in practice, the only practical way to update data in the database is via the API and DataManager UI and the UI limits what data can be entered.  

Ultimately its a good thing for the backlog

api/sample.py
```diff
@@ -34,13 +35,41 @@
...
 # ============= Post =============================================
 @router.post("", status_code=HTTP_201_CREATED)
 def add_sample(sample_data: CreateSample, session: session_dependency):
     """
     Endpoint to add a sample.
     """
-    return adder(session, Sample, sample_data)
+    try:
+        return adder(session, Sample, sample_data)
+    except IntegrityError as e:
```
jirhiker: This doesn't need to be separate exception clauses

`except (IntegrityError, ProgrammingError) as e:`

tests/test_sample.py
```diff
@@ -14,104 +14,201 @@
...
+    assert data["sample_method"] == payload["sample_method"]
+    assert data["duplicate_sample_number"] == payload["duplicate_sample_number"]
+    assert data["sample_top"] == payload["sample_top"]
+    assert data["sample_bottom"] == payload["sample_bottom"]
 
     # cleanup after adding the sample
-    sample_id = data["id"]
     with session_ctx() as session:
-        session.query(Sample).where(Sample.id == sample_id).delete()
+        session.query(Sample).where(Sample.id == data["id"]).delete()
```
jirhiker: for consistency this should be 

`session.get(Sample, data["id"}).delete()`

tests/test_sample.py
```diff
@@ -14,104 +14,201 @@
...
+    assert data["sample_date"] == new_sample_date[:-1]
+    assert data["sample_method"] == new_sample_method
 
-    # cleanup after patching the sample
+    # rollback after updating the sample
     with session_ctx() as session:
-        updated_sample = session.query(Sample).filter(Sample.id == sample.id).one()
-        updated_sample.sample_method = original_method_patch
-        updated_sample.sample_date = original_timestamp_patch
+        updated_sample = session.query(Sample).filter(Sample.id == sample.id).first()
```
jirhiker: `updated_sample = session.get(Sample, sample.id)`

jirhiker (changes requested):
looking good. a few minor change requests

api/sample.py
jacob-a-brown: This will be present in the updated push

tests/test_sample.py
jacob-a-brown: This will be present in the updated push

tests/test_sample.py
jacob-a-brown: This will be present in the updated push

jirhiker (approved):

PR #63, opened 2025-08-05
feat: added 'within' spatial search to thing api

jirhiker:
### **Why**
- Frontend needs this functionality to implement its spatial search dialog

### **How**
- added `within` parameter to `thing` api

### **Notes**
N/A

### DIFF
--- a/api/thing.py
+++ b/api/thing.py
@@ -78,7 +78,13 @@ def get_things(
         return paginate(query=sql, conn=session)
     else:
         return get_db_things(
-            filter_, order, query, session, sort, thing_type, with_location=True,
+            filter_,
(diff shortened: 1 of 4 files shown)


chasetmartin (approved):
I tried using within=POLYGON() and it worked well for me. Is there a simple test we can write to go along with this change to ensure within continues working properly?

jacob-a-brown:
`within` can be validated in `schemas_v2/thing.py` to ensure it is a valid `POLYGON` or `MULTIPOLYGON`.

The tests can be

1. expected things are returned when a valid WKT is passed
2. a 422 error is raised if an invalid WKT is passed

PR #65, opened 2025-08-07
NO TICKET: update all datetime fields to be timezone aware in the database and schemas

jacob-a-brown:
### **Why**
- In the 2025-08-07 WDI Dev Team Rendezvous we decided that all datetime records will be stored in UTC. The frontend will convert datetimes to the appropriate time zone(s) for display (probably MT)

### **How**
- For the database models...
  - set all `DateTime` columns to `DateTime(timezone=True)`
  - set all `TIMESTAMP` columns to `TIMESTAMP(timezone=True)`
  - for all columns where a server_default may be called I changed `func.now()` to `func.timezone("UTC", func.now()` to ensure that the time zone is UTC. It defaults to the time zone of the database for `func.now()` so this change ensures the stored time zone is UTC if that is not how it is set in the db.
- For the Pydantic schemas...
  - used `pydantic.AwareDatetime` instead of `datetime.datetime` to require time zone aware fields to be passed and returned
  - ensured that any user-defined datetime fields must not be in the future by using Pydantic's `Annotated` field validations
  - converted user-defined datetime fields to UTC if not already in UTC

### **Notes**
- since all datetime field responses are UTC this uses the `Z` notation
- did I miss any models or fields?
  - the sensor model has yet to be refactored, so that was not updated
- the time zone must be set by the application layer. Because something(s) may slip through the cracks, should we convert `AwareDatetime` responses to UTC? Just to be extra safe.

### DIFF
--- a/api/sample.py
+++ b/api/sample.py
@@ -60,7 +60,9 @@ def database_error_handler(
 
 # ============= Post =============================================
 @router.post("", status_code=HTTP_201_CREATED)
-def add_sample(sample_data: CreateSample, session: session_dependency):
+def add_sample(
+    sample_data: CreateSample, session: session_dependency
+) -> SampleResponse:
     """
     Endpoint to add a sample.
     """
--- a/db/base.py
+++ b/db/base.py
@@ -57,7 +57,11 @@ class ReleaseMixin:
 class AuditMixin:
     @declared_attr
     def created_at(self):
-        return Column(DateTime, nullable=False, server_default=func.now())
+        return Column(
+            DateTime(timezone=True),
+            nullable=False,
+            server_default=func.timezone("UTC", func.now()),
+        )
 
 
 class AutoBaseMixin(AuditMixin):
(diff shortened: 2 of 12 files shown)


chasetmartin (approved):
I tried to replicate the error I was getting on sample edit, and this fixed the issue. The response is returning the timezone, so even when the user does no editing before save, the front end now send back the expected format via the roundtrip. 👍

One question: is PastDatetime a built in Pydantic validator to check that the datetime is not in the future?

jacob-a-brown:
[PastDatetime](https://docs.pydantic.dev/2.0/usage/types/datetime/#pydantic-date-types) is a built-in Pydantic validator (and type) to ensure that the supplied datetime is not in the future. This is what an error looks like from it (status code is still 422)

image

jirhiker (approved):

ksmuczynski (commented):
Looks good to me!

PR #66, opened 2025-08-07
NO TICKET: removed admin and old schemas, added observation/water-chemistry

jirhiker:
### **Why**
- The DataManager will serve as the admin UI
- There was some old code that was cluttering the code base
- Needed a POST observation/water-chemistry endpoint for use with the Water Chemistry CSV Import app

### **How**
- deleted schemas, admin
- added observation/water-chemistry to observation router

### **Notes**
- Use bullet points here


api/observation.py
```diff
@@ -46,6 +47,18 @@ def add_groundwater_level_observation(
     return add_observation(session, obs_data)
 
 
+@router.post("/water-chemistry", status_code=HTTP_201_CREATED)
+def add_water_chemistry_observation(
```
ksmuczynski, line 51: Is it necessary to specify a specific type of observation vs adding a general observation? I think the `sample_matrix` field in the `Sample` table would indirectly identify the type of observation (groundwater, surface water, soil, etc.)

services/observation_helper.py
```diff
@@ -29,12 +30,22 @@ def add_observation(session: Session, data: BaseModel) -> Base:
...
     #         sample = Sample(thing_id=thing_id,
     #                         collection_method=data.get('collection_method', 'manual'),
-    #                         collection_timestamp=data.get('observation_timestamp'))
+    #                         collection_timestamp=data.get('observation_datetime'))
     #         session.add(sample)
     #         data['sample'] = sample
     #     else:
     #         raise ValueError('Cannot specify both thing_id and sample_id')
-
+    if "field_sample_id" in data:
```
jacob-a-brown, line 38: I think that this if block can be removed because

- The `observation` model and schema don't have a `field_sample_id` field (this may be an artifact from a previous iteration of the schema?)
- `sample_id` is also a non-nullable field with a foreign key constraint, so it should be in the payload (not popped).
  - If the corresponding `sample` does not exist the database will raise an error and that can be handled at the endpoint (raise a 409 error with an appropriate error message).

jacob-a-brown (changes requested):
There's really only one schemas directory now. Should the original `schemas` directory be removed and `schemas_v2` be renamed `schemas`?

db/observation.py
```diff
@@ -90,6 +91,13 @@ class Observation(Base, AuditMixin, ReleaseMixin):
         doc="Temperature of the geothermal observation in degrees Celsius",
     )
 
+    # water chemistry
+    value = mapped_column(
+        Float,
+        nullable=True,
+    )
+    units = lexicon_term()
```
ksmuczynski, line 99: These fields apply to any parameter/observed property. What is the reasoning for grouping under `water chemistry`?

tests/test_observation.py
```diff
@@ -19,11 +19,31 @@
 
 
 # ============= Post tests =================
+def test_add_water_chemistry_observation(location, thing, sample, sensor):
```
ksmuczynski, line 22: Similar to a previous comment I had, if the type of observation is not necessary, perhaps consider `def test_add_observation` instead.

services/observation_helper.py
```diff
@@ -29,12 +30,22 @@ def add_observation(session: Session, data: BaseModel) -> Base:
...
+        field_sample_id = data.pop("field_sample_id")
+        data.pop(
+            "sample_id", None
+        )  # Ensure sample_id is not set if field_sample_id is used
+
+        sql = select(Sample).where(Sample.field_sample_id == field_sample_id)
+        sample = session.scalar(sql)
+        if not sample:
+            raise ValueError(f"Sample with id {field_sample_id} does not exist")
+        data["sample"] = sample
```
jacob-a-brown, line 48: I think that this if block can be removed because

- The `observation` model and schema don't have a `field_sample_id` field (this may be an artifact from a previous iteration of the schema?)
- `sample_id` is also a non-nullable field with a foreign key constraint, so it should be in the payload (not popped).
  - If the corresponding `sample` does not exist the database will raise an error and that can be handled at the endpoint (raise a 409 error with an appropriate error message).

ksmuczynski (commented):
Looks good overall. Left a few comments questioning the need to ad a specific type of observation as opposed to a general observation. Open to discussion.

jacob-a-brown (commented):
See previous comments

api/observation.py
jirhiker: By limiting to a type of observation, we can perform better data validations, i.e, if all the necessary fields are provided.  This would not be as easy or explicit with a general add_observation endpoint.

services/observation_helper.py
jirhiker: This is necessary for how the frontend works to upload the csv. 

The way the frontend current works, when adding an `Observation`, it needs to refer to the `Sample` via the `field_sample_id`  not the `Sample.id`

jacob-a-brown: is that be cause `Sample.id` is not defined until `Sample` has been created? If that's the case, could you use the `sample_id` from the POST `/sample` response on the frontend? Or, would that not be feasible given how the frontend is setup.

api/observation.py
ksmuczynski: So this is something that's necessary at the validation level? Additional fields are not needed in the `sample` or `observation` tables, correct?

services/observation_helper.py
jirhiker: "is that be cause Sample.id is not defined until Sample has been created?" correct

Thats probably the correct way to do it eventually, e.g. using the sample_id from POST response.

It wasn't feasible with the current structure, but I have some ideas on how to better implement this in the future and we can remove this `if` block

db/observation.py
jirhiker: fixed.  changed to `general observations`

services/observation_helper.py
jacob-a-brown: Sounds good! We can make a note of this and then when the new implementation is done on the frontend this `if` block can be removed. If a ticket is made for that implementation this could be a subtask. We can also add a `# REFACTOR TODO` note here as well.

jacob-a-brown (approved):
I noted down the future removal of the `if field_sample_id in data` in my backlog notes too

PR #68, opened 2025-08-08
NO TICKET: update README to reflect current status of repo & rename /migration to /transfers

jacob-a-brown:
### **Why**
- The README does not encompass the changes and updates to the repo. This includes new files, directories, and the implementation of Docker. 
- `migrations/` is the same nomenclature as database migrations, so it was renamed to `/transfers` and `migration.py`/`migration2.py` were renamed `transfer.py`/`transfer2.py`

### **How**
- Updated README.md
  - The steps for installation were split into different sections because some of them are different depending on your environment, such as Mac/Windows or PostgreSQL hosted locally vs using Docker

### **Notes**
- This uses the current nomenclature and files. Like all repos, the README will need to be maintained and updated as the repo is updated and changes.
- I could add more files in the main `app/` directory, but didn't want to make it overcrowded
  - `.env` is included to remind the user that it's necessary, even if none are hosted in GitHub. Should this reference be removed in README?


jacob-a-brown:
This is what the tables with different options looks like. If it doesn't render correctly in GitHub I'll change the format.

image

README.md
```diff
@@ -29,42 +29,68 @@ NMSampleLocations is a FastAPI-based backend service designed to manage geospati
...
+├── .pre-commit-config.yaml # pre-commit hook configuration file
+├── constants.py            # Static variables used throughout the code
+├── docker-compose.yml      # Docker compose file to build database and start server
+├── entrypoint.sh           # Used by Docker to run database migrations and start server
+├── main.py                 # FastAPI entry point
+|
+├── alembic/                # Alembic configuration and migration scripts
+├── api/                    # Route declarations
+├── core/                   # Settings and application config
+├── db/                     # Database models, sessions, migrations
```
ksmuczynski: Database migration scripts are stored in the `alembic` folder. What are `sessions`?

ksmuczynski (commented):
Database migrations are stored within `alembic/`, not `db/`.

README.md
jacob-a-brown: The code in `/migrations` migrates some staging data from NM Aquifer to the dreama. Should we rename that for clarity? Something like `data_transfer`?

Also, I think that should read `# Database models, sessions, engines`. `sessions` are when connections are opened to the database, work is performed, and then committed or rolled back (and finally closed). There's at function at the bottom of `db/engine.py` that retrieves a session.

In the `api/` scripts each endpoint has `session: session_depdency`, which evaluated to `session: Depends(get_db_session)`. All that's really doing is creating a new database session for an interaction every time the endpoint function is invoked.

jacob-a-brown: https://fastapi.tiangolo.com/tutorial/sql-databases/?h=session#create-a-session-dependency

jacob-a-brown:
The db portion of `Project Structure` now reads

```
├── db/                     # Database models, sessions, and engine
```

README.md
ksmuczynski: I like `data_transfer`. In my opinion, it is clearer than `migrations`, which we are already using a lot.

ksmuczynski (approved):

README.md
jacob-a-brown: I'll make that update and push again. I agree that it gets confusing by using `migration` for different things

jacob-a-brown: To keep it simple and one word I'll just use `transfer` if that's okay

jacob-a-brown:
after PR #66 is merged into `pre-production` I'll merge with this branch and fix merge conflicts that may arise from renaming `/migration` to `/transfers`. Then I'll push back here for that to be present in this PR.

jacob-a-brown:
@jirhiker and @ksmuczynski I merged `pre-production` with this branch, made some fixes (like using HTML tables to have multi-line code blocks), and corrected some typos. This is ready for another review.

jirhiker (approved):

ksmuczynski (approved):
Looks good!

PR #74, opened 2025-08-11
NO TICKET: Cascade lexicon term updates

jirhiker:
### **Why**
- Since the lexicon term is used as the foreign key changes to it need to be cascaded 

### **How**
- added `onupdate='CASCADE'` to lexicon_term function

### **Notes**
- use the `lexicon_term` convenience function in all models that use the lexicon

### DIFF
--- a/db/base.py
+++ b/db/base.py
@@ -41,9 +41,14 @@ make_searchable(Base.metadata)
 
 
 def lexicon_term(**kw):
-    return mapped_column(String(100), ForeignKey("lexicon_term.term",
-                                                 onupdate='CASCADE',
-                                                 ), **kw)
+    return mapped_column(
+        String(100),
+        ForeignKey(
+            "lexicon_term.term",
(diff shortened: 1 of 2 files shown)


db/base.py
```diff
@@ -41,7 +41,14 @@ class Base(DeclarativeBase):
...
 def lexicon_term(**kw):
-    return mapped_column(String(100), ForeignKey("lexicon_term.term"), **kw)
+    return mapped_column(
+        String(100),
+        ForeignKey(
+            "lexicon_term.term",
+            onupdate="CASCADE",
+        ),
+        **kw,
+    )
```
jacob-a-brown: the `lexicon_term` that was deleted from `db/lexicon.py` had `ondelete="CASCADE"`. Should that be preserved here too?

jacob-a-brown (approved):

ksmuczynski (approved):

PR #76, opened 2025-08-11
NO TICKET. Update `sample_matrix` and `sample_method` fields to be required.

ksmuczynski:
### **Why**
_The `sample_matrix` and `sample_method` fields should be required.

### **How**
_Implementation summary:_
- `nullable=False` parameters were set on the `sample_matrix` and `sample_method` fields.
### **Notes**
- Follow-up work: update the pydantic schemas and validation accordingly.

### DIFF
--- a/db/sample.py
+++ b/db/sample.py
@@ -85,7 +85,7 @@ class Sample(Base, AutoBaseMixin, ReleaseMixin):
         default=0,
         comment="Identifier for duplicate samples (0 = original sample, not a duplicate, 1 = dup no.1, 2 = dup no.2, etc.).",
     )
-    # TODO: Consider user research to determine if sample_group_type field is necessary
+    # TODO: Consider user research to determine if sample_group field is necessary
     sample_group: Mapped[str] = mapped_column(
         comment="Defines a sample's purpose, category, or group (e.g., 'geochemistry', 'geothermal', 'groundwater level')."
     )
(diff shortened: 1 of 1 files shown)


db/sample.py
```diff
@@ -83,8 +83,9 @@ class Sample(Base, AutoBaseMixin, ReleaseMixin):
         default=0,
         comment="Identifier for duplicate samples (0 = original sample, not a duplicate, 1 = dup no.1, 2 = dup no.2, etc.).",
     )
-    sample_type: Mapped[str] = mapped_column(
-        comment="The type of sample (e.g., 'geochemical', 'geothermal', 'groundwater')."
+    # TODO: Consider user research to determine if sample_group field is necessary
+    sample_group: Mapped[str] = mapped_column(
+        comment="Defines a sample's purpose, category, or group (e.g., 'geochemistry', 'geothermal', 'groundwater level')."
```
jacob-a-brown, line 88: Renaming from `sample_type` to `sample_group` caused the tests to fail. The following need to be updated to correspond with this change so the tests pass:

- `Create`, `Update`, and `Response` schemas
- the `sample` fixture in `tests/conftest.py`
- references to `sample_type` in `tests/conftest.py`

jacob-a-brown (commented):
You note that schemas need to change because of these changes, but since they aren't yet updated the tests are failing. Since the model changes were made here, do should the schemas and tests be updated, too? So that all related changes are made in the same PR and can be easily referenced and understand

PR #78, opened 2025-08-11
BDMS-47 transfer

jirhiker:
FYI This is a bit of a hodgepodge of a PR. 

### **Why**
- Need to transfer data from NM_Aquifer to new system
- needed to evaluate authentication and authorization 

### **How**
- added transfer_springs 
- added transfer_waterlevels
- added user dependencies to things api
- added ability to extra user info/permissions from JWT
- fixed search for use with UIs search bar
- added created_by... concept to AuditMixin
- cleaned up lexicon.json for readability

### **Notes**
- created_by... needs to be addressed. it currently is storing both a created_by_id and created_by_name.  need to decide what identifier(s) we want to store. Do we need a "Users" table? (id very much prefer that user/Identifies be managed by Authentik)


api/observation.py
```diff
@@ -110,6 +110,7 @@ def get_groundwater_level_observations(
         sql = sql.where(Observation.observation_datetime <= end_time)
 
     sql = order_sort_filter(sql, Observation, sort, order, filter_)
+    sql = sql.order_by(Observation.observation_datetime.desc())
```
jacob-a-brown: Should this be contained in an `if` block incase the user requests an order? Something like 

```python
if order is None:
    sql = sql.order_by(Observation.observation_datetime.desc())
```

api/search.py
```diff
@@ -118,7 +152,7 @@ def search_api(q: str, session: Session = Depends(get_db_session)):
     results.extend(_get_thing_results(session, q))
     results.extend(_get_asset_results(session, q))
 
-    return results
+    return {"items": results, "total": len(results)}
```
jacob-a-brown: This could return many items. Should this be paginated? If it's not possible with the paginate functions we've been using it can be implemented manually. The question, though, is how to order the results since they come from different models

core/dependencies.py
jacob-a-brown, line 1: Is there documentation somewhere that describes the different roles an what the permissions mean? Either in Authentik, an ADR, or here? If not, that may be worth writing down to help us keep track of what is what.

jacob-a-brown:
It also just occurred to me, but we should override the authentication for the tests, as well as test that the authentication is being done correctly.

core/dependencies.py
jirhiker: Thats a good idea

jirhiker:
The authentication is being overriden for `test_thing`

jirhiker:
RE available ids for audit trail

TODO: there is probably no need to store both the id and name of the user. We should decide which attribute
    #  included in the token payload is the most appropriate to use as the user identifier.
    # included "claims"
    # sub:  the subject of the token, which is a unique identifier for the user
    # name: the name of the user
    # email: the email address of the user
    # given_name: the first name of the user
    # preferred_username: the preferred username of the user
    # nickname: the nickname of the user

api/search.py
jirhiker: added pagination and a limit parameter

jacob-a-brown (approved):

PR #79, opened 2025-08-12
BDMS 55: ensure location router is fully covered | remove expand functionality

jacob-a-brown:
### **Why**
- All routers should have full GET, POST, PATCH, and DELETE coverage and tests

### **How**
- the `location` router was missing the following endpoints, which have now been implemented and tested
  - DELETE `/location/{location_id}`
- GET `/location/{location_id}` has been simplified.
- POST and PATCH tests have been simplified and cleaned up
- 404 tests for GET `/location/{location_id}` and PATCH `/location/{location_id}` were added

### **Notes**
- The `expand` query parameter is not used for `/location` GET requests, so it was removed to simplify the code base
- Some tests in `tests/test_thing.py` were causing tests in `tests/test_location.py` to fail, so they were amended to improve test independence.
- The function `model_deleter` was added to `services/crud_helper.py` to be reusable for all models.

### DIFF
--- a/api/location.py
+++ b/api/location.py
@@ -172,12 +172,11 @@ async def get_location_by_id(
 @router.delete("/{location_id}", summary="Delete location by ID")
 async def delete_location(
     location_id: int, session: Session = Depends(get_db_session)
-) -> None:
+) -> Response:
     """
     Delete a sample location by ID from the database.
     """
-    model_deleter(session, Location, location_id)
-    return Response(status_code=204)
+    return model_deleter(session, Location, location_id)
 
 
 # ============= EOF =============================================
--- a/api/location.py
+++ b/api/location.py
@@ -15,7 +15,7 @@
 # ===============================================================================
 from typing import Union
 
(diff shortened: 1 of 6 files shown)


api/location.py
```diff
@@ -154,42 +150,34 @@ async def get_location(
...
-
-    return response_klass.model_validate(location)
+@router.delete("/{location_id}", summary="Delete location by ID")
+async def delete_location(
+    location_id: int, session: Session = Depends(get_db_session)
+) -> None:
+    """
+    Delete a sample location by ID from the database.
+    """
+    model_deleter(session, Location, location_id)
```
jirhiker: Make model_deleter return the 204 Response

`return model_deleter(...`

jirhiker (changes requested):

api/location.py
jacob-a-brown: `model_deleter` now returns 204 response

jirhiker (approved):

PR #81, opened 2025-08-13
BDMS 55: sensor coverage | reusable test functions | PydanticStyleException

jacob-a-brown:
### **Why**
- The `/sensor` router should be fully covered with GET, GET by ID, PATCH, POST, and DELETE endpoints
- Exceptions raised outside of Pydantic should maintain the same style to enable consistency and ease of use throughout the API

### **How**
- Added PATCH and DELETE endpoints
- Ensured all endpoints are covered in tests
- Write a PydanticStyleException class in `services/exception_helper.py`

### **Notes**
- Updated the `sensor` fixture to populate all `sensor` fields
- Updated the model and schema to rename `date_` fields `datetime_` to reflect their data type 
- Made all `datetime_` fields time zone aware and UTC
- Fixed custom validation tests for other schemas


tests/test_sample.py
```diff
@@ -61,8 +61,8 @@ def test_validate_sample_top_and_bottom():
             )
         except ValueError as e:
```
jirhiker: use `pytest.raises` to test for exceptions

jirhiker (changes requested):

tests/test_sample.py
jacob-a-brown: `pytest.raises` is now used to test for exceptions

jirhiker (approved):
👍

PR #83, opened 2025-08-16
NO TICKET: Coverage and Feature support

jirhiker:
### **Why**
- Need full functionality of lexicon editing
- Groups need project areas
- Need to transfer contacts to db
- Need to transfer assets to GCS and db 

### **How**
- added patch for category and term
- added user permission dependencies for all endpoints
- added sorting
- added get term and category by id
- added transfer_owners to transfer2.py
- added transfer_assets to transfer2.py
- dont use signed_urls for gcs.  

### **Notes**
- may need to revisit signing of GCS urls.


api/asset.py
```diff
@@ -24,32 +24,89 @@
...
+        audit_add(user, assoc)
+        thing = session.get(Thing, thing_id)
+        assoc.thing = thing
+        assoc.asset = asset
+        session.add(assoc)
+
+    session.add(asset)
+    session.commit()
+    session.refresh(asset)
+    asset.url = url
```
jacob-a-brown: The `url` is popped from the incoming data and then added back again at the end. Because it appears to be an integral part of the model, and to simplify this process, the `url` can be added to the database model and the schema. Then it won't need to be handled in this way.

api/asset.py
```diff
@@ -60,16 +117,9 @@ async def list_assets(
...
-    # if not assets:
-    #     return []
-
     def transformer(assets: List[Asset]) -> AssetResponse:
-        # blob = bucket.blob(asset.storage_path)
-        # asset.url = blob.generate_signed_url(expiration=timedelta(minutes=10), method="GET")
-        # return [AssetResponse.model_validate(asset) for asset in assets]
         for a in assets:
-            a.url = f"https://storage.googleapis.com/{GCS_BUCKET_NAME}/{a.storage_path}"
+            set_asset_url(a, bucket)
```
jacob-a-brown: see note about adding `url` to asset model and schema. If it's part of the model/schema it will have already been set and this step can be avoided

api/asset.py
```diff
@@ -99,59 +149,23 @@ async def get_asset(
     if not asset:
         raise HTTPException(status_code=404, detail="Asset not found")
 
-    blob = bucket.blob(asset.storage_path)
-    asset.url = blob.generate_signed_url(expiration=timedelta(minutes=10), method="GET")
-    return asset
-
+    set_asset_url(asset, bucket)
```
jacob-a-brown: see note about adding `url` to asset model and schema. If it's part of the model/schema it will have already been set and this step can be avoided

api/geospatial.py
```diff
@@ -55,6 +59,32 @@ async def get_geospatial(
...
+            {
+                "type": "Feature",
+                "geometry": json.loads(to_geojson(to_shape(group.project_area))),
+                "properties": {
+                    "group_id": group.id,
+                    "group_name": group.name,
+                    "group_description": group.description,
+                },
+            }
+        ]
```
jacob-a-brown: If the `group` does not have a defined `project_area` this will error out. To avoid this, define `features` as an empty set before the `if` statement and append the dictionary to it, or define it as an empty set in an `else` statement.

api/group.py
```diff
@@ -49,12 +56,14 @@ def create_group(group_data: CreateGroup, session: session_dependency):
     status_code=status.HTTP_201_CREATED,
 )
 def create_group_thing(
-    group_location_data: CreateGroupThing, session: session_dependency
+    group_location_data: CreateGroupThing,
+    session: session_dependency,
+    user: admin_dependency,
```
jacob-a-brown: In my update to the `contact` router I've named the endpoint `/thing-association` to clarify what is the `contact` is being associated with. That could be done here, too.

Also, in our standup today I spoke about how creating an association can be done by defining the `group_id` in the path or query parameters (in our discussion I talked about the `contact` router). I decided to go with path parameters (e.g. `/group/{group_id}/thing-association`), but the `group_id` can just as well be defined in the payload. Should that be the case here, or should that change in the `contact` router?  Either way, a decision should be made and should be implemented consistently consistent across the API.

jacob-a-brown: Now that I think about it, though, most POST endpoints let the user-defined parameters be defined in the payload (which `group_id` and `contact_id` are)...

api/thing.py
```diff
@@ -381,4 +382,16 @@ def update_thing_id_link(
...
+    well_screen_id: int,
+    well_screen_data: UpdateWellScreen,
+    session: session_dependency,
+    user: editor_dependency,
+) -> WellScreenResponse:
+    return model_patcher(
+        session, WellScreen, well_screen_id, well_screen_data, user=user
+    )
+
+
```
jacob-a-brown: Will you add a `# TODO` comment here to validate the incoming data against existing data and raise a 409 error if something is amiss? Such as the user updating screen top to be below the existing screen bottom

schemas/group.py
```diff
@@ -13,12 +13,29 @@
...
+    description: str | None = None
+    parent_group_id: int | None = None
+    project_area: str | None = None
+
+    @classmethod
+    @field_validator("project_area")
+    def validate_area_is_wkt(cls, wkt):
+        return validate_wkt_geometry(wkt)
+
+
```
jacob-a-brown: This `field_validator` can be put into its own class from which `CreateGroup` and `UpdateGroup` inherit so that it's defined for and utilized by both. I learned that for it to work, `project_area` will need to be defined in that class, too. 

```python

class ValidateGroup(BaseModel):

    project_area: str | None = None

    @classmethod
    @field_validator("project_area")
    def validate_area_is_wkt(cls, wkt):
        return validate_wkt_geometry(wkt)
```

schemas/location.py
```diff
@@ -41,24 +42,8 @@ class CreateLocation(BaseModel):
...
-class CreateGroup(BaseModel):
-    """
-    Schema for creating a group.
-    """
-
-    name: str
-    description: str | None = None
-    parent_group_id: int | None = None
+    def validate_point_is_wkt(cls, wkt):
+        return validate_wkt_geometry(wkt)
```
jacob-a-brown, line 46: See comment about sharing validators in `schemas/group.py`.

schemas/thing.py
```diff
@@ -164,6 +166,8 @@ class WellScreenResponse(ORMBaseModel):
     thing_id: int
     screen_depth_bottom: float
     screen_depth_top: float
+    screen_type: str | None = None
+    screen_description: str | None = None
 
 
 class GroupResponse(ORMBaseModel):
```
jacob-a-brown: There's already a `GroupResponse` schema defined in `schemas/group.py` that is the same as this. This can be removed and the one from `schemas/group.py` can be inherited and used

api/asset.py
jirhiker: the url is generated on demand and not saved to the database.

api/asset.py
jirhiker: set_asset_url is necessary because the URL is generated on demand. This is so that the url can be signed and access can expire

api/geospatial.py
jirhiker: added 
```python
if not group.project_area:
    raise HTTPException(status_code=404, detail="Group has no project area")
```

api/group.py
```diff
@@ -15,32 +15,38 @@
...
+    group_data: CreateGroup, session: session_dependency, user: admin_dependency
+) -> GroupResponse:
     """
     Create a new group in the database.
     """
-    return adder(session, Group, group_data)
+    return adder(session, Group, group_data, user=user)
 
 
 @router.post(
```
jacob-a-brown: Updates look good, but what are your thoughts on using `/thing-association` to be explicit about which table this association references? It may be a moot point here since there's only one association with a group (thing), but if other tables have multiple associations it'd be good to have clear nomenclature and then keep a consistent style throughout the API.

jacob-a-brown: I looked through the database and this is relevant for the `author` table (`AuthorContactAssociation`, `AuthorPublicationAssociation`), so if we decide on a nomenclature now I'll implement it throughout and write it up in the API discussion (which will turn into ADR)

(may also be relevant to others, but this showed up first in my search)

jacob-a-brown (commented):
Per our discussion today, comment out the `/association` endpoints. We may not want users to directly interact with these tables; that will be determined in conversations with AMP.

Should the responses have lists of the associations, at least their IDs? So `GroupResponse` could also have

```python
associations: dict[str, List[int]]
```

e.g.
```python
associations: {"thing": [1, 2, 3]}
```

if that group is associated with `things` 1, 2, and 3. This allows multiple associations to exist and be returned (like the authors)

Or keep the GET endpoints (and name their associations, like `/group/thing-association`

jacob-a-brown:
Another thought...

proceed with commenting out all `/association` references and leave them be so that we can get this PR merged and the PR with the contact coverage updated merged. Then on Thursday when we convene for the WDI dev team rendezvous we discuss/come to a consensus on how we want to handle associations (in the responses for the objects, by themselves, or something else altogether)

jacob-a-brown (approved):

PR #91, opened 2025-08-20
BDMS 55: observation coverage & other updates

jacob-a-brown:
### **Why**
- The `observation` router should be fully covered with GET, GET by ID, POST, PATCH, and DELETE endpoints

### **How**
- Added POST `/observation/groundwater-level`, `/observation/geothermal`, `/observation/water-chemistry` endpoints
- Added PATCH `/observation/groundwater-level/{observation_id}`, `/observation/geothermal/{observation_id}`, `/observation/water-chemistry/{observation_id}` endpoints
  - 404 errors are raised if the `observation_id` does not exist in the database
  - 404 errors are raised if the `observation_id` exists in the database but its observation class (e.g. water chemistry) does not correspond to the endpoint. the actual observation class of the record with that ID is specified in the error message
- Added GET `/observation/ and `/observation/{observation_id}`
  - Returns fields that are not relevant to a particular observation's observation class
- Added GET `/observation/groundwater-level`, `/observation/geothermal`, `/observation/water-chemistry` endpoints
- Added GET `/observation/groundwater-level/{observation_id}`, `/observation/geothermal/{observation_id}`, `/observation/water-chemistry/{observation_id}` endpoints
  - 404 errors are raised if the `observation_id` does not exist in the database
  - 404 errors are raised if the `observation_id` exists in the database but its observation class (e.g. water chemistry) does not correspond to the endpoint. the actual observation class of the record with that ID is specified in the error message
- Added DELETE `/observation//{observation_id}`
- Updated and added new tests

### **Notes**
- Different `observation_class`es (e.g. `groundwater level`, `water chemistry`) are defined by `observed_property`. This is pretty easy when there's only one property per class, like `groundwater level`, but can become overwhelming and difficult to maintain when there are multiple properties per class, like `water chemistry`. Maybe this will become a moot point when we make the views, and those can be used?
- add `deg C` to lexicon for geothermal observed property
- removed `__table_args__` and `id` from Observation model since the primary key is just the `id`, which is defined in `AutoBaseMixin`
- removed `depth_to_water` from the model since that's encapsulated in `value` and `observed_property`
- `GroundwaterLevelResponse` calculates depth to water below ground surface as `value - measuring_point_height`
  - should the calculated `depth_to_water_bgs` be called `calculated_depth_to_water_bgs` for clarity and to be explicit?


api/observation.py
```diff
@@ -40,62 +57,125 @@
...
+
+def specify_observation_class(observation_class: str | None):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            sql, session = func(*args, **kwargs)
+            if observation_class == "groundwater level":
+                sql = sql.where(Observation.observed_property == "groundwater level")
+            elif observation_class == "water chemistry":
+                sql = sql.where(Observation.observed_property != "groundwater level")
```
jirhiker: Lets think of a better long term solution for this.  A simple approach would be to add a prefix to observed_property .e.g. `water chemistry:pH`, `water chemistry:TDS` etc 

something more complex could involve adding a parent pk to Category. (allowing Categories of Categories)
sql = sql.join(LexiconTerm).join(Category).where(Category.parent == 'water chemistry')

api/observation.py
```diff
@@ -40,62 +57,125 @@
...
+            elif observation_class == "geothermal":
+                sql = sql.where(Observation.observed_property == "temperature")
+            return paginate(query=sql, conn=session)
+
+        return wrapper
+
+    return decorator
+
+
+def get_observations(
```
jirhiker: seems like `observation_class`  can be an arguement to `get_observations`

api/observation.py
```diff
@@ -114,7 +194,181 @@ def get_groundwater_level_observations(
...
+
+@router.delete(
+    "/{observation_id}",
+    summary="Delete an observation",
+    status_code=HTTP_204_NO_CONTENT,
+)
+def delete_observation(
+    session: session_dependency, user: amp_admin_dependency, observation_id: int
+) -> None:
+    return model_deleter(session, Observation, observation_id)
```
jirhiker, line 324: I general I lke the use of decorators but here it seems like it just is spreading out the logic between `get_observations` and `specify_observation_class`

db/observation.py
```diff
@@ -78,25 +59,12 @@ class Observation(Base, AuditMixin, ReleaseMixin):
     level_status = lexicon_term()
 
     # geothermal
-    depth = mapped_column(
+    observation_depth = mapped_column(
```
jirhiker, line 62: @jacob-a-brown, @ksmuczynski should this be part of the Sample model?

services/observation_helper.py
```diff
@@ -1,58 +1,92 @@
...
+
+from db import Observation
+from services.exceptions_helper import PydanticStyleException
+from services.query_helper import simple_get_by_id
+
+observation_class_to_observed_properties = {
+    "groundwater level": ["groundwater level"],
+    "geothermal": ["temperature"],
+    "water chemistry": ["pH", "Alkalinity as CaCO3"],
+}
```
jirhiker: see comment above regarding categorization of observerd properties

jirhiker (changes requested):

api/observation.py
jacob-a-brown: That's probably a simpler way to do it than this decorator. I was working to avoid having `observation_class` show up as a query parameter and was initially imagining that `get_observations` would also be used as the endpoint functions. Nevertheless, that seems like a much simpler solution. It was fun to make this decorator on decorator function though!

api/observation.py
jacob-a-brown: `observation_class` is now an argument of `get_observations`

api/observation.py
jacob-a-brown: The first approach seems to be simpler. It would be easy to retrieve those records as

`sql = sql.where(Observation.observed_property.like(observation_class)`

and in the response the prefix can be removed

```python

class ObservationResponse(...):
    observed_property: str

    @field_validator("observed_property")
    def remove_observation_class_from_observed_property(cls, observed_property):
        colon_index = observed_property.find(":")
        cleaned_observed_property = observed_property[colon_index:]
        return cleaned_observed_property
```

This finds the first instance of the search string (`":"`), so if a colon is in the name of the observed property it won't effect the result.

jirhiker: does the prefix need to be removed?

jacob-a-brown: I think for it to correspond more closely with how other data are returned it would be good to remove it. Rather than have this prefix, we could instead have a column called `observation_class` that gets automatically populated by the API on POST requests (it would be a `lexicon_term`)

jacob-a-brown:
The latest update gets the `observation_class` from the path of the request. This makes the code easier to maintain because that information is contained and retrieved from one location, rather than many different locations.

jirhiker (approved):

PR #92, opened 2025-08-21
HOT FIX:  Asset list fix

jirhiker:
### **Why**
- There is a bug in the get assets endpoint. There is a type error when using the pagination transformer

### **How**
- Fixed the transformer function. 
- added type annotations

### **Notes**
- !Remember! the transformer function takes and returns a list, not a single record

### DIFF
--- a/api/asset.py
+++ b/api/asset.py
@@ -108,7 +108,7 @@ async def list_assets(
 
     def transformer(records: list[Asset]):
         if thing_id is not None:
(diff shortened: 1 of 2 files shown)


jacob-a-brown (approved):

jacob-a-brown:
I ran into this same error and evaluated for a list in the `add_signed_url` function, but I think that it's better to take care of this in the transformer itself.

PR #97, opened 2025-08-22
BDMS 55: asset coverage

jacob-a-brown:
### **Why**
- The `asset` router should be fully covered with GET, GET by ID, POST, PATCH, and DELETE endpoints

### **How**
- Added PATCH endpoint
  - the only fields that can be updated are `name` and `label`. Is this correct? If they're updated in the database does anything need to change in GCS?
- Added DELETE endpoint 
  - DELETE the asset from the database
  - DELETE the asset from GCS

### **Notes**
- Did I implement DELETE `/asset/{asset_id}/remove` correctly to remove a blob from GCS? I'm using the asset's `uri` to remove.
- This is a sequential PR and should **NOT** be merged until PR #91 is reviewed, amended, and merged.


api/asset.py
```diff
@@ -31,20 +32,52 @@
...
 
 router = APIRouter(
     prefix="/asset", tags=["asset"], dependencies=[Depends(viewer_function)]
 )
 
 
-# ======= Create =========
+def database_error_handler(payload: CreateAsset, error: ProgrammingError) -> None:
+    """
+    Handle errors raised by the database when adding or updating a sample.
```
jirhiker: ... an asset.

api/asset.py
```diff
@@ -107,10 +150,7 @@ async def list_assets(
         )
 
     def transformer(records: list[Asset]):
-        if thing_id is not None:
-            bucket = get_storage_bucket()
-            records = [add_signed_url(ai, bucket) for ai in records]
-
+        records = [add_signed_url(ai, bucket) for ai in records]
```
jirhiker: the thing_id guard was there on purpose. I do not want to generate signed urls if the thing_id is not set.

jirhiker (changes requested):

api/asset.py
jacob-a-brown: whoops! this was copy-pasted from `api/sample.py`. it's always the same function, but the error message and circumstances change so I'm defining them uniquely in each router where it's applicable. think I should make this a function (`services/exceptions_helper.py`)?

api/asset.py
jacob-a-brown: If a user sends a GET request and doesn't filter for `thing_id`, but an `asset` is associated with a `thing`, should the response for that record have a signed URL? (and responses for records with no association will not have a signed URL)

jacob-a-brown: Essentially

```python
def transformer(records: list[Asset]):
    records = [add_signed_url(ai, bucket) if ai.things != [] else ai for ai in records]
    return records
```

jacob-a-brown: I implemented this in my latest push. One of my concerns is that this could be slow because calling `ai.things` does a lazy load, I think. I imagine this could be improved and made faster, but this is a solution that comes to mind initially.

I am also checking for thing associations in `/asset/{asset_id}`

jirhiker: no. the signed_url should only be generated when the user specifies a thing_id via the get_assets function.  The original code was sufficient

There is no reason to generate signed urls when listing all assets.  There is a reason to generate signed urls when listing assets for a given thing_id because this is used by the front end to display a gallery of images all at once. This is the only case in which signer urls should be generated for a list of assets. A signed url is always generated when retrieving assets individually

api/asset.py
```diff
@@ -107,10 +151,9 @@ async def list_assets(
         )
 
     def transformer(records: list[Asset]):
-        if thing_id is not None:
-            bucket = get_storage_bucket()
-            records = [add_signed_url(ai, bucket) for ai in records]
-
+        records = [
+            add_signed_url(ai, bucket) if ai.things != [] else ai for ai in records
```
jirhiker: this is not the pythonic way to test for an empty list. typically this would be
```python
add_signed_url(ai, bucket) if ai.things else ai ...

or
add_signed_url(ai, bucket) if ai.things and len(ai.things)>0 else ai ...
```

in this case however the original code should be used

```python
 def transformer(records: list[Asset]):
        if thing_id is not None:
            bucket = get_storage_bucket()
            records = [add_signed_url(ai, bucket) for ai in records]

        return records
```

jirhiker (changes requested):

api/asset.py
```diff
@@ -110,7 +151,6 @@ def transformer(records: list[Asset]):
         if thing_id is not None:
```
jirhiker, line 163: Lets add a comment explaining this behavior e.g.

```Do not generate signed urls when listing ALL assets. There is a reason to generate signed urls when listing assets for a given `thing_id` because this is used by the front end to display a gallery of images all at once. This is the only case in which signed urls should be generated for a list of assets. A signed url is always generated when retrieving assets individually```

jirhiker (changes requested):
One last small change. Otherwise looking great

api/asset.py
jacob-a-brown: This note was added at the beginning of the `GET` section of the file since it applies to both GET `/asset` and GET `/asset/{asset_id}`

jirhiker (approved):

PR #98, opened 2025-08-22
BDMS 55: group coverage & other updates

jacob-a-brown:
### **Why**
- The `group` router should be fully covered with GET, GET by ID, POST, PATCH, and DELETE endpoints

### **How**
- Added DELETE `/group/{group_id}` endpoint
- Added check to `ValidateGroup` to ensure the `project_area` WKT is a `MULTIPOLYGON`
- Updated tests to use fixtures
- Added authentication to `group`, `sensor`, `sample`, and `location` routers
- Cleaned up `sensor`, `sample`, and `location` routers to use `session_dependency` per standard that was established after those routers were established

### **Notes**
- This is a sequential PR and should **NOT** be merged until PR #97 is reviewed, amended, and merged


jirhiker:
can we avoid sequential merges in the future. Its hard to do the review because its a mix of changes from this PR and PR #97.

jirhiker (approved):

PR #104, opened 2025-08-25
BDMS-81 fix: correct GCS_SERVICE_ACCOUNT_KEY echo syntax and update relationship in Thing model

jirhiker:
### **Why**
- Github action that deploys staging version does not set the GCS_SERVICE_ACCOUNT_KEY correctly

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/.github/workflows/staging_deploy.yml
+++ b/.github/workflows/staging_deploy.yml
@@ -48,7 +48,7 @@ jobs:
             echo "  CLOUD_SQL_DATABASE: \"${{ secrets.CLOUD_SQL_DATABASE }}\"" >> app.yaml
             echo "  CLOUD_SQL_USER: \"${{ secrets.CLOUD_SQL_USER }}\"" >> app.yaml
             echo "  CLOUD_SQL_PASSWORD: \"${{ secrets.CLOUD_SQL_PASSWORD }}\"" >> app.yaml
-            echo "  GCS_SERVICE_ACCOUNT_KEY: \"${{ secrets.GCS_SERVICE_ACCOUNT_KEY}}\"" >> app.yaml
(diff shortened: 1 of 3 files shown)


.github/workflows/staging_deploy.yml
```diff
@@ -48,7 +48,7 @@ jobs:
             echo "  CLOUD_SQL_DATABASE: \"${{ secrets.CLOUD_SQL_DATABASE }}\"" >> app.yaml
             echo "  CLOUD_SQL_USER: \"${{ secrets.CLOUD_SQL_USER }}\"" >> app.yaml
             echo "  CLOUD_SQL_PASSWORD: \"${{ secrets.CLOUD_SQL_PASSWORD }}\"" >> app.yaml
-            echo "  GCS_SERVICE_ACCOUNT_KEY: \"${{ secrets.GCS_SERVICE_ACCOUNT_KEY}}\"" >> app.yaml
+            echo "  GCS_SERVICE_ACCOUNT_KEY: ${{ secrets.GCS_SERVICE_ACCOUNT_KEY }} >> app.yaml
```
jacob-a-brown, line 51: Closing quotes should be present at the end of the line that is being echoed and then redirected to `app.yaml`

jirhiker: I suspect this is part of the problem.  Since the service key is JSON that uses double quotes

jacob-a-brown: For `echo` to work I think that the full string needs to be in double quotes, but the escaped quotes in the inner bit should be removed if they're causing issues

```bash
 echo "  GCS_SERVICE_ACCOUNT_KEY: ${{ secrets.GCS_SERVICE_ACCOUNT_KEY }}" >> app.yaml
```

PR #106, opened 2025-08-26
BDMS 55: lexicon coverage & other updates

jacob-a-brown:
### **Why**
- The `lexicon` router should be fully covered with GET, GET by ID, POST, PATCH, and DELETE endpoints
- tests that relied on spillover need to be amended for the use of fixtures
- staging data scripts need to be usable by all operating systems

### **How**
- Cleaned up POST `/triple` and `/category` to adhere with most recent styles (i.e. use `model_adder`)
- Added PATCH `/term/{term_id}`, `/category/{category_id}`, `/triple/{triple_id}`
- Added GET `/category/{category_id}`, `/triple`, `/triple/{triple_id}`
- Added DELETE `/term/{term_id}`, `/category/{category_id}`, `/triple/{triple_id}`

### **Notes**
- `adder` was moved from `db/__init__.py` to `services/crud_helper/py` and renamed `model_adder` to maintain a consistent style and nomenclature for crud functions.
- `Lexicon` was renamed `LexiconTerm` for clarity. `Category` was renamed `LexiconCategory` for clarity.
- `LexiconTerm` and `LexiconCategory` are now associated via their IDs, not the term's term and category's name.
- **LexiconTerm**
  - Previously a new `LexiconTerm` could only be associated with a single `LexiconCategory`. In this PR it can now be associated with multiple `LexiconCategories`.
  - When creating a new `LexiconTerm`, if any associated `LexiconCategories` are not present in the database they will be added (this functionality was maintained, though expanded to create multiple new records if they didn't previously exist). Do we want to maintain this functionality? Or should categories only be created via POST `/category`?
  - Newly crated `LexiconTerms` must be associated with at least one category.
  - Because when a lexicon term is added it can be associated with multiple categories, **core/lexicon.json** was updated to rename `category` `categories`, which evaluates to a list of dictionaries with the category `name` and `description`.
- **LexiconTriple**
  - A `LexiconTriple` is now deleted if the `subject` or `object_` are deleted because of foreign key constraints.
  - When a `LexiconTriple` is added it can be associated with any `LexiconTerm` (`subject` and `object_`). If that `LexiconTerm` does not exist it will be created (and therefore new `LexiconCategories` can be created when the terms are created). Do we want to maintain this functionality? Or should terms and categories only be created via their POST endpoints?
- **transfers**
  - `transfers/transfer.py` was deleted since it is out of date. `transfers/transfer2.py` was then renamed `transfers/transfer.py`.
  - `__init__.py` was added to `transfers/` and all import paths are defined relative to the root directory.
  - Data transfers can now be done by calling `python -m transfers.transfer` from the root directory.


jirhiker (changes requested):
the images in the transfer folder should not be included.  the transfer directory contains an assets folder where images go. this folder is in .gitignore.  All the data in the transfer folder is just for testing and does not need to be version controlled

jirhiker (changes requested):
all the changes to transfer should be removed.  These are in conflict with https://github.com/DataIntegrationGroup/NMSampleLocations/tree/jir-transfer

jirhiker (approved):

PR #107, opened 2025-08-28
BDMS-47: Transfer Utilities and updates

jirhiker:
### **Why**
- Need to transfer data from NM_Aquifer to Ocotillo

### **How**
- updated transfer scripts

### **Notes**
- Necessary schema updates and tests have not been implemented yet


core/lexicon.json
jacob-a-brown, line 1: Now that a `LexiconTerm` can be associated with multiple categories `category` should be renamed `categories`. It should evaluate to a list of categories. For example

```
{"categories": [{"name": "relation", "description: None}], "term": "OSEWelLTagID", "definition": "NM OSE well tag ID"}
```

This update was made for `lexicon` coverage and is present in `pre-production`. I'm not sure why it's not showing up in the comparison on the left.

api/asset.py
```diff
@@ -204,6 +205,8 @@ async def update_asset(
 async def delete_asset(
     asset_id: int, session: session_dependency, user: admin_dependency
 ):
+
+    # TODO: Interesting issue here.  we don't have a way of tracking who deleted a record
```
jacob-a-brown, line 209: This can be tracked by [SQLAlchemy-Continuum ](https://sqlalchemy-continuum.readthedocs.io/en/latest/api.html#sqlalchemy_continuum.VersioningManager.track_deletes)(at least as far as the database is concerned)

transfers/owner_transfer.py
```diff
@@ -0,0 +1,95 @@
...
+
+    odf = filter_to_valid_point_ids(session, odf)
+    for i, row in odf.iterrows():
+        thing = session.query(Thing).where(Thing.name == row.PointID).first()
+        if thing is None:
+            print(f"Thing with PointID {row.PointID} not foaund. Skipping owner.")
+            continue
+
+        # TODO: put in guards for null values
+        contact1 = Contact(name=f"{row.FirstName} {row.LastName}", role="Primary")
```
jacob-a-brown, line 36: The roles are often defined in the `OwnerComment` field. It may need to be a manual process to correctly identify all of them. We could, however, look for key words

jacob-a-brown (approved):
I like adding the `# RULE` note for easy reference.

We can using Python's `logging` to record everything. For general print statements we can have the level be `info`, or skipped records it can be `warning`, and for unexpected errors it can be `error`

PR #111, opened 2025-08-30
No TIcket: hot fix

jirhiker:
### **Why**
- ThingResponse needs to include the active_location

### **How**
- updated get_db_things to query Thing and Location simultaneously
- updated ThingResponse

### **Notes**
- Use bullet points here

### DIFF
--- a/core/app.py
+++ b/core/app.py
@@ -18,6 +18,7 @@ from http import HTTPStatus
 from typing import AsyncGenerator
 
 from fastapi import FastAPI
+from sqlalchemy import text
 
 from .dependencies import session_dependency
 from .initializers import init_db, init_lexicon
@@ -45,7 +46,7 @@ app = FastAPI(
 
 @app.get("/_ah/warmup")
 async def warmup(session: session_dependency):
-    session.execute("SELECT 1")
+    session.execute(text("SELECT 1"))
     return HTTPStatus.OK
 
 
--- a/core/app.py
+++ b/core/app.py
@@ -14,10 +14,12 @@
 # limitations under the License.
 # ===============================================================================
 from contextlib import asynccontextmanager
+from http import HTTPStatus
 from typing import AsyncGenerator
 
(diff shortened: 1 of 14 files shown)


schemas/thing.py
```diff
@@ -100,6 +100,7 @@ class BaseThingResponse(ORMBaseModel):
     name: str
     thing_type: str
     release_status: str
+    active_location: LocationResponse | None = None
```
jacob-a-brown, line 103: Will you add `active_location` to the tests in `tests/test_thing.py`?

jacob-a-brown (commented):
I was mistaken in removing `location` from `thing`. It is an associated property of a `thing` like how `emails` and `phones` are associated properties of a `contact`. Should we list out all of the locations? It could be a list that is ordered by `effective_start` and `effective_end`. Or we could have `active_location` and `historic_locations` fields. (Or is that getting ahead of myself and we won't need to create that functionality until it is needed?)

core/app.py
```diff
@@ -40,4 +42,11 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     lifespan=lifespan,
 )
 
+
+@app.get("/_ah/warmup")
+async def warmup(session: session_dependency):
+    session.execute("SELECT 1")
```
jacob-a-brown: sentry raised an issue that this should be `session.execute(text("SELECT 1"))`. The following error was raised:

```
ArgumentError
Textual SQL expression 'SELECT 1' should be explicitly declared as text('SELECT 1')
```

schemas/thing.py
jirhiker: Currently there is no need to return all the locations with a thing, tho we may find it necessary in the future. 
Perhaps we start added an `response_fields` query parameter. That lets the user specify which fields to return.  This is similar to how the ESRI API functions

jacob-a-brown (approved):

PR #117, opened 2025-09-02
NO TICKET: near-universal adoption of `release_status` and `created_at` | complete implementation of `active_location` for the `/thing` router | async coverage for all endpoints

jacob-a-brown:
### **Why**
- All responses should have the `created_at` field
- All responses, except for those from the `/lexicon` router, should have a `release_status` field
- All POST requests, except for those for the `/lexicon` router, should require `release_status` to be filled
- All PATCH requests, except for those for the `/lexicon` router, should allow `release_status` to be edited
- All responses from the `/thing` router should include an `active_location` field
- All endpoints should be `async`

### **How**
- Added `ReleaseMixin` to all database models where it was previously absent, except for `Lexicon` models
- Created `BaseCreateModel`, `BaseUpdateModel`, and `BaseResponseModel` in `/schemas/__init__.py` to enable universal fields to be present for all relevant models
- updated tests for `id`, `created_at`, and `release_status` where appropriate
- updated `tests/test_thing.py` for `active_location`

### **Notes**
- `create_contact` was made more flexible for future model/schema changes

### DIFF
--- a/db/thing.py
+++ b/db/thing.py
@@ -13,11 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
-from xmlrpc.client import Boolean
-
 from sqlalchemy import Integer, ForeignKey, String, Column, Float
 from sqlalchemy.ext.associationproxy import association_proxy
-from sqlalchemy.orm import relationship, mapped_column, Mapped
+from sqlalchemy.orm import relationship, mapped_column
 from sqlalchemy_utils import TSVectorType
 
 from db import lexicon_term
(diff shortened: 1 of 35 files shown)


db/thing.py
```diff
@@ -99,7 +97,7 @@ class ThingIdLink(Base, AutoBaseMixin):
     thing = relationship("Thing", backref="links")
 
 
-class WellScreen(Base, AutoBaseMixin):
+class WellScreen(Base, AutoBaseMixin, ReleaseMixin):
```
jirhiker, line 100: A release status on WellScreen might be a bit excessive but its ok to have for completeness. I wonder tho if we should make it default to `public` in CreateWellScreen

jirhiker (approved):

db/thing.py
jacob-a-brown: I can have it default to `public` in the `CreateWellScreen` schema.

jacob-a-brown: Should that be a no ticket/hot fix? Or just hold off on that for the time being?

jirhiker: no urgency on this, it can wait

PR #119, opened 2025-09-03
BDMS 57: updated contact model, schema, tests, and transfer

jacob-a-brown:
### **Why**
- The fields `organization` and `nma_pk_owners` were missing from Ocotillo

### **How**
- Added `organization` as a nullable string field
- Made `name` nullable so that `organization` can be the primary name of a contact if that's how it's been recorded
- An error is raised if either `organization` or `name` are null when creating or updating a contact
- Added `Company` column to `ownersdata.csv` to more closely mimic NM_Aquifer
- Added `nma_pk_owners` field to contact to tie records back to NM_Aquifer. `nma_pk_owners` is only to be used when transferring data. It is not usable by the API.
- When transferring data from the tables to Ocotillo DB any person without `(first name or last name) and company` are skipped because there's no identifying information.
  -  Should this be the case? Or should the `OwnerKey` become the name?

### **Notes**
- Added clear print statements to `transfers/transfer.py` to help the user know when a transfer function is being called
- Added `transfers` to the beginning of paths so that the transfer script can be called from the root directory as `python -m transfers.transfer`
- Renamed `owners_transfer` to `contacts_transfer` to be consistent in style and naming
- `ownersdata.csv` has `PointID` as a column. In NM_Aquifer there is an OwnerLink table to connect owners to locations. When creating the tables for the transfer will we include `PointID`s in the owners data table, or should I add an OwnerLink table to mimic NM_Aquifer more closely? (and update the transfer script appropriately)

### DIFF
--- a/api/contact.py
+++ b/api/contact.py
@@ -307,6 +307,41 @@ async def update_contact(
     :param session: Database session
     :return: Updated contact response
     """
+    contact = simple_get_by_id(session, Contact, contact_id)
+
+    if (
+        contact.name is None
+        and contact_data.name is None
+        and contact_data.organization is None
+    ):
+        raise PydanticStyleException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=[
+                {
+                    "loc": ["body", "organization"],
+                    "msg": "organization cannot be None if name is None.",
+                    "type": "value_error",
+                    "input": {"organization": contact_data.organization},
+                }
+            ],
+        )
+    elif (
+        contact.organization is None
+        and contact_data.organization is None
+        and contact_data.name is None
+    ):
+        raise PydanticStyleException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=[
+                {
+                    "loc": ["body", "name"],
+                    "msg": "name cannot be None if organization is None.",
+                    "type": "value_error",
+                    "input": {"name": contact_data.name},
+                }
+            ],
+        )
+
     return model_patcher(session, Contact, contact_id, contact_data, user=user)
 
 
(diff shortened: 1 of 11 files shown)


api/contact.py
```diff
@@ -307,6 +307,41 @@ async def update_contact(
...
+                    "type": "value_error",
+                    "input": {"organization": contact_data.organization},
+                }
+            ],
+        )
+    elif (
+        contact.organization is None
+        and contact_data.organization is None
+        and contact_data.name is None
+    ):
```
jirhiker, line 332: this should be 
```python
if contact_data.name is None and contact_data.organization is None:
   if contact.name is None:
      ....
   elif contact.organization is None:
      .... 
```

db/contact.py
```diff
@@ -34,14 +34,18 @@ class ThingContactAssociation(Base, AutoBaseMixin):
...
+    nma_pk_owners = Column(String(100), nullable=True)
 
     phones = relationship("Phone", back_populates="contact", passive_deletes=True)
     emails = relationship("Email", back_populates="contact", passive_deletes=True)
     addresses = relationship("Address", back_populates="contact", passive_deletes=True)
 
-    search_vector = Column(TSVectorType("name", "role"))
+    search_vector = Column(
+        TSVectorType("name", "role", "organization", "nma_pk_owners")
+    )
```
jirhiker, line 48: We should think about whether or not we should make nma_pk_owners searchable

jirhiker (approved):

db/contact.py
jacob-a-brown: That data is encoded in either the `name` or `organization`, so that may be redundant and lead to confusion... It's also not returned in any responses. I'll remove this in a PR. Should `role` be searchable too? That's not really unique to a particular contact (different contacts can have the same roles).

jirhiker: lets leave role it in for now, but i get your point.  Someone probably won't search by the "role" keyword

PR #121, opened 2025-09-04
BDMS 57: well screen & other updates

jacob-a-brown:
### **Why**
- `WellScreen` should have `nma_pk_wellscreens` to enable an audit of the data transfer
- type annotations should not set nullability. That should be explicitly set in `mapped_column(...)`
- Pydantic should be used to validate data

### **How**
- Added `nma_pk_wellscreens` to `WellScreen` table to enable transfer audits
- Used type annotations to set the type, but not nullability. That is now set as `nullable=True` or `nullable=False` in `mapped_column(...)`
- Pydantic is now used to validate contact transfer data
- Because `nma_pk_wellscreens` are not part of the Pydantic model/schema well screens are now added explicitly with the session instead of with `model_adder`

### **Notes**
- The type that's hinted at in `Mapped[<type?>]` is what SQL Alchemy uses. It does not need to be set in `mapped_column(...)`. To set the string length, however, `String(<length>)` must be used in `mapped_column(...)`.

### DIFF
--- a/db/contact.py
+++ b/db/contact.py
@@ -24,10 +24,10 @@ from db.base import Base, AutoBaseMixin, ReleaseMixin, lexicon_term
 
 class ThingContactAssociation(Base, AutoBaseMixin):
     thing_id: Mapped[int] = mapped_column(
-        Integer, ForeignKey("thing.id", ondelete="CASCADE"), nullable=False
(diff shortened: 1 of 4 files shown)


ksmuczynski (approved):
This looks good to me.

jirhiker (approved):

PR #122, opened 2025-09-08
location_model_update_kas

ksmuczynski:
### **Why**
- Missing fields needed to be added to the Ocotillo models
- Models syntax should conform to the latest type-hinting/mapping styles for SQL Alchemy 2.0
- Association proxies should be added. They help manage many-to-many relationships and allow the user to interact with primary model objects and not the intermediary join model. 

### **How**
- Updated:
	- The Location model field syntax to conform to the latest syntax style for SQLAlchemy 2.0
	- The Location model relationship syntax to conform to the latest syntax style for SQLAlchemy 2.0
	- The relationships in the LocationThingAssociation model to reflect the use of proxies
- Added:
	- `thing_associations` relationship to the Location model
	- An association proxy to the Location model.
	- Imported the `association_proxy` function and `AssociationProxy` class so proxies can be implemented
	- (TD) The following fields: `nma_pk_location`, `nma_notes_location`

### **Notes**
- `schema` files were not updated. The only fields that were added were fields whose purpose is to preserve information from NMAquifer.
- `test` files have not been updated. I'm not sure if they need to be.

_Questions_
    - I think the Group model needs a relationship to the Location table. Thoughts?
    - In the `schemas\location.py` file, Should the `effective_start` and `effective_end` fields of the `LocationThingAssociation` be included in the UPDATE schema?
    - I noticed the `Thing` model uses the `overlaps` relationship attribute. Should this be used in the `Location` model, too?
    - I assigned the `nma_pk_location` field a `Mapped[UUID]` type attribute. Should  `nma_pk_tablename` fields be assigned `Mapped[str]` or `Mapped[UUID]` type attributes? I think NMAquifer stored primary keys as GUIDs.

### DIFF
--- a/db/location.py
+++ b/db/location.py
@@ -26,17 +26,17 @@ from sqlalchemy import (
     DateTime,
     func,
     Text,
+    UUID,
 )
 from sqlalchemy.orm import relationship, Mapped, mapped_column
 from sqlalchemy.ext.associationproxy import association_proxy, AssociationProxy
+
+from db import Thing
 from db.base import Base, AutoBaseMixin, ReleaseMixin
 from db.lexicon import lexicon_term
 
 
 class Location(Base, AutoBaseMixin, ReleaseMixin):
-    # name = Column(String(100), nullable=True)
-    # description = Column(String(255), nullable=True)
-    # visible = Column(Boolean, default=False, nullable=False)
     __versioned__ = {}
 
     nma_pk_location: Mapped[UUID] = mapped_column(
--- a/db/location.py
+++ b/db/location.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
+import datetime
+
 from geoalchemy2 import Geometry, WKBElement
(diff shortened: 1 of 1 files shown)


db/location.py
```diff
@@ -36,15 +39,30 @@ class Location(Base, AutoBaseMixin, ReleaseMixin):
     # visible = Column(Boolean, default=False, nullable=False)
```
jirhiker: `visible` can be removed

db/location.py
```diff
@@ -36,15 +39,30 @@ class Location(Base, AutoBaseMixin, ReleaseMixin):
...
     __versioned__ = {}
 
-    name = mapped_column(String(255), nullable=True)
-    notes = mapped_column(Text, nullable=True)
+    nma_pk_location: Mapped[UUID] = mapped_column(
```
jirhiker, line 42: need to import UUID

jirhiker (changes requested):
tests failing.  need to import UUID.

jirhiker (approved):

PR #123, opened 2025-09-09
location_model_update_kas

ksmuczynski:
### **Why**
- Fields need to be added to the database model and associated files so that legacy data from NMAquifer can be migrated effectively and new data can be stored appropriately.

### **How**
- Added new fields to the Location database model.
- Updated the CREATE, RESPONSE, and UPDATE schema models to reflect the addition of new fields in the database model
- Updated the POST, PATCH, and GET tests to reflect the addition of new fields
- Updated the `util.py` transfer script file to include the new fields.

### **Notes**
- TODO: Create functions so that the `state`, `county`, and `quad_name` fields are auto-populated by the API.
- TODO: confirm whether PointID should map to `location.name` or `thing.name`


db/location.py
```diff
@@ -53,6 +53,12 @@ class Location(Base, AutoBaseMixin, ReleaseMixin):
     quad_name: Mapped[str] = mapped_column(String(100), nullable=True)
     notes: Mapped[str] = mapped_column(Text, nullable=True)
     nma_notes_location: Mapped[str] = mapped_column(Text, nullable=True)
+    nma_coordinate_notes: Mapped[str] = mapped_column(Text, nullable=True)
+    elevation_accuracy: Mapped[float] = mapped_column(nullable=True)
+    elevation_method: Mapped[str] = lexicon_term(nullable=True)
+    coordinate_accuracy: Mapped[float] = mapped_column(nullable=True)
+    coordinate_accuracy_unit: Mapped[str] = lexicon_term(nullable=True)
```
jacob-a-brown: Will all `elevation_accuracy` values be in meters? If so, should `elevation_accuracy` be renamed `elevation_accuracy_m`? Or, should we add a field called `elevation_accuracy_unit`?

Same for `coordinate_accuracy`: will it always be in degrees since we are storing coordinate values in WGS84.

schemas/location.py
```diff
@@ -38,6 +38,11 @@ class CreateLocation(BaseCreateModel):
     notes: str | None = None
     point: str  # point is required and should be in WKT format
     release_status: str | None = "draft"
+    elevation_accuracy: float | None = None
+    elevation_method: str | None = None
+    coordinate_accuracy: float | None = None
+    coordinate_accuracy_unit: str | None = None
```
jacob-a-brown: If we decide to keep a `coordiante_accuracy_unit` field should this always be `degree`?

transfers/util.py
```diff
@@ -207,15 +207,18 @@ def make_location(row: pd.Series) -> Location:
...
-        # elevation_method=row.AltitudeMethod,
+        elevation_accuracy=row.AltitudeAccuracy,
+        elevation_method=row.AltitudeMethod,
         # created_at=created_at,
-        # point_accuracy=row.CoordinateAccuracy,
-        # point_method=row.CoordinateMethod,
+        coordinate_accuracy=row.CoordinateAccuracy,
+        coordinate_method=row.CoordinateMethod,
+        nma_coordinate_notes=row.CoordinateNotes,
+        nma_notes_location=row.LocationNotes,
```
jacob-a-brown, line 221: I can update this when I merge your work with the lexicon updates, but now that we are not using lookup table codes I think we'll need to amend this section (and for other tables/fields that are lexicon/lookup tables) to get the `MEANING` from the LU tables. Otherwise they won't be found in the lexicon term table and we'll run into errors.

On that note, I'll wait to open a PR for lexicon so that I can take care of that in the transfer scripts where it is relevant.

db/location.py
ksmuczynski: According to the NMAquifer documentation, elevation accuracy is in units of feet. I don't think we need a separate units field here. We can either append "_ft" to the field name or just note the units in the documentation describing the field. I'm kind of inclined to do both.

jirhiker: I generally prefer to avoid `Hungarian Notation` i.e.  `elevation_accuracy_m`

jacob-a-brown: As long as it's documented somewhere that should be okay. If the units are always going to be the same we don't need a separate `..._unit` table and can note it. Then in the response schema we can return a units field that evaluates to the correct unit.

Should the `elevation_accuracy` always be `meters` because elevation is stored in `meters`? And `coordinate_accuracy` should always be in `decimal degrees` because lat/long are stored in `decimal degrees`?

jirhiker: that seems logical

schemas/location.py
ksmuczynski: The reason I added a units field for coordinates is because the NMAquifer lookup table, LU_CoordinateAccuracy, has units of second, minute, and meters. I thought a one-to-one mapping would be the fastest way to migrate the data,  but we could convert coordinate accuracy units into a standard unit during the migration. From what I've read, it seems best practice and industry standard is to store coordinate accuracy as a single, consistent linear unit, with **meters** being favored.

Converting arc-seconds to linear units is a little tricky because the linear distance of a longitude
arc-second depends on the latitude, so lat/long will need to be a variable in the conversion. Just something to be aware of.

db/location.py
ksmuczynski: From what I've read, best practice and industry standard is to store coordinate accuracy as a single, consistent linear unit, with meters being favored.

ksmuczynski: Elevation in NMAquifer is stored in feet, so the elevation accuracy should also be stored in feet.

jacob-a-brown: In the transfer script we can convert elevation accuracy units from ft to meters

As for coordinate accuracy I'm okay to store that in meters, too, if that's the standard

schemas/location.py
jacob-a-brown: I'll write a function for the second/minute conversion to meters and be sure to highlight it in the transfer PR for review. Hopefully Python has a function or library to do this already, but that may not be the case 🫰

jirhiker (approved):

PR #124, opened 2025-09-10
NO-TICKET: Internal and Public API specs

jirhiker:
### **Why**
- Ocotillo API needs to serve both Ocotillo UI for internal users and a public api for anonymous external users.  The public api should be a subset of the total api. (maybe just a subset of the GET routes)

### **How**
- Added `/docs-auth` and `/docs` routes.  
- `/docs` returns the public api spec `/docs-auth` returns the full api
- added the `public_route` decorator to indicated which routes should be included in the public openapi spec

### **Notes**
- I tried and struggled to make the `/docs-auth` only visible once the user has logged in, but went in circles with AI trying to accomplish a reasonable UX.  This should be revisited at a later date

### DIFF
--- a/api/thing.py
+++ b/api/thing.py
@@ -25,6 +25,7 @@ from starlette.status import (
 )
 
 from api.pagination import CustomPage
+from core.app import public_route
 from core.dependencies import (
     session_dependency,
     amp_admin_dependency,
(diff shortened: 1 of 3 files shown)


jacob-a-brown (approved):

chasetmartin:
@jirhiker So does this mean that the any script the front end has to retrieve the openapi.json spec will have to send an auth token to get the full non-public spec?

jirhiker:
In the current implementation, no.  both specs are publically available, its only the routes that are protected and need auth to use.

PR #125, opened 2025-09-10
BDMS 121: log transfers

jacob-a-brown:
### **Why**
- Transfer processes should be logged and persisted for understanding and audits

### **How**
- Replaced all print statements with Python's logger

### **Notes**
- I'm currently using `info` for print statements and `warning` for when something goes wrong. Should we differentiate between `warning` and `error`?

### DIFF
--- a/transfers/transfer.py
+++ b/transfers/transfer.py
@@ -32,6 +32,7 @@ from transfers.thing_transfer import (
     transfer_ephemeral_stream,
     transfer_met,
 )
+from transfers.util import logger
 
 
 def erase_and_initalize(session: Session) -> None:
@@ -43,6 +44,9 @@ def erase_and_initalize(session: Session) -> None:
 
 
 def main_transfer():
+    logger.log("")
+    logger.log("Starting transfer\n")
+
     init = True
 
     transfer_well_flag = False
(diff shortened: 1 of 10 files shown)


transfers/util.py
```diff
@@ -26,6 +26,16 @@
 
 from db import Thing, Location
```
jirhiker, line 29: Id like a new log file every time `main_transfer` is run instead of a single log file that is appended

jirhiker (changes requested):

transfers/util.py
jacob-a-brown: new log files are now created every time. they have a timestamp transfer appended to the name

e.g.

`transfer_2025-09-10T12h30m42s.log`

transfers/transfer.py
```diff
@@ -60,51 +64,70 @@ def main_transfer():
 
     limit = 100
     with session_ctx() as sess:
-
         if init:
             erase_and_initalize(sess)
 
         if init or transfer_well_flag:
-            print("\n", "*" * 10, "TRANSFERRING WELLS", "*" * 10)
+            msg = "*" * 10 + "TRANSFERRING WELLS" + "*" * 10
```
jirhiker, line 71:
```python
message("TRANSFERRING WELLS")

def message(msg, pad='*', pad_len=10):
     pad *= pad_len
     logger.info(f"{pad} {msg} {pad}")

```

jirhiker (changes requested):
Is this PR up to date with the `transfer` branch?

transfers/transfer.py
jacob-a-brown: I didn't know of the `transfer` branch... I've just been working off of `pre-production` assuming that was the base for all of the project. Is that `message` function the only update? If so, I can also merge `transfer` into my branch, make updates/resolve conflicts, and push back to this PR

jirhiker: the other updates are to allow transfers to run via Cloud Run

jirhiker (approved):

PR #126, opened 2025-09-10
BDMS 122: lexicon updates & other updates

jacob-a-brown:
### **Why**
- The lexicon table should contain all NM_Aquifer lookup table values so that data can be transfered

### **How**
- Added values to `core/lexicon.json`

### **Notes**
- Used full names except for organizations since they are commonly referred to with their abbreviations, like USGS
- The transfer scripts will need to be updated to use code/meaning mappings so the transferred data will be in the lexicon term table
- Added some verbiage to the README to help with future model revisions


core/lexicon.json
```diff
@@ -176,5 +454,407 @@
...
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Field blank", "definition": "Field blank"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Field duplicate", "definition": "Field duplicate"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Field parameters only", "definition": "Field parameters only"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Precipitation", "definition": "Precipitation"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Repeat sample", "definition": "Repeat sample"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Standard field sample", "definition": "Standard field sample"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Soil or Rock sample", "definition": "Soil or Rock sample"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Trip blank", "definition": "Trip blank"},
+  {"categories": [{"name": "sample_type", "description": null}], "term": "Source water blank", "definition": "Source water blank"}
 ]
```
jirhiker, line 487: I worry about transferring the `formation` and `lithology` fields without doing some cleanup.

For instance, is there a distinction between `Sand, silt and gravel` and `Gravel, sand and silt`, or would one record suffice? Doesn't the USGS have a controlled vocabulary for lithology?

Formation records seem a mess

jacob-a-brown: To facilitate the transfer of data we can keep all of the lookup table values as-is, then we can work to clean up those terms. Since a `lexicon_term` has `onupdate="CASCADE"` any changes should propagate to the appropriate places. I could also remove `formation` and `lithology` from the lexicon, but I'd be concerned that would inhibit data transfers. (If we wait to record lexicon terms for `formation` and `lithology` we could just wait to transfer all `formation` and `lithology` values until it's cleaned up and mapped to the new terms.)

ksmuczynski (commented):
Looks good. I agree that there are several lookup tables that could benefit from a clean up. Modifying the lexicon terms later doesn't seem like a big deal, but what about cases where the cleanup is significant and lexicon terms are removed (but the removed terms are referenced in other tables? Should we set some kind of an ON DELETE constraint or flag it in some other way? Is it important to figure out now or later?

jacob-a-brown:
That's a good point, particularly if some are duplicative and we decide to just keep one (e.g. `sand, silt, gravel` and `gravel, sand, silt`).

So remove them for now?

jirhiker:
Lets remove for now. We need AMP to review and cleanup.  This could be something Kimball could help with

jirhiker (approved):

PR #136, opened 2025-09-12
BDMS 122: map lookup table codes to lexicon terms

jacob-a-brown:
### **Why**
- The new ocotillo database uses the full terms instead of the lookup table codes. If we just transfer the codes (which are present in the various tables like `Location` and `WellData`) then foreign key errors will be raised. This mapper enables the transfer scripts to transfer the correct data.

### **How**
- Read lookup table csvs from bucket and map mapper
 
### **Notes**
- Use bullet points here


jacob-a-brown:
Many LU tables share the same code (e.g. `M`). To allow the same code from multiple tables to be used the table name has been prepended to the key as `{table name}:{code}`.

jacob-a-brown:
These updates will be in a future PR for location transfer revision

PR #137, opened 2025-09-15
BDMS 111: location updates & other

jacob-a-brown:
### **Why**
- tests should be run on PRs for update branch names (`production` and `staging`)
- updates to the location model needed to be incorporated into the transfer script
- `elevation` needed to be separated from `point` for a location
- datetime records are stored in UTC
- lookup table codes/values needed to be transferable and interoperable with lexicon terms

### **How**
- Updated transfer script for new fields
- Converted elevation to meters if exists, else retrieved from EPQS
- queried outside sources for a location's `state`, `county`, and `quand_name` fields using a location's `point` x/y values
- made mapper from a lookup table's code to meaning so that the values can be transferred and interoperable with lexicon terms. because many codes are the same between tables they are prepended with the table name to ensure uniqueness (e.g. `LU_CoordinateAccuracy:M` is a key in the table as well as `LU_SiteType:M`)
- In data transfers all datetime values are converted to UTC. If a time is midnight (`12:00:00 AM` or `00:00:00`) the time doesn't change but the time zone is set to UTC

### **Notes**
- `name` was removed from `Location` until further consideration and debate
- many utilities used in `transfers/util.py` are applicable and used throughout the API, so they were moved to `services/util.py` for repo-wide access
- updated SRID to use constants from `constants.py` for style
- NM_Aquifer CoordinateAccuracy values are not yet transferred. AMP needs to be consulted before this transfer is made.


api/geospatial.py
```diff
@@ -100,7 +100,9 @@ def get_feature_collection(
 
     things = get_thing_features(session, thing_type, group)
 
-    def make_feature_dict(thing, geometry, *other):
+    def make_feature_dict(thing, geometry, elevation, *other):
```
jirhiker, line 103: add anotations

db/group.py
```diff
@@ -28,7 +29,7 @@ class Group(Base, AutoBaseMixin, ReleaseMixin):
     description: Mapped[str] = mapped_column(String(255), nullable=True)
     name: Mapped[str] = mapped_column(String(100), nullable=False, unique=True)
     project_area: Mapped[Optional[WKBElement]] = mapped_column(
```
jirhiker, line 31: rename to region_of_interest (?)

jirhiker (approved):

PR #138, opened 2025-09-16
statushistory_model

ksmuczynski:
### **Why**
- Need a central, polymorphic log for all time-variant operational statuses (e.g., Use Status, Access Status)

### **How**
- New `status_history.py` file was created to establish relevant fields.
- A new `HasStatusHistory` mixin was added to the `db/base.py` file.
- Unrelated to the Jira ticket, docstrings were added to the `db/base.py` file to summarize the contents of the file and describe the mixin functionality.

### **Notes**
- The `HasStatusHistory` mixin has not yet been added to/inherited by any primary or core tables.

### DIFF
--- a/db/base.py
+++ b/db/base.py
@@ -13,6 +13,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
+"""
+db/base.py
+
+This file defines the foundational components for the SQLAlchemy models.
+It includes:
+1.  The declarative base class (`Base`) that all models will inherit from.
+2.  A helper function (`lexicon_term`) to create standardized foreign key columns
+    referencing the `lexicon_term` table.
+3. A helper function (`pascal_to_snake`) to convert class names from PascalCase to snake_case
+    for automatic table naming.
+4. Mixins for common functionality:
+    - `AutoBaseMixin`: Adds automatic table naming and an auto-incrementing primary key.
+    - `PropertiesMixin`: Adds a JSONB properties column for storing additional attributes.
+    - `ReleaseMixin`: Adds a release status column referencing the `lexicon_term` table.
+    - `AuditMixin`: Adds standard audit columns (created_at, created_by, updated_at, updated_by).
+5.  A simple `User` model for tracking user information in audit columns.
+6.  Polymorphic helper mixins (`HasStatusHistory`, `HasNotes`, `HasAttribution`, etc.)
+    which provide a clean, reusable way to add relationships to the polymorphic
+    metadata tables. Any model that can have a status history (like Thing or Location)
+    can simply inherit from the `HasStatusHistory` mixin.
+7.  An `AuditMixin` to add standard audit columns to tables.
+"""
+
 from sqlalchemy import (
     Column,
     DateTime,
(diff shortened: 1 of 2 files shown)


db/status_history.py
```diff
@@ -0,0 +1,38 @@
...
+    start_date: Mapped[datetime.datetime] = mapped_column(
+        DateTime(timezone=True), nullable=True
+    )
+    end_date: Mapped[datetime.datetime] = mapped_column(
+        DateTime(timezone=True), nullable=True
+    )
+    reason: Mapped[str] = mapped_column(Text, nullable=True)
+
+    # Polymorphic relationship columns
+    statusable_id: Mapped[int] = mapped_column(Integer, nullable=False)
```
jacob-a-brown, line 37: why have this column? can we just use the `id` field that comes from `AutoBaseMixin`?

db/base.py
```diff
@@ -129,7 +171,29 @@ def properties(self):
         )
 
 
+# ============= Polymorphic Helper Mixins =============================================
+class HasStatusHistory:
```
jacob-a-brown: To keep the style consistent should this be called `StatusHistoryMixin`?

db/base.py
```diff
@@ -129,7 +171,29 @@ def properties(self):
...
+
+    @declared_attr
+    def status_history(self):
+        # One-to-Many polymorphic relationship
+        return relationship(
+            "StatusHistory",
+            primaryjoin=f"and_({self.__name__}.{self.__name__.lower()}_id==StatusHistory.statusable_id, "
+            f"StatusHistory.statusable_type=='{self.__name__}')",
+            cascade="all, delete-orphan",
+            lazy="selectin",
```
jacob-a-brown, line 190: Is this achievable via a foreign key? Or does the `StatusHistory` table store the `id` from the mixed-in table (e.g. `thing`), which is why this type of join is necessary? that is, the `id` from `StatusHistory` is not stored in the mixed-in table

jacob-a-brown (commented):
some questions about the `StatusHistory` tab and mixin. I like the documentation you wrote a lot 👍

db/status_history.py
ksmuczynski: The `statusable_id` is the ID of the parent record this status belongs to (e.g., the thing_id or location_id). The table will still have an `id` primary key field from the AutoBaseMixin.

db/base.py
ksmuczynski: Yes, the `id` from the core table (thing, location, etc.) is being stored here in the `statusable_id` field. Here's an example of what it might look like.
| **status_history_id** | **statusable_id** | **statusable_type** | **status_type** | **status_value** | **start_date** | **end_date** | **reason**                         |
| --------------------- | ----------------- | ------------------- | --------------- | ---------------- | -------------- | ------------ | ---------------------------------- |
| 1                     | 102               | **'Thing'**         | Use Status      | Active           | 2018-01-15     | NULL         | Installed for regional study.      |
| 2                     | 102               | **'Thing'**         | Well Integrity  | Good             | 2018-01-15     | 2025-06-15   | Post-installation inspection.      |
| 3                     | 25                | **'Location'**      | Access Status   | Accessible       | 2018-01-15     | 2025-07-20   |                                    |
| 4                     | 102               | **'Thing'**         | Well Integrity  | Damaged          | 2025-06-16     | 2025-09-01   | Casing cracked during storm event. |

jacob-a-brown (approved):

jirhiker (approved):

PR #140, opened 2025-09-19
permission_model

ksmuczynski:
### **Why**
- Need a polymorphic table to manage and historicize access and usage agreements.

### **How**
- Created new Permimssion model with new fields.
- Added new PermissionMixin to db/base/py file to automatically create a polymorphic One-to-Many relationship to the
Permission table.

### **Notes**
- This PR address the creation of the Permission table only. 
- Related validation schema files and test files still need to be created. 
- Core tables that will inherit from the Permission table still need to be updated (Thing, Location)

### DIFF
--- a/db/permission.py
+++ b/db/permission.py
@@ -1,0 +1,85 @@
+"""
+models/permission.py
+
+This model defines the `Permission` table, a polymorphic table that tracks
+all legal and administrative agreements related to site access and activity.
+Its purpose is to track who granted permission, what activities they authorized,
+which entity the permission applies to, and for what period of time.
+"""
+
+from typing import TYPE_CHECKING
+
+from sqlalchemy import (
+    Integer,
+    ForeignKey,
+    String,
+    Boolean,
+    Date,
+    Text,
+    and_,
+)
+from sqlalchemy.orm import relationship, Mapped, mapped_column, foreign
+
+from db.base import Base, AutoBaseMixin, ReleaseMixin
+
+if TYPE_CHECKING:
+    from db.contact import Contact
+    from db.thing import Thing
+    from db.location import Location
+
+
+class Permission(Base, AutoBaseMixin, ReleaseMixin):
+    """
+    Represents a specific grant of permission from a Contact for a
+    specific entity (e.g., a Thing or Location).
+    """
+
+    # --- Foreign Keys ---
+    contact_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("contact.contact_id"), nullable=False
+    )
+
+    # --- Columns ---
+    allow_sampling: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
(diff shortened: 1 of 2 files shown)


db/permission.py
```diff
@@ -0,0 +1,85 @@
...
+        Integer, ForeignKey("contact.contact_id"), nullable=False
+    )
+
+    # --- Columns ---
+    allow_sampling: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
+    allow_installation: Mapped[bool] = mapped_column(
+        Boolean, nullable=False, default=False
+    )
+    start_date: Mapped[Date] = mapped_column(Date, nullable=True)
+    end_date: Mapped[Date] = mapped_column(Date, nullable=True)
```
jacob-a-brown, line 49: Should these be timezone datetime? that's used elsewhere when we may only want a date. If using `Date` here, we should change them in those other places to plane old `Date`

db/permission.py
```diff
@@ -0,0 +1,85 @@
...
+    notes: Mapped[str] = mapped_column(Text, nullable=True)
```
jacob-a-brown, line 50: Would this relate back to the polymorphic notes table? rather than have a `notes` field

db/permission.py
```diff
@@ -0,0 +1,85 @@
...
+        viewonly=True,
+    )
+    _location_target: Mapped["Location"] = relationship(
+        "Location",
+        primaryjoin=and_(
+            foreign(permissible_id) == Location.location_id,
+            permissible_type == "Location",
+        ),
+        viewonly=True,
+    )
```
jacob-a-brown, line 77: For this to work do the `thing` and `location` tables need to inherit from the permission mixin?

jacob-a-brown: Wait! That's been noted in your notes. Should it be present in this PR? Or wait to implement that so that it can be implemented elsewhere, too, like services and tests?

db/permission.py
ksmuczynski: Yes, you're right. Will update to datetime.

db/permission.py
ksmuczynski: I like the idea of a polymorphic notes table, but thought it might be easier to implement single note fields in individual tables to start. Later we could go back and implement the polymorphic notes table, but I'm open to discussion. If we have the time I think it could be worth implementing now. We could still retain NMAquifer notes in `nma_notes` fields where appropriate.

db/permission.py
ksmuczynski: Right. The idea with this PR is to just create the table. Then a separate PR for Thing and Location updates, where they would inherit these new tables. It seemed simpler to break down the PRs into their individual component tasks and focus on one thing at a time. Personal opinion of course, open to discussion re: best practices for PRs.

jirhiker (approved):

PR #141, opened 2025-09-19
BDMS 113 & 114: sample & observation updates & revisions | other updates and revisions

jacob-a-brown:
# !!!ALERT!!! There is a lot in this PR. There have been a number of big and consequential design choices and changes to the database and therefore the API. Please reach out to me here, Slack, or elsewhere with any questions or if you need clarification.

### **Why**
- The `Sample` model was determined to be non-normalized, so it was split into `FieldEvent`, `FieldActivity`, and `Sample`
- AMP colleagues want to know all personnel who were present at a particular `FieldEvent`
- lexicon terms should be plain-text and not contain extra information that ought to be retrieved from the database and subsequently used by the API
- geothermal observations/samples are not ready to be implemented

### **How**
- The `Sample` model was split into `FieldEvent`, `FieldActivity`, and `Sample` to better capture when/where/how a sample is created, as well as who was present and who took the sample
  - AMP colleagues specifically asked that the collecting organization be recorded, which is occurring in the `FieldEvent` table
  - AMP colleagues specifically asked that all present at the field event, even if they are not the event leader nor a participant, be recorded. @ksmuczynski and I spoke through the best way to do this and we determined that using the `contact` table would be appropriate. It has all of the requisite information for people. When this gets implemented on the frontend we can add an `organization` query parameter to GET `/contact` for a dropdown menu of eligible participants (if we want to filter by organization. I can definitely add other query parameters, too).
- Because the `activity_type` is stored in the `FieldActivity` table that is being used when retrieving/patching different observations rather than prepending the observation class to the observed property.
- the contacts that are associated with a sample are now restricted to those present in the many-to-many relationship between contact and field event
- all references to geothermal have been deprecated until we can get feedback from colleagues. this is not currently used and has been inhibiting development.

### **Notes**
- @jirhiker and @chasetmartin because of the significant database revisions @ksmuczynski should provide feedback/approval for this PR before being merged into `staging`.
- this PR does *not* include the full updates to the transfer scripts. It's mostly there, but now that someone in the `contact` table *must* be associated with a `sample` I'm not too sure how to proceed.
  - Should `field_event_contact_id` be nullable in the `sample` table so that the records can be transferred to the new database? This can be updated to be non-nullable in the future, at which point we'd need to figure out how to correctly associate old records. Perhaps put all of them into the `contact` table and have some sort of note saying "legacy data from NM_Aquifer"? Or another flag?
  - Should that field remain non-nullable and then the `MeasuredBy` field in the `WaterLevels` table will be transferred to `Contacts` before creating water levels, thereby not breaking the foreign key constraint?
    - What do we do if the `MeasuredBy` field in the `WaterLevels` table from `NM_Aquifer` is `NULL`? Have a contact named `NM Aquifer NULL`?
- to restrict contact associations to a sample (ie the sampler) to those that participated in the field event we use the `FieldEventContactAssociation` table. Will this cause issues with the frontend?
  - This association table will be populated when a new `FieldEvent` is created. I plan to allow the user to submit multiple contacts that associate with a field event. After the `FieldEvent` record is created, and therefore the `FieldEventContactAssociation` recirds are created, they will be in the database and useable when creating/editing a `sample` record.
- the `/field` router does not yet exist, and therefore the `Field...` are not fully covered (if at all). This will come in a future PR. I wanted to open this PR, though, to get the updated models into staging (`sample`/`observation` in particular). `/field` coverage will be my next task.


api/sample.py
```diff
@@ -47,23 +54,23 @@ def database_error_handler(
     error_message = error.orig.args[0]["M"]
     if (
         error_message
-        == 'duplicate key value violates unique constraint "sample_field_sample_id_key"'
+        == 'duplicate key value violates unique constraint "sample_sample_name_key"'
```
jirhiker, line 57: I worry about the fragility of this. If the error text ever changes even slightly this breaks. No action needed now but let's consider ways to improve the database error handler

jirhiker (dismissed):

jirhiker:
Looks good to me.   @jacob-a-brown @ksmuczynski @chasetmartin waiting on additional reviews before merging.

jirhiker (approved):

chasetmartin (approved):
This looks good to me. I think the implementation on the front will take some user research/testing to make sure creating a sample entry is straightforward and doesn't have too many steps. We can talk about the best approach once simple admin table CRUD is handled.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+        comment="Foreign key to the FieldEvent table.",
+    )
+    contact_id: Mapped[str] = mapped_column(
+        ForeignKey("contact.id", ondelete="CASCADE"),
+        nullable=False,
+        comment="Foreign key to the Contact table",
+    )
+
+    # TODO: get AMP feedback on the roles
+    field_event_role: Mapped[str] = lexicon_term(
```
ksmuczynski: Consider renaming to `field_contact_role`. Role describes the contact, not the field event.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+    # --- Relationships ---
+    field_event: Mapped["FieldEvent"] = relationship("FieldEvent")
+    contact: Mapped["Contact"] = relationship("Contact")  # noqa: F821
+
+    # map associated contacts to samples to restrict the people who could have
+    # taken a sample to those present at the field event
+    samples: Mapped[list["Sample"]] = relationship(  # noqa: F821
+        "Sample",
+        back_populates="field_event_contact",
+        cascade="all, delete-orphan",
```
ksmuczynski: Remove this cascade parameter. The data stored in the `FieldEventContactAssociation` table is essentially administrative metadata. Deleting an admin record should never automatically trigger the deletion of core scientific data.

For example, say someone notices Ethan Mamer is listed as a field participant during an event he wasn't at and they delete the record. If `cascade="all, delete-orphan" ` were active, the database would say, "Okay, Ethan Mamer's participation record is gone. Now find and delete every Sample record that he is listed as the collector for." The database would delete all associated  `Sample`  records and then, because the `Sample` table should have a cascade to `Observation` table, the database would delete all the lab results associated with those samples.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+
+    # TODO: get AMP feedback on the roles
+    field_event_role: Mapped[str] = lexicon_term(
+        nullable=False, comment="Role of the contact in the field event"
+    )
+
+    # --- Relationships ---
+    field_event: Mapped["FieldEvent"] = relationship("FieldEvent")
+    contact: Mapped["Contact"] = relationship("Contact")  # noqa: F821
```
ksmuczynski: Add `back_populates="field_event_contact"` to both of these relationships. Needed to ensure a robust, two-way relationship and help make sure the Python objects are always in sync.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # map associated contacts to samples to restrict the people who could have
+    # taken a sample to those present at the field event
+    samples: Mapped[list["Sample"]] = relationship(  # noqa: F821
+        "Sample",
+        back_populates="field_event_contact",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
```
ksmuczynski, line 46: Can prob remove this, too, since the cascade will be removed.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+class FieldEvent(Base, AutoBaseMixin, ReleaseMixin):
+    """
+    This table serves as the master log for all field visits. Each
+    record in this table represents a single, continuous collection event at a
+    specific Thing (e.g., a well) by a specific person on a specific date.
+
+    This table's purpose is to store event-level metadata that is true for the
+    entire visit, such as the date, time, and the person responsible. It acts as
+    the parent container for all activities performed and all samples collected
+    during that single visit.
```
ksmuczynski, line 59: Would re-word this to reflect recent changes. Maybe something like,

> This table serves as the master log for all field visits. Each record represents a single, continuous collection event at a specific Thing (e.g., a well). It is the parent container for all activities, samples, and participants associated with that visit.

> Its purpose is to store the "what, where, and when" of the event. Information about _who_ participated is managed in the `FieldEventContactAssociation` table.

api/sample.py
jacob-a-brown: I see what you mean - it's another place where we need to keep track of names. Instead of having separate if statements for each potential error, we can keep track of all constraints - most often `foreign key` and `unique` - and then catch those errors if they happen and return appropriate messages by decoding the error message.

db/field.py
jacob-a-brown: I changed it slightly to include information about the `FieldActivity` and `Sample` tables. Does this look okay?

> Its purpose is to store the "where and when" of the event.
    Information about who participated is managed in the
    FieldEventContactAssociation table. Information about the "what" of the
    event is managed in the FieldActivity and Sample tables.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+    # Columns
+    event_date: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True),
+        nullable=False,
+        comment="Date and time of the field event.",
+    )
+    collecting_organization: Mapped[str] = lexicon_term(
+        nullable=False,
+        default="NMBGMR",  # TODO: put default in schema
+        comment="The organization that is collecting and storing the samples from the field event",
```
ksmuczynski: Regarding the text, "and storing": are we certain that the organization that collects the sample also stores its associated data? There are records in NMAquifer indicating another organization (e.g., USGS) collected the sample, but NMBGMR is storing the data.

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+    notes: Mapped[str] = mapped_column(
+        nullable=True,
+        comment="Notes or comments about the field event.",
+    )
+    # --- Relationships ---
+    thing: Mapped["Thing"] = relationship(back_populates="field_events")  # noqa: F821
+    field_activities: Mapped[list["FieldActivity"]] = relationship(
+        back_populates="field_event"
+    )
```
ksmuczynski, line 88: Think you might be missing the table name after relationship. Should be relationship("FieldActivity"), back_populates="field_event")

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+        nullable=False,
+        comment="Foreign key to the FieldEvent table.",
+    )
+
+    # Columns
+    activity_type: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="The type of activity performed during the field event (e.g., 'groundwater level', 'water chemistry', 'geothermal').",
+    )
+    notes: Mapped[str] = mapped_column(
```
ksmuczynski, line 132: Think the mapped_column is missing the data type parameter. Should be, `mapped_column(Text, nullable=True, comment="Notes or comments about the field activity",)`

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+        nullable=True,
+        comment="Notes or comments about the field activity.",
+    )
+
+    # Relationships
+    field_event: Mapped["FieldEvent"] = relationship(back_populates="field_activities")
```
ksmuczynski: Doesn't the table name need to be included in the relationship definition? Think it should be: `field_event: Mapped["FieldEvent"] = relationship("FieldEvent", back_populates="field_activities")`

db/field.py
```diff
@@ -0,0 +1,137 @@
...
+    samples: Mapped[list["Sample"]] = relationship(  # noqa: F821
+        back_populates="field_activity"
+    )
```
ksmuczynski, line 146: Add a `cascade="all, delete-orphan"` here as `FieldActivity` to `Sample` is a direct parent-child (or cause-and-effect) relationship.

db/observation.py
```diff
@@ -13,61 +13,58 @@
...
 from db.base import Base, AutoBaseMixin, ReleaseMixin, lexicon_term
 
 
 class Observation(Base, AutoBaseMixin, ReleaseMixin):
     __versioned__ = {}
 
-    sample_id = mapped_column(
-        Integer,
+    # NM_Aquifer fields for audits
+    nma_pk_waterlevels: Mapped[str] = mapped_column(nullable=True)
```
ksmuczynski, line 30: I think the primary key for NMAquifer WaterLevels table is of type UUID. Confirm and update data type from `str` to `UUID` if necessary.

db/observation.py
```diff
@@ -13,61 +13,58 @@
...
+    observation_depth: Mapped[float] = mapped_column(
         nullable=True,
         info={"unit": "feet"},
         doc="Depth of the geothermal observation in feet",
     )
 
-    sensor = relationship("Sensor")
-    sample = relationship("Sample")
+    sensor: Mapped["Sensor"] = relationship("Sensor")  # noqa: F821
+    sample: Mapped["Sample"] = relationship("Sample")  # noqa: F821
```
ksmuczynski: Add `back_populates="observations"` to these relationships. Because these relationships are bidirectional, adding `back_populates` ensures objects on both sides of the relationship are kept in sync.

ksmuczynski: Add `passive_deletes=True` here. Passive_deletes must be True for the ondelete="CASCADE" (which is set on the foreign key) to work.

db/sample.py
```diff
@@ -28,102 +28,80 @@
...
+    # nullable because sample can be collected by steel tape
     sensor_id: Mapped[Optional[int]] = mapped_column(
         ForeignKey("sensor.id"),
         comment="Foreign key for the specific equipment used.",
+        nullable=True,
     )
 
-    # Sample Attributes
+    field_event_contact_id: Mapped[Optional[str]] = mapped_column(
+        ForeignKey("field_event_contact_association.id")
```
ksmuczynski: Should `nullable=True` be set here? I know it says 'Optional,' but I thought we agreed to also be explicit.

db/sample.py
```diff
@@ -28,102 +28,80 @@
...
+    # --- Foreign Key Definitions ---
+    field_activity_id: Mapped[int] = mapped_column(
+        ForeignKey("field_activity.id"), nullable=False
     )
+
+    # nullable because sample can be collected by steel tape
     sensor_id: Mapped[Optional[int]] = mapped_column(
         ForeignKey("sensor.id"),
         comment="Foreign key for the specific equipment used.",
+        nullable=True,
```
ksmuczynski: I don't think this foreign key is needed at the Sample level. Information regarding the equipment used to collect the sample can be captured more accurately at the `Deployment` and/or `Observation` levels.

db/sample.py
```diff
@@ -28,102 +28,80 @@
...
-    # thing_id = mapped_column(
-    #     Integer, ForeignKey("thing.id", ondelete="CASCADE"), nullable=False
-    # )
-    # thing = relationship("Thing")Key("thing.id", ondelete="CASCADE"), nullable=False
-    # )
-    # thing = relationship("Thing")
-
-    # wells = association_proxy("author_associations", "author")
+    # association proxies to help keep code DRY
+    field_event = association_proxy("field_activity", "field_event")
```
ksmuczynski: I think this is the way to define proxies with the SQLAlchemy 2.0 syntax:
`field_event = AssociationProxy["FieldEvent"] = association_proxy("field_activity", "field_event")`

db/sample.py
```diff
@@ -28,102 +28,80 @@
...
+    thing = association_proxy("field_activity", "field_event.thing")
```
ksmuczynski: I think this is the way to define proxies with the SQLAlchemy 2.0 syntax:
`thing = AssociationProxy["Thing"] = association_proxy("field_activity", "field_event.thing")`

db/sample.py
```diff
@@ -28,102 +28,80 @@
...
+    contact = association_proxy("field_event_contact", "contact")  # noqa: F821
```
ksmuczynski: I think this is the way to define proxies with the SQLAlchemy 2.0 syntax:
`contact = AssociationProxy["Contact"] = association_proxy("field_event_contact", "contact")`

db/field.py
jacob-a-brown: Good catch! I'll just say "The organization that is collecting the samples from the field event"

db/field.py
jacob-a-brown: Because we are now annotating the attribute [the data type is inferred ](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapped-column-derives-the-datatype-and-nullability-from-the-mapped-annotation)from `Mapped[str]` and doesn't need to be defined in `mapped_column`. The only time we'd need to put `String` inside `mapped_column` is to restricted the character limit (e.g. `Mapped[str] = mapped_column(String(50))`

jacob-a-brown: [Some more documentation](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#orm-declarative-table)

[And](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.mapped_column)

> [mapped_column()](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.mapped_column) is normally used with explicit typing along with the [Mapped](https://docs.sqlalchemy.org/en/20/orm/internals.html#sqlalchemy.orm.Mapped) annotation type, where it can derive the SQL type and nullability for the column based on what’s present within the [Mapped](https://docs.sqlalchemy.org/en/20/orm/internals.html#sqlalchemy.orm.Mapped) annotation

db/field.py
jacob-a-brown: also added `passive_deletes=True`

db/observation.py
jacob-a-brown: For the bidirectional relationship I also added 

```python
observations: Mapped[list["Observation"]] = relationship(...)
```

to the `sample` and `sensor` models

db/sample.py
jacob-a-brown: yes! I think AI wrote that line and I didn't audit correctly. good catch

db/sample.py
jacob-a-brown: All `association_proxy`s are now type-hinted with `AssociationProxy`

db/sample.py
jacob-a-brown: All `association_proxy`s are now type-hinted with `AssociationProxy`

db/sample.py
jacob-a-brown: All `association_proxy`s are now type-hinted with `AssociationProxy`

db/observation.py
jacob-a-brown: I don't think it needs to be `UUID` because it won't be populated for new records. That is, this field is just for auditing purposes. I asked CoPilot and here is its response:

> No, it does not matter if you store a UUID from another database as a string in your model, especially if:
>
>You are not creating new records in the source database.
  The field is used only for auditing or reference purposes in your system.
  Storing UUIDs as strings is common and safe for audit/reference fields. Just ensure the string is long enough (typically 36 characters for standard UUIDs).
  If you ever need to query or compare, you may want to normalize the format (e.g., always lowercase, with hyphens).
>
>Summary:
>
>Storing a UUID as a string is fine for audit/reference.
  No impact on integrity as long as you don’t need to enforce UUID type constraints or generate new UUIDs for the source system.

db/sensor.py
```diff
@@ -40,11 +40,9 @@ class Sensor(Base, AutoBaseMixin, ReleaseMixin):
     recording_interval: Mapped[int] = mapped_column(Integer, nullable=True)
     notes: Mapped[str] = mapped_column(String(50), nullable=True)
 
-    sample = relationship(
+    samples: Mapped[list["Sample"]] = relationship(  # noqa: F821
         "Sample",
         back_populates="sensor",
```
ksmuczynski: Don't think this relationship to the Sample is necessary here. Long-term deployed equipment is managed by the `Deployment `table. Other analytical equipment (or field equipment, if we want) is managed by the `sample `relationship in the `Observation `table.

jacob-a-brown: `sample`s are no longer associated with `sensor`s

ksmuczynski (approved):
Nice work, looks good!

PR #144, opened 2025-09-24
feat: create new Deployment model

ksmuczynski:
### **Why**
- Create new Deployment table to manage the many-to-many relationship between `Things` and `Sensors`.
- Allows the storage of a complete, auditable history of which specific piece of physical hardware (**`Sensor`**) was installed at which `Thing` and for what period of time.

### **How**
- Create Deployment model and populate with appropriate fields and relationships.

### **Notes**
- This PR addresses the creation of the `Deployment` table only.
- Relationships to the Deployment table still need to be defined in the core tables (`Thing`, `Sensor`).
- Transfer script needs to be created.

### DIFF
--- a/db/deployment.py
+++ b/db/deployment.py
@@ -24,10 +24,10 @@ class Deployment(Base, AutoBaseMixin, ReleaseMixin):
 
     # --- Foreign Keys ---
     thing_id: Mapped[int] = mapped_column(
-        Integer, ForeignKey("thing.thing_id"), nullable=False
+        Integer, ForeignKey("thing.id"), nullable=False
     )
     sensor_id: Mapped[int] = mapped_column(
-        Integer, ForeignKey("sensor.sensor_id"), nullable=False
+        Integer, ForeignKey("sensor.id"), nullable=False
     )
 
     # --- Columns ---
--- a/db/deployment.py
+++ b/db/deployment.py
@@ -1,0 +1,47 @@
+"""
+This table is an installation log that creates a many-to-many relationship
+between Things and Equipment, tracking which piece of hardware was installed
+at which Thing and for what period of time.
+"""
+
+from typing import TYPE_CHECKING
+
+from sqlalchemy import Integer, ForeignKey, String, Date, Numeric, Text
+from sqlalchemy.orm import relationship, Mapped, mapped_column
+
+from db.base import Base, AutoBaseMixin, ReleaseMixin, lexicon_term
+
+if TYPE_CHECKING:
+    from db.thing import Thing
+    from db.sensor import Sensor
+
+
+class Deployment(Base, AutoBaseMixin, ReleaseMixin):
+    """
+    Represents the installation of a specific piece of equipment (Sensor) at a Thing
+    for a defined period. This is an Association Object.
+    """
+
+    # --- Foreign Keys ---
+    thing_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("thing.thing_id"), nullable=False
+    )
+    sensor_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("sensor.sensor_id"), nullable=False
+    )
+
+    # --- Columns ---
+    installation_date: Mapped[Date] = mapped_column(Date, nullable=False)
+    removal_date: Mapped[Date] = mapped_column(Date, nullable=True)
+    recording_interval: Mapped[int] = mapped_column(String(50), nullable=True)
+    recording_interval_units: Mapped[str] = lexicon_term(nullable=True)
+    hanging_cable_length: Mapped[float] = mapped_column(Numeric, nullable=True)
+    hanging_point_height: Mapped[float] = mapped_column(Numeric, nullable=True)
+    hanging_point_description: Mapped[str] = mapped_column(Text, nullable=True)
+    notes: Mapped[str] = mapped_column(Text, nullable=True)
+
+    # --- Relationships ---
+    # Many-To-One: A Deployment is for one Thing.
+    thing: Mapped["Thing"] = relationship("Thing", back_populates="deployments")
+    # Many-To-One: A Deployment is of one piece of equipment (sensor).
+    sensor: Mapped["Sensor"] = relationship("Sensor", back_populates="deployments")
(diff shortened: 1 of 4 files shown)


db/deployment.py
```diff
@@ -0,0 +1,47 @@
...
+        Integer, ForeignKey("thing.thing_id"), nullable=False
+    )
+    sensor_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("sensor.sensor_id"), nullable=False
+    )
+
+    # --- Columns ---
+    installation_date: Mapped[Date] = mapped_column(Date, nullable=False)
+    removal_date: Mapped[Date] = mapped_column(Date, nullable=True)
+    recording_interval: Mapped[int] = mapped_column(String(50), nullable=True)
```
jirhiker: Should this be `Mapped[int] = mapped_column(Integer, ... `  ?

ksmuczynski: Oops, yes, good catch!

db/deployment.py
```diff
@@ -0,0 +1,47 @@
...
+    hanging_cable_length: Mapped[float] = mapped_column(Numeric, nullable=True)
+    hanging_point_height: Mapped[float] = mapped_column(Numeric, nullable=True)
+    hanging_point_description: Mapped[str] = mapped_column(Text, nullable=True)
+    notes: Mapped[str] = mapped_column(Text, nullable=True)
+
+    # --- Relationships ---
+    # Many-To-One: A Deployment is for one Thing.
+    thing: Mapped["Thing"] = relationship("Thing", back_populates="deployments")
+    # Many-To-One: A Deployment is of one piece of equipment (sensor).
+    sensor: Mapped["Sensor"] = relationship("Sensor", back_populates="deployments")
```
jacob-a-brown, line 47: Should the `Thing` and `Sensor` tables have relationships back to the `Deployment` table for a bi-directional relationship? If that occurs, then `association_proxy`s can also be added to those tables for easy/quick access from one table to the other.

db/deployment.py
jacob-a-brown: Because the fields are type-annotated with `Mapped[<type>]` you don't need to define the type in `mapped_column` (unless you want to restrict string lengths with `String(n)` where `n` is the max number of characters)

db/deployment.py
ksmuczynski: The Deployment table is the join, or bridge, table so proxies are not necessary. The primary purpose of an association proxy is to hide the bridge. If you are already working with the bridge object, there is no intermediate step to hide.

db/deployment.py
jacob-a-brown: That being said, it doesn't hurt to have them defined in `mapped_column` - just an extra step during development

db/deployment.py
ksmuczynski: Yes, the Thing and Sensor tables should have relationships back to the Deployment table for a bi-directional relationship. I will add these relationships to the necessary models in separate PRs.

jacob-a-brown (approved):

db/deployment.py
ksmuczynski: Oh, maybe I misunderstood your proxy question. I thought you were asking if Thing and Sensor proxies should be added to the Deployment table. For that, the answer is no. But Deployment proxies should be placed on the Thing and Sensor tables. Will also add that in the coming PRs which are focused on updates to the Thing and Sensor models.

ksmuczynski: Added Deployment relationships to the `Thing `and `Sensor `models. Added `deployment` proxy to the Thing table.
The `Sensor `table doesn't need a Deployment proxy since it already has a direct 1:M relationship to the `Deployment `table. I did add a Thing proxy to the `Sensor` table.

db/sensor.py
```diff
@@ -16,10 +16,18 @@
...
 from sqlalchemy.orm import relationship, mapped_column, Mapped
 
 from db.base import Base, AutoBaseMixin, ReleaseMixin
 
+from typing import List, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .deployment import Deployment
+    from .thing import Thing
+    from .observation import Observation
```
jacob-a-brown: relative imports may fail when running this as an app from the root. to prevent this from happening you can change these to

`from db.deployment import Deployment`
`from db.thing import Thing`
`from db.observation import Observation`

db/thing.py
```diff
@@ -87,10 +96,23 @@ class Thing(Base, AutoBaseMixin, ReleaseMixin):
...
     )
 
+    # One-To-Many: A Thing can have many Deployments of sensors (equipment) over time.
+    deployments: Mapped[List["Deployment"]] = relationship(
+        "Deployment", back_populates="thing", cascade="all, delete-orphan"
+    )
+
+    # --- Association Proxies ---
+    # Proxy to directly access the Sensor deployed at this Thing.
+    sensor: AssociationProxy[List["Sensor"]] = association_proxy(
```
jacob-a-brown: should this be named `sensors` since multiple sensors may be associated with a `thing`?

jacob-a-brown (commented):
I forgot that deployments need to be imported into `db/__init__.py` for their mappers to be configured and for them to be found/mapped in the DB. This can be done by adding the line

`from db.deployment import *`

to that file

db/deployment.py
```diff
@@ -0,0 +1,47 @@
...
+    Represents the installation of a specific piece of equipment (Sensor) at a Thing
+    for a defined period. This is an Association Object.
+    """
+
+    # --- Foreign Keys ---
+    thing_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("thing.thing_id"), nullable=False
+    )
+    sensor_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("sensor.sensor_id"), nullable=False
```
jacob-a-brown: should these foreign keys be

`thing.id`
`sensor.id`

since they point to the `id` field in the `thing` and `sensor` tables?

ksmuczynski: Yes, you're right! Good catch.

jacob-a-brown (approved):

jirhiker (approved):

PR #146, opened 2025-09-25
NO TICKET: fix sample response & eager load contact attributes

jacob-a-brown:
### **Why**
- the staging database is not yet up-to-date with `FieldEvent`, `FieldEventContact`, and `FieldActivity`, so the fields `field_event_id` and `field_event_contact_id` evaluate to `None` in staging though they are not nullable
- lazily-loading related tables to `contact` (`email`, `phone`, `address`, and `thing`) causes performance issues

### **How**
- commented out `field_event_id` and `field_event_contact_id` from the `sample` response.
- used `joinedload` to eagerly load tables related to `contact`

### **Notes**
- `field_event_id` and `field_event_contact_id` probably don't need to be added back because the related tables (and their responses) are returned with the `sample` response
- add `thing_id` query param to `/sample`
- no `Field...` tables are populated in staging (yet). Will this break the responses for `contact`, `field_event`, `field_actvity`, and `thing` in the `SampleResponse`?

### DIFF
--- a/api/contact.py
+++ b/api/contact.py
@@ -27,7 +27,7 @@ from core.dependencies import (
     amp_editor_dependency,
     amp_viewer_dependency,
 )
-from db import ThingContactAssociation, Thing, Contact, Email, Phone, Address
+from db import Contact, Email, Phone, Address
 from schemas.contact import (
     CreateContact,
     CreateAddress,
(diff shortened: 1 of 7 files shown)


jacob-a-brown:
Both sample fixtures are associated with the same `thing`, so the `thing_id` query parameter returns both samples. This is the behavior we want (get samples related to a `thing`). Should I, however, associate one of the samples (and thereby its parent `FieldActivity` and `FieldEvent`) to a different `thing` to ensure that we are only returning the one we want?

jirhiker:
I think 3 sample fixtures is probably best, so that you are fully testing the functionality

jirhiker (approved):

PR #147, opened 2025-09-25
NO TICKET: Transfer

jirhiker:
### **Why**
- The transfer required modification to the schema. nma_pk_location was set as `unique` but there are non-unique LocationIDs in NMAquifer

### **How**
- removed `unique` flag from `nma_pk_location`

### **Notes**
- includes updated transfer_asset function


transfers/waterlevels_transfer.py
```diff
@@ -63,13 +63,15 @@ def transfer_water_levels(session):
...
 
-            if not pd.isna(row.TimeMeasured):
-                dt_measured = f"{row.DateMeasured} {row.TimeMeasured}"
-            else:
+            if pd.isna(row.TimeMeasured):
+                fmt = "%Y-%m-%d %I:%M:%S %p"
                 dt_measured = f"{row.DateMeasured} 12:00:00 AM"
+            else:
+                fmt = "%Y-%m-%d %I:%M:%S"
+                dt_measured = f"{row.DateMeasured} {row.TimeMeasured}"
```
jacob-a-brown, line 77: should the formats be switched? And we just don't add midnight if `TimeMeasured` is NULL? Not sure if doing that would correctly set the time to midnight, though.

jirhiker: Why do the formats need to be switched

jacob-a-brown: For some reason when I read it I thought that the `%p%` was in the wrong place - don't know why I saw that. I think that it all looks good though

jacob-a-brown (approved):

PR #150, opened 2025-09-26
BDMS-118: Create analysis method table

ksmuczynski:
### **Why**
- Create an AnalysisMethod table to store all the abstract methodologies used to generate data.
- Storing this information in a dedicated table ensures that every data point can be linked to a citable, defensible, and unambiguous procedure.
- This table is distinct from the Sensor (Equipment) table, as it deals with intangible procedures, not physical objects.

### **How**
- Created a new AnalysisMethod model and populated with appropriate fields and relationships.
- Updated the Observation table to include a foreign key to the AnalysisMethod table.
- Updated the Observation table to include a relationship to the AnalysisMethod table.
- Updated the lexicon.json file to include valid values for the analysis_method_type field.

### **Notes**
- Transfer script needs to be created.


jirhiker (approved):

db/analysis_method.py
```diff
@@ -0,0 +1,45 @@
...
+    # --- Columns ---
+    analysis_method_code: Mapped[str] = mapped_column(
+        nullable=True,
+        unique=True,
+        comment="The official code or identifier for the method (e.g., 'EPA 300.0').",
+    )
+    analysis_method_name: Mapped[str] = mapped_column(
+        nullable=False,
+        comment="The common, human-readable name of the method (e.g., Ion Chromatography for Anions).",
+    )
```
jacob-a-brown, line 32: should this be a `lexicon_term` to restrict the available values?

db/observation.py
```diff
@@ -37,7 +46,11 @@ class Observation(Base, AutoBaseMixin, ReleaseMixin):
         ForeignKey("sensor.id", ondelete="CASCADE"),
         nullable=True,
     )
+    analysis_method_id: Mapped[int] = mapped_column(
+        Integer, ForeignKey("analysis_method.id"), nullable=True
+    )
```
jacob-a-brown, line 51: Should this be non-nullable?

ksmuczynski: No, I think it should be nullable to capture scenarios where a result is derived from a direct reading from a long-term, deployed piece of equipment (like a transducer). If you're taking a water level measurement, then `analysis_method_id` would be NULL because the transducer, a piece of physical equipment, produced the result (and links back to the `Sensor` table). If the result is derived from a specific, documented set of steps, then `analysis_method_id` would be populated (and links to the `AnalysisMethod` table).

db/analysis_method.py
ksmuczynski: Maybe? I think it's kind of dependent on the lab. I imagine this field and the `analysis_method_code` fields would be coming in straight from lab EDDs. I think they are all generally standardized. I might leave as is for now and revise once we start digging into chemistry/lab data.

jacob-a-brown: We can ask AMP (and chem folks like Bonnie) if they would like for the names to be standardized. If yes then we can change to lexicon term and can keep them as string for now. If you are down with this approach, will you add a comment

```python
# TODO: get AMP feedback on if this should be restricted (i.e. lexicon term)
```
?

(I've been adding comments with AMP in the name to quickly and easily find where I need their feedback)

ksmuczynski: Yes, good idea!

jacob-a-brown (approved):

PR #151, opened 2025-09-26
feat: add lexicon permissions dependencies for editor and admin roles

jirhiker:
### **Why**
- Wanted to elevate the permission necessary to edit/admin the lexicon

### **How**
- added lexicon_editor and lexicon_admin dependencies
- added LexiconEditor and LexiconAdmin permissions to authentik

### **Notes**
- @chasetmartin we need to update the Access Control rules in OcotilloUI to set authorization for Lexicon UI components

### DIFF
--- a/api/lexicon.py
+++ b/api/lexicon.py
@@ -59,9 +59,7 @@ from services.query_helper import (
     simple_get_by_id,
 )
 
-router = APIRouter(
-    prefix="/lexicon", tags=["lexicon"]
-)
+router = APIRouter(prefix="/lexicon", tags=["lexicon"])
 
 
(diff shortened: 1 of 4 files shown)


jacob-a-brown (changes requested):
The tests in `tests/test_lexicon.py` are failing because the lexicon dependencies are not being overwritten. The function to override these dependencies is at the beginning of that file and needs to be amended for the tests to pass.

jacob-a-brown (approved):

chasetmartin:
@jirhiker ok I'll make a ticket to look into lexicon control rules in OcotilloUI and update accordingly.

PR #152, opened 2025-09-26
BDMS-139: Create `Parameter` table

ksmuczynski:
### **Why**
- Need a new `Parameter` table to store centralized metadata about each parameter, such as its category (e.g., 'Metals', 'Field Parameter') and its unique chemical identifier (CAS Number).
- The primary purpose of the `Parameter` table is is to ensure that every observation in the database is linked to a single, authoritative definition of what was measured.
- Need a new `Regulatory_Limit` table to store all the specific, citable standards and limits, regulatory and health-based, associated with a given parameter.
- The `Regulatory_Limit` table allows a single `Parameter` to have multiple different limits set by various agencies (e.g., a federal EPA limit and a state-level NMED limit).

### **How**
- Created a new `Parameter` table and populated with appropriate fields and relationships.
- Created a new `Regulatory_Limit` table and populated with appropriate fields and relationships.
- Updated the `lexicon.json` file to include valid values for the `limit_type` field.
- Added `Parameter` and `RegulatoryLimit` tables to the `__init__.py` file

### **Notes**
- Transfer script(s) needs to be created.


db/regulatory_limit.py
```diff
@@ -0,0 +1,50 @@
+"""
+This table stores the various regulatory or health-based limits for a given
+parameter, sourced from different agencies or standards.
```
jirhiker, line 3: @ksmuczynski How is this table going to stay up to date with the authoritative source? Are the regulatory limits shared by the regulators in machine-readable formats? Should we be storing this information, or simply referencing it?

db/parameter.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # --- Relationships ---
+    # One-To-Many: A Parameter can have many Observations.
+    observations: Mapped[List["Observation"]] = relationship(
+        "Observation", back_populates="parameter"
+    )
+
+    # One-To-Many: A Parameter can have many associated RegulatoryLimits.
+    # If a Parameter is deleted, all its associated limits are deleted as well.
+    reg_limits: Mapped[List["RegulatoryLimit"]] = relationship(
```
jirhiker: change this to `regulatory_limits`

jirhiker (changes requested):

db/parameter.py
```diff
@@ -0,0 +1,55 @@
...
+class Parameter(Base, AutoBaseMixin, ReleaseMixin):
+    """
+
+    Represents an analyte or property that can be measured (e.g., Chloride).
+    """
+
+    __versioned__ = {}
+
+    # --- Columns ---
+    # TODO: Parameter names are currently associated with the 'observed_property' category in the lexicon. Should we update the lexicon category name to 'parameter_name'?
```
jacob-a-brown, line 26: That's fine by me. if we do that we should also change the name of the field in the `observation` table.

jacob-a-brown: Would we even need an `observed_property` field? If each observation needs a `parameter_id` then that seems redundant and not necessary.

db/regulatory_limit.py
jacob-a-brown: My gut is telling me to fall on the side of referencing it - keeping this table up to date could be difficult and time-consuming. If someone from NMBGMR needs to know this information they can look it up.

ksmuczynski: No, as far as I know, most regulatory standards are shared in human-readable formats, like PDF docs or webpages. Keeping this table up to date would require human responsibility. I think the table itself is still valuable, though, especially for identifying and visualizing exceedances, which I think Sianin mentioned AMP does on occasion. 

If we want to keep this table, I might also add fields for `effective_date` and `end_date` to create a more complete historical record.

db/parameter.py
ksmuczynski: Can different fields reference the same lexicon term? Or does the category have to match the field name? I'm not sure I quite understand how the models use/reference the lexicon, maybe we can discuss early next week.

db/regulatory_limit.py
jirhiker: I really don't want to have to keep a copy of regulatory limits that must be updated by hand.  I agree about the value of being able to compare to reg limits, but relying on users/manager/admin to maintain is a recipe for disaster

db/parameter.py
jacob-a-brown: Different fields can reference the same lexicon term. happy to discuss this more this week

db/regulatory_limit.py
ksmuczynski: I don't know how frequently regs get updated, but I agree it would be better if the table could keep itself up to date. How should we move forward? Should we remove the table for now and introduce it later when we have a better solution for how to maintain it?

db/parameter.py
```diff
@@ -0,0 +1,55 @@
...
+        unique=True,
+        comment="The official, full name of the parameter (e.g., 'Arsenic, Dissolved').",
+    )
+    parameter_type: Mapped[str] = lexicon_term(
+        nullable=True,
+        comment="A controlled vocabulary field defining the category of the parameter (e.g., 'Metals', 'Nutrients', 'Field Parameter'). Used for grouping and filtering.",
+    )
+    cas_number: Mapped[str] = mapped_column(
+        nullable=True,
+        comment="he Chemical Abstracts Service (CAS) registry number, a globally unique identifier for a chemical substance.",
```
jacob-a-brown: "The" is missing a "T" at the beginning

db/parameter.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+    default_unit: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="The standard, preferred unit for reporting this parameter (e.g., 'ug/L', 'mg/L', 'pH units').",
```
jacob-a-brown, line 45: Should there be default units if they may change depending on the matrix? e.g. mg/kg for soil/rock and mg/L for aqueous

jacob-a-brown: or would there be multiple records for the same parameter depending on the matrix?

e.g.

arsenic | water | mg/L
arsenic | soil | mg/kg

would both be different records in this table

ksmuczynski: The purpose of the `Observation.unit` field is to store the actual, raw unit that was reported for a single, specific measurement. The `Parameter.default_unit` field would store the Bureau's preferred standard, or official reporting unit for a particular analyte. The idea behind the `default_unit` field is that it provides us the ability to standardize data. If data are received in a different unit than what the Bureau typically uses, then we can perform necessary unit conversions for plotting/vis/reporting purposes.

Ultimately, the `Parameter.default_unit` field is for ease of unit standardization. Are there other methods or strategies we should consider?

db/regulatory_limit.py
jirhiker: I think that we should keep it for now because the benefits outweigh the issues.  we should add regulatory_limits.json to populate with current values. We should reach out to an EPA contact and see if they can provide a machine readable version, or guidance on how to stay up to date

db/parameter.py
jacob-a-brown: I'm okay to have the `default_unit` field. When we return data we'd probably want to have a unit conversion function so that we can easily and quickly standardize units (and convert values) in the response to the `default_unit`.

What about my above question? If it's the same analyte, but different matrices and units, would there be two different records in the `parameter` table?

ksmuczynski: That's a great question. Yes, there should be two different records in the `Parameter` table, because the `Parameter` table is not just a list of chemical substances, it's a list of analytes, and an analyte takes the matrix into account as well. This is an important distinction as a single chemical, like arsenic, is a completely different analyte from a regulatory and scientific perspective when it's in water versus when it's in soil. They have different units, different analytical methods, and different regulatory limits.

To capture this accurately, I will add a `matrix` column to the `Parameter` table.

db/parameter.py
```diff
@@ -29,13 +29,17 @@ class Parameter(Base, AutoBaseMixin, ReleaseMixin):
         unique=True,
         comment="The official, full name of the parameter (e.g., 'Arsenic, Dissolved').",
     )
+    matrix: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="A controlled vocabulary field defining the physical medium the analyte is measured in (e.g., 'Water', 'Soil', 'Air').",
+    )
```
jacob-a-brown, line 34: this is also currently recorded in the `sample` table. should we have it in both? or remove it from `sample`?

ksmuczynski: The `matrix` field should be stored in both the `Sample` and the `Parameter` tables. `Sample.matrix` records the factual, physical medium of the sample, while `Parameter.matrix` defines the medium an analyte is intended to be measured in.

jacob-a-brown (approved):

db/regulatory_limit.py
ksmuczynski: @jirhiker If the table is populated from the values in the .json file, what is the benefit of storing values in the .json file versus in the table itself? EPA is only one of many regulatory agencies - should we reach out to all of them?

We don't need to populate with all regulatory agencies and all their standards right now, this table is meant to capture the regs AMP currently checks against and provide the flexibility to add/update in the future.

ksmuczynski: This table is not crucial to the water level functionality and can be populated at a later time.

jirhiker:
@ksmuczynski please update the lexicon.json file to use the new normalized structure. Then I will merge this PR

jirhiker:
@ksmuczynski tests are failing. once they are fixed i will merge PR

jacob-a-brown:
@jirhiker three updates needed to be made to ensure the tests past and that the API is working as expected:

- The scope of all test fixtures needed to be changed to `function` (the default), otherwise too many database connections occurred and we'd receive the error `sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00`. Because the fixtures are now scoped to functions they are torn down after the test is complete. This is to keep the tests independent.
- After changing all of the fixtures the only test that was failing was **tests/test_search.py::test_search_api**. It would find the well and the spring, but not the contact. This happened because the concatenation of 
````
vector = (
    Contact.search_vector
    | Email.search_vector
    | Phone.search_vector
    | Address.search_vector)
````
meant that if `Contact`, `Email`, `Phone`, or `Address` were `NULL` then the whole vector would be `NULL`. With the help of AI it was fixed by using

```
vector = (
        func.coalesce(Contact.search_vector, text("''::tsvector"))
        .op("||")(func.coalesce(Email.search_vector, text("''::tsvector")))
        .op("||")(func.coalesce(Phone.search_vector, text("''::tsvector")))
        .op("||")(func.coalesce(Address.search_vector, text("''::tsvector")))
    )
```

This error didn't occur previously because after the `contact`, `email`, `phone`, and `address` fixtures were invoked elsewhere they would persist. Now that invoking the `contact` fixture only creates a `contact` its related `email`, `phone`, and `address` attributes are `NULL` unless those fixtures are also invoked.
- For similar reasons, the search results for contacts should use `outerjoin` incase one of the attributes does not exist.

jirhiker (approved):

PR #155, opened 2025-09-29
Update dependabot.yml

jirhiker:
### **Why**
-  Dependabot is targeting the `pre-production` branch, but it should target `staging`

### **How**
- Updated dependabot.yml

### **Notes**
- Use bullet points here


jacob-a-brown (approved):
shoot! i missed this when going through the workflows

PR #157, opened 2025-09-29
BDMS-141

jirhiker:
### **Why**
- see BDMS-141

### **How**
- updated order_sort_filter 

### **Notes**
- Use bullet points here


jacob-a-brown (approved):
@chasetmartin after this gets merged we should verify the sorting is happening correctly on the frontend. i can't recall which columns from which were causing it to error out

PR #162, opened 2025-09-29
NO TICKET: cover all endpoints with auth dependencies

jacob-a-brown:
### **Why**
- every endpoint should be covered with an auth dependency

### **How**
- added `user: <auth dependency>` to every endpoint

### **Notes**
- Some routers had auth dependencies injected at the router, not each endpoint. I changed it to be endpoint-specific so that the `user` object can be utilized. This will allow us in the future to return objects, like `things` or `locations`, and hide some fields (e.g. `notes`), from unauthenticated users. To do this we'll need to set the viewer dependencies to `optional=True`. But that's an issue for our future selves.
  - This doesn't really adhere to DRY, but if we want to hide specific fields or records from GET endpoints I think that the auth dependency should be injected at the endpoint, not the router.
- All of the data currently being modeled is from AMP. I'm not sure when to use `amp` auth dependencies or general dependencies. This is most pertinent to the `/thing`, `/sample`, and `/observation` routers. Should they be general auth dependencies for now? Does it matter for the time being since everyone using the app is AMP?

### DIFF
--- a/api/thing.py
+++ b/api/thing.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
-from fastapi import APIRouter, Depends, Query, Request
+from fastapi import APIRouter, Query, Request
 from fastapi_pagination.ext.sqlalchemy import paginate
 from sqlalchemy import select
 from sqlalchemy.exc import ProgrammingError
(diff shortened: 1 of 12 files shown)


api/thing.py
```diff
@@ -366,7 +375,7 @@ async def create_well(
     thing_data: CreateWell,
     session: session_dependency,
     request: Request,
-    user: amp_admin_dependency,
```
jirhiker, line 369: Why was this changed?

jirhiker: ahh i see now after reading the notes

jirhiker (approved):

PR #166, opened 2025-09-30
BDMS-112 Add missing fields to `Thing` model, schema, and tests

ksmuczynski:
### **Why**
- The `Thing` model needed to inherit the new tables that have been created.
- The `Thing` model needed a `first_visit_date` field to store the date of NMBGMR's first recorded interaction with a specific Thing.
- A `well_purpose` field was needed to capture the primary function of a well (Monitoring, Irrigation, etc)

### **How**
- New `StatusHistoryMixin` and `PermissionMixin` were imported and inherited by the `Thing` model.
- Relationships were added and updated in the `Thing` model and related models.
- Syntax/notation was updated to reflect the SQLAlchemy 2.0 style.
- A `first_visit_date` field was added.
- A `well_purpose` field was added.
- A `well_type` field existed in schema files, test files, and the lexicon.json. The `well_type` field served the same purpose as the `well_purpose` field, despite not existing in the `Thing` model. Instances of `well_type` were replaced by `well_purpose` based on AMP feedback received on 2025-08-29.

### **Notes**
- Script files were not updated.


api/search.py
```diff
@@ -103,7 +103,7 @@ def make_well_response(thing: Thing) -> dict:
             "Wells",
             thing,
             {
-                "well_type": thing.well_type,
+                "well_purpose": thing.well_type,
```
jacob-a-brown: `well_type` was changed to `well_purpose` in the updates, so `thing.well_type` should change to `thing.well_purpose` here

db/contact.py
```diff
@@ -30,8 +35,12 @@ class ThingContactAssociation(Base, AutoBaseMixin):
...
     )
 
-    contact: Mapped[List["Contact"]] = relationship("Contact")
-    thing: Mapped[List["Thing"]] = relationship("Thing")  # noqa: F821
+    contact: Mapped["Contact"] = relationship(
+        "Contact", back_populates="thing_associations"
+    )
+    thing: Mapped["Thing"] = relationship(
+        "Thing", back_populates="contact_associations"
+    )  # noqa: F821
```
jacob-a-brown, line 45: `# noqa: F821` can be removed now that we are using `if TYPE_CHECKING`

db/deployment.py
```diff
@@ -4,6 +4,7 @@
 at which Thing and for what period of time.
 """
 
+from tkinter.constants import CASCADE
```
jacob-a-brown: `tkinter` is a GUI library for Python. Is this an appropriate use of it? Or, if there are cascades that are commonly used they can be defined in `constants.py` in the root directory (and then imported where needed)

db/thing.py
```diff
@@ -30,104 +36,189 @@
...
-    well_casing_description = Column(String(50))
 
-    well_construction_notes = Column(String(250))
+    # One To-Many: A Thing can be in many Groups over time.
+    thing_groups: Mapped[List["GroupThingAssociation"]] = relationship(
+        "GroupThingAssociation",
+        back_populates="thing",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
```
jacob-a-brown: this is called `thing_associations` in the `Group` table. For consistent style and naming should this be called `group_associations` here?

db/thing.py
```diff
@@ -30,104 +36,189 @@
...
+    description: Mapped[str] = mapped_column(String(500))
+    thing_type: Mapped[str] = lexicon_term(
+        nullable=True,
+        comment="A controlled vocabulary field defining the type of infrastructure (e.g., 'Well', 'Spring', 'Stream Gauge').",
+    )
+    first_visit_date: Mapped[Date] = mapped_column(
+        Date,
+        nullable=True,
+        comment="The date of NMBGMR's first recorded interaction with this specific `Thing`.",
+    )
```
jacob-a-brown, line 66: this could be an AMP question: they've been calling it `inventory date` for the longest time (i.e. `well inventory form`). Should this correspond with that vocabulary?

db/thing.py
```diff
@@ -30,104 +36,189 @@
...
+    # Proxy to directly access the Location associated with this Thing
+    locations: AssociationProxy[list["Location"]] = association_proxy(  # noqa: F821
+        "location_associations", "location"
     )
 
-    # One-To-Many: A Thing can have many Deployments of sensors (equipment) over time.
-    deployments: Mapped[List["Deployment"]] = relationship(
-        "Deployment", back_populates="thing", cascade="all, delete-orphan"
+    # Proxy to directly access the Contact objects associated with this Thing
+    contacts: AssociationProxy[list["Contact"]] = association_proxy(  # noqa: F821
```
jacob-a-brown: `# noqa: F821` can be removed since `TYPE_CHECKING` is invoked.

tests/conftest.py
```diff
@@ -51,6 +51,7 @@ def water_well_thing(location):
     with session_ctx() as session:
         water_well = Thing(
             name="Test Well",
+            first_visit_date="2023-03-03",
             thing_type="water well",
             release_status="draft",
             well_type="Production",
```
jacob-a-brown: this needs to change to `well_purpose` for revisions

db/thing.py
jacob-a-brown, line 66: should this be nullable?

schemas/thing.py
```diff
@@ -46,14 +48,15 @@ class CreateBaseThing(BaseCreateModel):
     location_id: int | None = None  # Optional location ID for the thing
     group_id: int | None = None  # Optional group ID for the thing
     name: str  # Name of the thing
+    first_visit_date: date | None = None  # Date of NMBGMR's first visit
```
jacob-a-brown: This should be `PastDate` to ensure that the date is not in the future. Also, should this be nullable when created?

jacob-a-brown: https://docs.pydantic.dev/2.2/usage/types/datetime/

schemas/thing.py
```diff
@@ -93,6 +96,7 @@ class BaseThingResponse(BaseResponseModel):
     name: str
     thing_type: str
     active_location: LocationResponse | None = None
+    first_visit_date: date | None = None
```
jacob-a-brown: should this field be nullable?

schemas/thing.py
jacob-a-brown, line 1: should `first_date_visit` be changeable so people can update that field if they made a mistake during input? or found better documentation

db/deployment.py
```diff
@@ -24,7 +25,9 @@ class Deployment(Base, AutoBaseMixin, ReleaseMixin):
 
     # --- Foreign Keys ---
     thing_id: Mapped[int] = mapped_column(
-        Integer, ForeignKey("thing.id"), nullable=False
+        Integer,
+        ForeignKey("thing.id", ondelete=CASCADE),
```
jirhiker: like the other instances of `ondelete=...` this should be `ondelete="CASCADE"`.  And then `from tkinter.constants import CASCADE` can be removed

db/deployment.py
jirhiker: this should be removed.   like the other instances of `ondelete`,  CASCADE should be a str e.g.  `ondelete="CASCADE"`

db/thing.py
ksmuczynski: They can keep referring to it as `inventory date` if they want, but in my opinion, `first_visit_date` is less ambiguous and more accurately describes what the field is storing. I think table and field names should be as explicit as possible.

ksmuczynski: Yes. A NULL value here would indicate that the date of NMBGMR's first visit is unknown, which is a possibility, particularly with legacy data.

schemas/thing.py
ksmuczynski: `PastDate `makes sense, will update.

Yes, I think it should be nullable.  A NULL value would indicate that the date of NMBGMR's first visit is unknown, which is a possibility, particularly with legacy data.

jacob-a-brown: We can allow it to be nullable in the data transfer, but should it be required for new records? If so, then it shouldn't be nullable in `CreateBaseThing` otherwise they'll be allowed to submit NULL values for new records

schemas/thing.py
ksmuczynski: Yes, I think so? If a user requests legacy data information from a Thing and its first visit date is unknown, shouldn't NULL be returned?

schemas/thing.py
ksmuczynski: Yes, I think it should be editable. The following is what I have in the UPDATE schema - is this correct?
`first_visit_date: date | None = None `

schemas/thing.py
ksmuczynski: Oh, I see. No, you are right, it should be required for new records. Updating this in the `schema/thing.py` file will only affect what the user is able to do, is that correct? The database will still be able to store NULL values for the legacy data?

jacob-a-brown: That's correct, this will only effect incoming data. We don't use this for the transfer of legacy data. Since it is still nullable in the database it will be able to store the legacy NULL values

db/thing.py
ksmuczynski: Yes, good catch. `thing_association` is also more descriptive of the relationship. I will update this relationship name and its reference in the `groups` proxy.

schemas/thing.py
jacob-a-brown: Correct! Though it should be `PastDate` so the user can't enter a future date.

There's another style we are going to adopt for update that makes it so fields don't need to be in the payload but if they are present they must be of a certain type, but that's a future us problem.

db/contact.py
```diff
@@ -41,53 +52,64 @@ class Contact(Base, AutoBaseMixin, ReleaseMixin):
...
     )
-    things: AssociationProxy[list["Thing"]] = association_proxy(  # noqa: F821
-        "thing_associations", "thing"
-    )
-
-    # Proxy to directly access the FieldEvent objects in which this Contact participated.
-    # fmt: off
-    field_event_contact_associations: Mapped[list["FieldEventContactAssociation"]] = (  # noqa: F821
+    # One-To-Many: A Contact can participate in many Field Events.
+    field_event_contact_associations: Mapped[list["FieldEventContactAssociation"]] = (
```
jacob-a-brown: After this gets merged I'll update my PR (171) to prevent merge conflicts (`field_event_contact_assocations` --> `field_event_participants`)

db/contact.py
```diff
@@ -41,53 +52,64 @@ class Contact(Base, AutoBaseMixin, ReleaseMixin):
...
+
+    # Proxy to directly access the Thing objects associated with this Contact
+    things: AssociationProxy[list["Thing"]] = association_proxy(
+        "thing_associations", "thing"
+    )
+
+    # Proxy to directly access the FieldEvent objects in which this Contact participated.
+    field_events: AssociationProxy[list["FieldEvent"]] = association_proxy(
+        "field_event_contact_associations", "field_event"
+    )
```
jacob-a-brown, line 106: After this gets merged I'll update my PR (171) to prevent merge conflicts (`field_event_contact_assocations` --> `field_event_participants`)

schemas/thing.py
```diff
@@ -192,11 +196,12 @@ class UpdateThing(BaseUpdateModel):
     """
 
     name: str | None = None  # Optional name for the thing
+    first_visit_date: date | None = None  # Date of NMBGMR's first visit
```
jacob-a-brown: This should be `PastDate` to prevent the user from entering future dates

db/contact.py
ksmuczynski: Updated the following instances: `FieldEventContactAssociation`--> `FieldEventParticipants`
`field_event_contact_associations `--> `field_event_participants`

schemas/thing.py
jirhiker: @ksmuczynski  see above

ksmuczynski: @jirhiker Done!

jirhiker (approved):

jacob-a-brown (approved):
looks good to me

PR #168, opened 2025-10-01
Transfer

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


jacob-a-brown (approved):
`field_events` (and associated activities and contacts) should still be made even if `DepthToWater` is NULL because AMP wants to record that they went out even if they couldn't determine a water level. I can make this fix when I work on the mapper some more after it's been merged with `staging`

jacob-a-brown:
Or, if the fix should be made ahead of time, move the check `if pd.isna(row.DepthToWater)` after the field events/activites/contacts have been created right above `# ---Samples---`

PR #169, opened 2025-10-01
NO TICKET: move level_status from observation to field_activity & rename | make observation value non-nullable

jacob-a-brown:
### **Why**
- an observation must have a value to be a valid observation, otherwise something occurred to prevent the observation from being observed. 

### **How**
- made the `value` field in `observation` non-nullable
- move `level_status` from the `observation` table to the `field_activity`  table and rename `activity_status`. This allows multiple activities to occur during an event, such as groundwater level and maintenance, but the status of one won't affect the status of the other.

### **Notes**
- Use bullet points here


jirhiker (approved):

ksmuczynski:
Thinking more about lab data and `value` being nullable or not, I think it SHOULD be nullable. It's very common for labs to analyze samples, but the chemical is not present at a high enough concentration to be quantified. In these situations, the value is null, but there are lab qualifiers and reporting limits to provide further context.

db/observation.py
```diff
@@ -56,7 +56,8 @@ class Observation(Base, AutoBaseMixin, ReleaseMixin):
     )
     observed_property: Mapped[str] = lexicon_term(nullable=False)
     value: Mapped[float] = mapped_column(
-        nullable=True,
+        nullable=False,
+        comment="The value of the observation. If a value could not be retrieved then an observation could not be made. The reason for the missing value should be documented in the correct table, like field_activity.",
```
ksmuczynski, line 60: Thinking more about lab data and `value` being nullable or not, I think it SHOULD be nullable. It's very common for labs to analyze samples, but the chemical is not present at a high enough concentration to be quantified. In these situations, the value is null, but there are lab qualifiers and reporting limits to provide further context.

I agree NULL values can be a pain to work with sometimes, but no result _is_ a result and the Observation table has additional fields to explain why.

ksmuczynski (changes requested):

jacob-a-brown:
Closing PR for more consideration, debate, and feedback from AMP before proceeding.

PR #171, opened 2025-10-02
BDMS 114 & Others: add value_reason to `observation` | update FieldEventContactAssociation | remove geothermal clutter

jacob-a-brown:
### **Why**
- A user needs to record the reason for a NULL water level value using a controlled vocabulary
- `FieldEventContactAssociation` is an unclear name
- The geothermal group needs to be consulted before their data are imported and legacy code is cluttering the code base

### **How**
- Changed `level_status` to `value_reason` and made it a non-nullable field for *all* observations. The `value_reason` field allows users to provide context for a single observation value. This is particularly important for `NULL` water levels so we can deduce which are `NULL` because the well is dry (as opposed to other reasons like an obstruction)
- `FieldEventContactAssociation` was renamed `FieldEventParticipant` for clarity
- legacy geothermal code in the `observation` table was removed to declutter the code base

### **Notes**
- The consensus on when a groundwater level observation is `NULL`: 
  - If anything occurs before an attempt is made to retrieve a groundwater level and it prevents such an action, then a `FieldEvent` record is made but no `Observation` or `Sample` records are made. It is then noted in the `FieldEvent` record why this happened. A restricted vocabulary could be useful, but there are so many things that can go wrong we thought to just have it noted in the `notes` field. This could be anything from a flat tire to a locked gate to muddy roads to a destroyed well.
    - If a well is found to be destroyed there should be a workflow to update the `StatusHistory`, too.
    - A `WaterLevels` record that indicates a destroyed well will create a `FieldEvent` record but no `sample` or `observation`
  - If anything happens *after* a sample/observation has been attempted (such as an obstruction, equipment failure, or dry well), then an `observation` record is made, `value` is set to `NULL`, and controlled vocabulary is used in `value_reason` to indicate *why* it is `NULL`. As written above, this is particularly important for water levels so we can deduce if a well is dry or not.


jirhiker (approved):

db/contact.py
```diff
@@ -74,23 +78,21 @@ class Contact(Base, AutoBaseMixin, ReleaseMixin):
...
     )
-    things: AssociationProxy[list["Thing"]] = association_proxy(  # noqa: F821
+    things: AssociationProxy[list["Thing"]] = association_proxy(
         "thing_associations", "thing"
     )
 
     # Proxy to directly access the FieldEvent objects in which this Contact participated.
-    # fmt: off
-    field_event_contact_associations: Mapped[list["FieldEventContactAssociation"]] = (  # noqa: F821
+    field_event_participants: Mapped[list["FieldEventParticipant"]] = (  # noqa: F821
```
ksmuczynski: FieldEvent is included in TYPE_CHECKING so you can likely remove the `#noqa: F821`

db/contact.py
```diff
@@ -74,23 +78,21 @@ class Contact(Base, AutoBaseMixin, ReleaseMixin):
...
         relationship(
-            "FieldEventContactAssociation",
+            "FieldEventParticipant",
             back_populates="contact",
             cascade="all, delete-orphan",
             passive_deletes=True,
         )
     )
-    # fmt: on
     field_events: AssociationProxy[list["FieldEvent"]] = (  # noqa: F821
```
ksmuczynski: FieldEvent is included in TYPE_CHECKING so you can likely remove the #noqa: F821

db/observation.py
```diff
@@ -67,15 +72,6 @@ class Observation(Base, AutoBaseMixin, ReleaseMixin):
...
-    # geothermal
-    observation_depth: Mapped[float] = mapped_column(
-        nullable=True,
-        info={"unit": "feet"},
-        doc="Depth of the geothermal observation in feet",
-    )
-
     # --- Relationships ---
     # Many-To-One: An Observation can be generated by one piece of Equipment.
     sensor: Mapped["Sensor"] = relationship(  # noqa: F821
```
ksmuczynski: Sensor is included in TYPE_CHECKING so you can likely remove the #noqa: F821

ksmuczynski (approved):

TylerAdamMartinez (approved):
Great refactor here! 😄

api/observation.py
TylerAdamMartinez, line 1: As this file grows, it would make sense to refactor the `api/observation.py` section into smaller, more focused modules (e.g., `api/observation/get.py`, `api/observation/post.py`, etc.) to improve maintainability and clarity.

jirhiker (approved):

api/observation.py
jacob-a-brown: Would you suggest that for all routers? Every one has `GET`, `POST`, `PATCH`, and `DELETE` methods defined in them.

PR #172, opened 2025-10-03
NO TICKET: rename contact -> participant where appropriate

jacob-a-brown:
### **Why**
- I forgot to update the fields after renaming `FieldEventContactAssociation` to `FieldEventParticipant`

### **How**
- renamed `field_contact_role` --> `participant_role`
- renamed mapped `contact` in `FieldEventParticipant` table --> `participant`

### **Notes**
- Kept the name `contact` in the `sample` table


jirhiker (approved):

core/lexicon.json
ksmuczynski, line 1: Does it make sense to add these roles to the `role` category that exists for Contacts, or is it better to have a separate category for `participant_role`?

ksmuczynski (approved):

core/lexicon.json
jacob-a-brown: At first I thought to do that, but they kind of serve two different purposes. `role` describes the person (or organization) (e.g. `Geologist`, `Biologist`, `Technician` etc), whereas `participant_role` describes what role the person had in the field event (`Lead`, `Participant`, `Observer` etc). Trevor Kludt, for example, is a `Technician`, but during the field events he can be a `Lead` or a `Participant`

PR #184, opened 2025-10-10
jir-transducer-data

jirhiker:
### **Why**
- Need to transfer WaterLevelsContinous_Pressure to Ocotillo

### **How**

### **Notes**
- Use bullet points here


jacob-a-brown (commented):
Why have a separate table for transducer observations?

ksmuczynski:
@jacob-a-brown Because it's storing time series data. If continuous data were captured and stored in the same manner as our discrete samples, every 15-minute reading from a pressure transducer would create a new `FieldEvent`, `FieldActivity`, `Sample`, and `Observation` record. The core tables would become massive and slow, not to mention redundant.

I think a separate table is def the way to go, but I have some minor comments that I'm working on writing up and integrating into the review :)

jacob-a-brown:
When we retrieve all observations from a well, should it also retrieve the continuous observations? or will that be separate?

ksmuczynski:
@jacob-a-brown I think it should return the continuous data, but aggregated based on the length of a user's requested date range. Like, if the date range is long (e.g., > 1 month), then the raw 15-minute data is not returned. The API could ask the database to calculate daily or weekly averages. If the range is short (e.g., < 3 days), the API can then decide it's safe to fetch the raw, continuous data.

jirhiker:
There likely is a business use case for correlated continuous and manual waterlevel measurements, and time series statistics.

db/transducer.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    # sensor_id: Mapped[int] = mapped_column(Integer, ForeignKey("sensor.id"))
+    parameter_id: Mapped[int] = mapped_column(Integer, ForeignKey("parameter.id"))
+
+    value: Mapped[float] = mapped_column(Float)
+    observation_datetime: Mapped[datetime] = mapped_column(DateTime(timezone=True))
+
+    thing: Mapped["Thing"] = relationship("Thing")
+    # sensor: Mapped["Sensor"] = relationship("Sensor")
+    parameter: Mapped["Parameter"] = relationship("Parameter")
```
ksmuczynski, line 63: Instead of naming the table 'TransducerObservation`, I suggest a more general name that's inclusive of other sensors that are also capable of collecting continuous measurements.

The Deployment table is really the table that provides the context for which `Thing` and which `Sensor`, or piece of equipment, generated the data. I suggest swapping out the `Thing` references and ForeignKeys with `Deployment` ones.

See my suggestions below.

```suggestion
if TYPE_CHECKING:
    from db.deployment import Deployment
    from db.parameter import Parameter


class ContinuousData(Base, AutoBaseMixin, ReleaseMixin):
    """
    Represents a single, raw data point from a continuous monitoring sensor.
    """
    __tablename__ = "continuous_data"

    # --- Foreign Keys ---
    deployment_id: Mapped[int] = mapped_column(
        Integer, ForeignKey("deployment.id", ondelete="CASCADE"), nullabe=False
    )
    parameter_id: Mapped[int] = mapped_column(Integer, ForeignKey("parameter.id"), nullable=False)

    # --- Columns ---
    timestamp: Mapped[DateTime] = mapped_column(DateTime, nullable=False,)
    value: Mapped[float] = mapped_column(Float, nullable=False)
    
   # --- Relationships ---
    # Many-To-One: A data point belongs to one Deployment.
    deployment: Mapped["Deployment"] = relationship("Deployment", back_populates="continuous_data", lazy="joined")
    
    # Many-To-One: A data point is for one Parameter.
    parameter: Mapped["Parameter"] = relationship("Parameter", back_populates="continuous_data", lazy="joined")
```

db/__init__.py
```diff
@@ -39,6 +39,7 @@
 from db.sensor import *
 from db.status_history import *
 from db.thing import *
+from db.transducer import *
```
ksmuczynski, line 42: If the table name changes, this import statement should be updated, too.

ksmuczynski (changes requested):
See my comment/suggestion in the `db/transduce.py` file about referencing the `Deployment` table instead of the `Thing` table.

db/transducer.py
jirhiker: Continuous is ambiguous/inaccurate.  

I've made updates so that the observations are associated with a `deployment` not a `thing`

jirhiker:
@jacob-a-brown  have a look how I added dynamic enum generation for use with Pydantic validation

db/transducer.py
ksmuczynski: But it also kind of implies that the table is for transducer data only. What about other sensors/instruments?

jirhiker: All sensors are pretty much transducers,  "A transducer is a device that usefully [converts](https://en.wikipedia.org/wiki/Energy_transformation) energy from one form to another.[[1]](https://en.wikipedia.org/wiki/Transducer#cite_note-Stonier_2012-1) Usually a transducer converts a [signal](https://en.wikipedia.org/wiki/Signal) in one form of energy to a signal in another.[[2]](https://en.wikipedia.org/wiki/Transducer#cite_note-2) Transducers are often employed at the boundaries of [automation](https://en.wikipedia.org/wiki/Automation), [measurement](https://en.wikipedia.org/wiki/Measuring_instrument), and [control systems](https://en.wikipedia.org/wiki/Control_system), where electrical signals are converted to and from other physical quantities (energy, force, torque, light, motion, position, etc.)."

core/enums.py
jacob-a-brown, line 1: I like creating the enums for validations. If, some updates a term or category, we'll need to refresh these enums for the live app, otherwise they'll remain outdated. This could go into `lexicon_helper.py`.

db/transducer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    start_time: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
+    end_time: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
+
+    comment: Mapped[Optional[str]] = mapped_column(Text)
+    reviewer: Mapped[Optional[str]] = mapped_column(Text)
```
jacob-a-brown: The style we've adopted for type-hinted model columns is 

`Mapped[<type>]=mapped_column(...,nullable=[True/False])`

 `Optional[<type>]` or `<type> | None` both work too, but we landed on `nullable=[True/False]` in `mapped_column` to be extra explicit

db/transducer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    parameter: Mapped["Parameter"] = relationship("Parameter")
+
+    # Direct relationship to observations
+    observations: Mapped[List["TransducerObservation"]] = relationship(
+        "TransducerObservation",
+        back_populates="qc_block",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
```
jacob-a-brown: should this be an eager load? Or will that overwhelm the system because of all of the data that will be retrieved?

db/transducer.py
jacob-a-brown, line 1: `parameter_id` is defined in both the observation block and the observation. is it needed in both places?

db/transducer.py
```diff
@@ -0,0 +1,107 @@
...
+from sqlalchemy.orm import mapped_column, Mapped, relationship
+
+from db import Base, AutoBaseMixin, ReleaseMixin, lexicon_term
+
+if TYPE_CHECKING:
+    from db.parameter import Parameter
+    from db.deployment import Deployment
+
+
+class TransducerObservationBlock(Base, AutoBaseMixin, ReleaseMixin):
```
jacob-a-brown, line 38: Since this represents a contiguous block of observations that share a QC status, we may want to consider adding [UniqueConstraints](https://docs.sqlalchemy.org/en/20/core/constraints.html#unique-constraint), perhaps on `qc_status`, `start_time`, and `end_time`.

db/transducer.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+    qc_status: Mapped[str] = lexicon_term()
+
+    start_time: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
+    end_time: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
```
jacob-a-brown: rename `start_datetime` and `end_datetime` for clarity?

tests/test_contact.py
```diff
@@ -781,15 +782,15 @@ def test_patch_contact_409_null_organization(third_contact):
 def test_patch_contact_409_bad_contact_type(third_contact):
```
jacob-a-brown: This test should be renamed `test_path_contact_422_bad_contact_type` to reflect the new status code.

jacob-a-brown (commented):
I like the use of enums to validate POST and PATCH data. This can (and should) be used elsewhere.

db/transducer.py
jirhiker: tbd

db/transducer.py
jirhiker: removed from observation

jacob-a-brown (approved):

db/transducer.py
```diff
@@ -0,0 +1,112 @@
...
+
+    start_datetime: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
+    end_datetime: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True), nullable=False, index=True
+    )
+
+    comment: Mapped[str] = mapped_column(Text, nullable=True)
+    reviewer: Mapped[str] = mapped_column(Text, nullable=True)
```
jacob-a-brown: Like `field_event_participant` and `sample`, this can relate back to the `contact` table to restrict this field to a pre-defined list of people. Or, this can be set automatically to the person who is logged in since they are the one doing the reviewing (like audit mixin).

db/transducer.py
```diff
@@ -0,0 +1,112 @@
...
+    from db.parameter import Parameter
+
+
+class TransducerObservationBlock(Base, AutoBaseMixin, ReleaseMixin):
+    """
+    Represents a contiguous block of transducer observations that share a QC status.
+    """
+
+    parameter_id: Mapped[int] = mapped_column(
+        ForeignKey("parameter.id", ondelete="CASCADE"), nullable=False, index=True
```
ksmuczynski, line 44: Did you mean to put parameter_id here? I don't understand it's purpose.

db/transducer.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+
+    qc_status: Mapped[str] = lexicon_term()
```
ksmuczynski: What do you think about `review_status` instead of `qc_status`? I usually think of QC as quality assessment (Good, Fair, Poor, Rejected, etc). It seems like we're trying to track the data processing workflow with this field.

db/transducer.py
jirhiker: its to identify which TransducerObservations this block is referencing. without it it wouldnt be clear which observations within the date range this block is being applied to. e.g. This set of groundwater levels is approved vs this set of temperatures is approved.

jirhiker:
#197 Has these changes

PR #186, opened 2025-10-10
BDMS 170: allow well to have multiple purposes and casing materials

jacob-a-brown:
### **Why**
- **NM_Aquifer** sometimes records multiple purposes and casing materials for a well, so new tables/relationships needed to be defined to foster DB normalization
- the transfer scripts needed to be updated for these new tables

### **How**
- Added `WellPurpose` and `WellCasingMaterial` tables and related them to `thing`
- Updated schemas and tests. The `WellResponse` schema just returns a list of purposes and materials, not the full objects
- update well transfer script for the new tables

### **Notes**
- updated `patch_thing` to remove associated purposes/materials and add new ones if the user wants to edit them (since there's no direct CRUD access to those tables). Is this the right way to proceed?
- the search API now uses different vectors for wells and springs because of the two new tables

### DIFF
--- a/services/thing_helper.py
+++ b/services/thing_helper.py
@@ -38,6 +38,11 @@ from services.query_helper import make_query, order_sort_filter, simple_get_by_i
 from shapely import wkb
 from shapely.geometry import mapping
 
+WELL_CHILD_MODEL_MAP = {
+    "well_purposes": (WellPurpose, "purpose"),
+    "well_casing_materials": (WellCasingMaterial, "material"),
+}
+
 
 def wkb_to_geojson(wkb_element):
     if wkb_element is None:
@@ -140,12 +145,11 @@ def add_thing(
         thing_type = get_thing_type_from_request(request)
 
     if isinstance(data, BaseModel):
-        data = data.model_dump()
+        well_child_table_list = list(WELL_CHILD_MODEL_MAP.keys())
+        data = data.model_dump(exclude=well_child_table_list)
 
     location_id = data.pop("location_id", None)
     group_id = data.pop("group_id", None)
-    well_purposes = data.pop("well_purposes", None)
-    well_casing_materials = data.pop("well_casing_materials", None)
 
     try:
         thing = Thing(**data)
@@ -172,22 +176,6 @@ def add_thing(
             assoc.location_id = location_id
             assoc.thing_id = thing.id
             session.add(assoc)
-
-        if well_purposes:
-            for well_purpose in well_purposes:
-                wp = WellPurpose()
-                audit_add(user, wp)
-                wp.thing_id = thing.id
-                wp.purpose = well_purpose
-                session.add(wp)
-
-        if well_casing_materials:
-            for well_casing_material in well_casing_materials:
-                wcm = WellCasingMaterial()
-                audit_add(user, wcm)
-                wcm.thing_id = thing.id
-                wcm.material = well_casing_material
-                session.add(wcm)
 
         session.commit()
         session.refresh(thing)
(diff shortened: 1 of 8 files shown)


services/thing_helper.py
```diff
@@ -213,6 +239,43 @@ def patch_thing(
 
     verify_thing_type_correspondence(thing, request)
 
+    data = payload.model_dump(exclude_unset=True)
```
jirhiker: I don't think we have a business use case for this currently.  Lets leave it out for now

jirhiker (changes requested):

services/thing_helper.py
jirhiker: everything else is looking good!

jacob-a-brown: Remove the ability to update `well_purpose`?

jirhiker: yes.  updating the well_purpose and well_casing_materials should follow the same CRUD structure as all other models.  just because its not in the patch payload doesn't mean it should be deleted

jacob-a-brown: It's only deleting the previous records and creating new ones if `well_purposes` is in the payload (same with `well_casing_materials`); if `well_purposes` is not in the payload nothing happens. Since a well may have many purposes, if the user changes the purposes from `["Domestic", "Irrigation"]` to `["Observation"]` I don't know which one to change. So rather than change an existing record I remove the existing ones and create new ones for the new purposes.

jacob-a-brown: Rather than have this handled in the generic `patch_thing` function, though, I can handle these cases at the PATCH `/thing/water-well/{thing_id}` endpoint. That'll keep the code much cleaner and separate concerns

jacob-a-brown: Which, if we want to proceed with that updating method (deletion/insertion), this seems to make a lot more sense for code usability, maintainability, and readability

services/thing_helper.py
```diff
@@ -231,23 +231,23 @@ def patch_thing(
...
+    This function is to add and update well descriptor tables when a Thing is created
+    or updated. It deletes existing descriptor table records for the Thing if they
     exist and then adds the new data.
     """
     try:
-        for child_table in WELL_CHILD_MODEL_MAP.keys():
-            db_table, field_name = WELL_CHILD_MODEL_MAP[child_table]
-            child_table_data = payload.model_dump(exclude_unset=True).pop(
-                child_table, None
+        for descriptor_table in WELL_DESCRIPTOR_MODEL_MAP.keys():
```
jirhiker, line 243: `for key,value in WELL_DESCRIPTOR_MODEL_MAP.items():`

services/thing_helper.py
```diff
@@ -231,23 +231,23 @@ def patch_thing(
...
+            db_table, field_name = WELL_DESCRIPTOR_MODEL_MAP[descriptor_table]
+            descriptor_table_data = payload.model_dump(exclude_unset=True).pop(
+                descriptor_table, None
             )
-            if child_table_data:
+            if descriptor_table_data:
                 session.query(db_table).filter(db_table.thing_id == thing.id).delete()
-                for ctd in child_table_data:
+                for ctd in descriptor_table_data:
```
jirhiker, line 250: Explain again why the record needs to be deleted and re-inserted?

jirhiker (changes requested):

services/thing_helper.py
jacob-a-brown: To keep the data normalized the descriptor tables (`well_purpose` and `well_casing_material`) essentially store an array of values. If the purposes of a well change how do I know which records to update? For example:

Let's say that well `A` has two purposes: `Domestic` and `Irrigation`. If, in the future, someone wants to change its purposes to just `Observation`, which record in the `well_purpose` table do we change? And which do we delete? Rather than deciding on this, I just figured it would be easier to remove the associated `well_purpose` records and add the new one(s) in.

I could also talk with @chasetmartin and @TylerAdamMartinez about how they envision this being handled by the frontend. If a user wants to change purposes would they change one at a time (and therefore hit an endpoint to update a single well purpose record), or would the purposes be submitted in the payload like for POST `/thing/water-well`? If the former I can create a PATCH endpoint for well purposes, but if the latter I think that deletion/insertion would be the way to go.

jirhiker: "Let's say that well A has two purposes: Domestic and Irrigation. If, in the future, someone wants to change its purposes to just Observation, which record in the well_purpose table do we change? And which do we delete? Rather than deciding on this, I just figured it would be easier to remove the associated well_purpose records and add the new one(s) in."  thanks that makes sense

jirhiker (approved):

PR #189, opened 2025-10-17
BDMS-115 Sensor model updates

ksmuczynski:
### **Why**
This PR introduces enhancements and fixes across multiple modules related to sensor management and data validation. These changes enhance the system's ability to manage sensor deployments and validate data in support of groundwater monitoring workflows.

 - Updates the `Sensor` model to serve as the central asset inventory for all physical hardware used for data collection.
- Refactors the `Deployment` model to better track sensor installations and their relationships with `Thing` entities.
- Improves schema definitions for creating, updating, and responding to sensor-related data.
- Improves the sensor transfer logic and aligns it with the updated sensor model.
### **How**
- Added new fields to track equipment details:
  - `model` - Equipment model/make
  - `serial_no` - Serial number
  - `pcn_number` - Property control number 
  - `owner_agency` - Owning organization
  - `sensor_status` - Current status (In Service, In Repair, etc.)
  - `notes` - General notes field
- Added controlled vocabulary terms for:
  - Sensor types (Pressure Transducer, Data Logger, etc.)
  - Sensor status values (In Service, In Repair, Retired, Lost)
- Added `CreateSensor`, `UpdateSensor`, and `SensorResponse` schemas in `schemas/sensor.py`.
- Updated the `Deployment` model in `db/deployment.py` to include detailed attributes like `hanging_cable_length` and `hanging_point_description`.
- Introduced new fixtures in `tests/conftest.py` for testing sensor and deployment relationships.
- Removed outdated validation logic for `datetime_installed` and `datetime_removed` in the `Sensor` model.
- Added and updated unit tests to cover the new transfer logic.
- Updated documentation strings for better clarity.

### **Notes**
- Transfer related notes:
  - Sensors are present multiple times in the Equipment table, once for each deployment. To avoid having duplicates the sensor is only created once when the script first comes across it in NM_Aquifer. The nma_pk_equipment field is populated from the record from which the sensor was created. This assumes that the data is the same between the different records; the data is not checked from one record to another.
  - Sensors for wells are first grouped by DateInstalled and then added. If the DateRemoved field is NULL for the last deployment the sensor_status field is set to In Service. Otherwise if DateRemoved is not NULL for the latest deployment the sensor_status field is set to Retired. Is this the correct way to proceed?
  - The equipment types that are kept, for now, are `Pressure transducer`, `Acoustic sounder`, and `Barometer`


jirhiker (approved):
looks good. excited to merge this with the transducer pr

db/sensor.py
```diff
@@ -31,22 +33,54 @@
...
+        nullable=True,
+        unique=True,
+        comment="The Property Control Number (PCN) assigned to equipment for inventory tracking. This is only available for equipment owned by the NMBGMR.",
+    )
+    owner_agency: Mapped[str] = lexicon_term(
+        nullable=True, comment="The agency or organization that owns the equipment."
+    )
+    sensor_status: Mapped[str] = lexicon_term(
+        nullable=True,
+        comment="A controlled vocabulary field to indicate the current status of the equipment (e.g., 'In Service', 'In Repair', 'Retired', 'Lost', etc).",
```
jacob-a-brown, line 77: Is this something that goes in the `StatusHistory` table?

api/sensor.py
```diff
@@ -127,8 +131,8 @@ async def delete_sensor(
 async def get_sensors(
     session: session_dependency,
     user: viewer_dependency,
-    thing_id: int = None,  # Optional filter for thing_id
-    observed_property: str = None,  # Optional filter for observed_property
+    thing_id: int = None,  # Optional filter for thing_id. Filter by the Thing where equipment is deployed
+    parameter_id: int = None,  # Filter by the parameter the sensor/equipment measures
```
jacob-a-brown, line 135: Since these query parameters are now present in this endpoint we should make tests to ensure they work as expected and to safeguard future development.

PR #196, opened 2025-10-20
NO TICKET: implement GET /thing/{thing_id}/deployment

jacob-a-brown:
### **Why**
- The well detail page needs to be able to display all sensors that have been deployed at that well

### **How**
- Return a list of all deployments at a well

### **Notes**
- The endpoint works by getting all deployments at a well. It eagerly loads the sensor for each deployment to avoid N+1 problems. Do we want to proceed through the deployment table, or does it make sense to go through the sensor table?
  - If we go through the sensor table I can return a list of sensors and their deployment history, though I'd need to put in a little of extra work to ensure that the deployment history only pertains to the appropriate well


jirhiker (approved):

TylerAdamMartinez (approved):
lgtm

chasetmartin (approved):

PR #198, opened 2025-10-23
BDMS-195: Create Polymorphic Notes Table

ksmuczynski:
### **Why**
- Current implementation stores notes as text strings directly in models
- This approach lacks structure, metadata, and searchability
- Different entity types (Things, Locations) need a consistent way to store and retrieve notes
- String-based notes don't support historical tracking or attribution

### **How**

- Created a new polymorphic notes model with NotesMixin for reuse across entities
- Updated Thing and Location models to use the new polymorphic relationship
- Modified Pydantic schemas to represent notes as List[NoteResponse]
- Added note_type attribute to categorize different kinds of notes
- Implemented content field to store the actual note text

### **Notes**
- Tests are failing, but that is because test fixtures still need to be updated to use the new polymorphic notes structure.
- Should discuss the advantages/disadvantages of having a separate endpoint for notes.

### DIFF
--- a/core/lexicon.json
+++ b/core/lexicon.json
@@ -23,7 +23,7 @@
     {"name": "limit_type", "description": null},
     {"name": "measurement_method", "description": null},
     {"name": "monitoring_status", "description": null},
-    {"name": "noteable_type", "description": null},
+    {"name": "note_type", "description": null},
     {"name": "parameter_name", "description": null},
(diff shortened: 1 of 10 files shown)


jirhiker:
superseded by https://github.com/DataIntegrationGroup/NMSampleLocations/pull/241

PR #205, opened 2025-10-24
NO TICKET: Make local tests pass

kbighorse:
### **Why**
- Tests pass against docker, but
- Tests fail against the file system

### **How**
- Handled database cleanup between test runs
- Addressed timezone issues

### **Notes**
- All tests pass locally now

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,6 +59,7 @@ jobs:
         MODE: development
         POSTGRES_HOST: localhost
         POSTGRES_PORT: 5432
+        POSTGRES_USER: postgres
         POSTGRES_PASSWORD: postgres
         DB_DRIVER: postgres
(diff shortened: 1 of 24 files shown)


jacob-a-brown (commented):
I like that all datetime objects are serialized as UTC with `UTCAwareDatetime`. Our intention is to store all datetime records in the database as UTC. Should `UTCAwareDatetime` be used in all `Create...` schemas and return UTC objects so that both storage and serialization are guaranteed to be UTC? Or, do you think that should always be handled by a `field_validator` in all of the appropriate Pydantic schemas?

kbighorse:
Definitely the former. The latter couldn't hurt as well, or at least test for it?

kbighorse:
I created an [issue for UTC enforcement](https://github.com/DataIntegrationGroup/NMSampleLocations/pull/205#pullrequestreview-3378103510)

jacob-a-brown (approved):

PR #210, opened 2025-10-28
BDMS-200: Transfer and Feature tests

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


.github/workflows/tests.yml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+#      run: |
+#        mkdir -p tests/features
+#        cp -r bdd/features/backend/* tests/features/
+#
+#    - name: Run BDD tests
+#      env:
+#        BASE_URL: ${{ secrets.BACKEND_URL }}
+#      run: |
+#        uv run behave tests/features --tags=@backend,@approved --no-capture
+
```
jacob-a-brown, line 85: For clean code and now that the features are in this repo should this be removed instead of commented out?

api/geospatial.py
```diff
@@ -20,15 +20,16 @@
...
-# from starlette.responses import FileResponse
+from starlette.responses import JSONResponse
 
 from core.dependencies import session_dependency, viewer_dependency
 from db import Group
 from schemas.thing import FeatureCollectionResponse
 from services.geospatial_helper import create_shapefile, get_thing_features
 from services.query_helper import simple_get_by_id
 
+# from starlette.responses import FileResponse
```
jacob-a-brown: this line can be removed rather than commented out since it is no longer used

api/geospatial.py
```diff
@@ -55,7 +56,9 @@ async def get_geospatial(
     """
 
     if format_ == "geojson":
-        return get_feature_collection(session, thing_type, group)
+        # return get_feature_collection(session, thing_type, group)
```
jacob-a-brown: this line can be removed rather than commented out since it is no longer used

schemas/series.py
jacob-a-brown, line 1: this file can be removed if it is not used (and we don't plan to use it in the future)

tests/features/steps/api_fixture.py
jacob-a-brown, line 1: Using `return` statements when adding testing data will persist that data in the DB, causing tests to no longer be independent of each other, potentially masking unintended behavior. Rather than `return`, you can `yield` the testing data, then after the `yield` statement remove those objects from the DB to keep it clean with `session.delete(object)`. Or, remove the objects at the very end of the script

```python
    session.delete(loc)
    session.delete(loc2)
    ...
```

jacob-a-brown: Or, use behave's `fixtures` to setup and teardown testing data to prevent spillover and keep tests independent of each other: https://behave.readthedocs.io/en/latest/fixtures/

tests/test_observation.py
```diff
@@ -209,6 +211,14 @@ def test_patch_water_chemistry_observation_404_wrong_activity_type(
 # ============= Get tests =================
 
 
+def test_get_transducer_observations():
+    response = client.get("/observation/transducer-groundwater-level")
+    assert response.status_code == 200
+    data = response.json()
+    assert data["total"] == 0
+    assert data["items"] == []
```
jacob-a-brown, line 219: you can make a fixture (or fixtures) in `tests/conftest.py` to test that the correct data is returned and that it's formatted as expected

jirhiker:
I will update the feature test fixtures and CI in another PR

jacob-a-brown (approved):

PR #211, opened 2025-10-28
BDMS-158: Company to Organization Mapping & Other Updates

jacob-a-brown:
### **Why**
- records from `OwnersData.Company` need be mapped to `organizations` in the lexicon table to enable the transfer to occur
- `time.tzset()` only works on Unix systems, preventing cross-platform compatability

### **How**
- map companies to organizations
- add organizations to lexicon
- comment out `time.tzset()` in `tests/__init__.py`

### **Notes**
- To prevent duplicate contacts I first if the contact exists in `transfers/waterlevels_transfer.py`. This can be kind of slow. I didn't run into any of these situations but wanted to account for the possibility. Is this necessary?
- Ethan identified `MR` as `Madeline Richards`, who was a graduate student at NMT
- Ethan and I tried to consolidate as much as possible, but he also requested departments be distinct from the entity if they both exist (e.g. `City of Santa Fe`/`City of Santa Fe WWTP` or `BLM`/`BLM Taos Office`)


transfers/contact_transfer.py
```diff
@@ -99,21 +114,26 @@ def transfer_contacts(session):
...
 
 
-def _add_first_contact(session, row, thing):
+def _add_first_contact(session, row, thing, co_to_org_mapper):
     # TODO: extract role from OwnerComment
     # role = extract_owner_role(row.OwnerComment)
     role = "Owner"
     release_status = "private"
 
     name = _make_name(row.FirstName, row.LastName)
```
kbighorse: suggestion: it's probably worth assigning a local `company` variable here, rather than keep reading from `row`

kbighorse:
suggestion: consider `zoneinfo` in place of `tzset`. We use python 3.13, so it should work.

transfers/waterlevels_transfer.py
```diff
@@ -171,6 +171,20 @@ def transfer_water_levels(session):
                 for name, organization, role in zip(*contact_info):
                     if (name, organization) in created_contacts:
                         contact = created_contacts[(name, organization)]
+                    elif (
```
jirhiker: is this necessary?  have we encountered duplicates?  Shouldn't the db prevent duplicates from being added?

transfers/data/owners_organization_mapper.json
```diff
@@ -0,0 +1,159 @@
+[
```
jirhiker: Another approach to this would be to only define an entry if `source` and `target` are different.

transfers/contact_transfer.py
```diff
@@ -44,6 +52,13 @@ def extract_owner_role(comment):
 
 def transfer_contacts(session):
 
+    co_to_org_mapper_path = get_transfers_data_path("owners_organization_mapper.json")
+    with open(co_to_org_mapper_path, "r") as f:
+        co_to_org_mapper = {}
+        co_to_org_json = json.load(f)
```
jirhiker: `owners_organization_mapper.json` should be written such that it can be loaded and assigned directly to `co_to_org_mapper`

transfers/contact_transfer.py
```diff
@@ -99,21 +114,26 @@ def transfer_contacts(session):
...
-def _add_first_contact(session, row, thing):
+def _add_first_contact(session, row, thing, co_to_org_mapper):
     # TODO: extract role from OwnerComment
     # role = extract_owner_role(row.OwnerComment)
     role = "Owner"
     release_status = "private"
 
     name = _make_name(row.FirstName, row.LastName)
 
+    if row.Company and row.Company in co_to_org_mapper:
```
jirhiker: `organization = co_to_org_mapper.get(row.Company, row.Company)`

transfers/contact_transfer.py
```diff
@@ -186,18 +206,23 @@ def _add_first_contact(session, row, thing):
             contact.addresses.append(address)
 
 
-def _add_second_contact(session, row, thing):
+def _add_second_contact(session, row, thing, co_to_org_mapper):
 
     release_status = "private"
     name = _make_name(row.SecondFirstName, row.SecondLastName)
 
+    if row.Company and row.Company in co_to_org_mapper:
```
jirhiker: `organization = co_to_org_mapper.get(row.Company, row.Company)`

.pre-commit-config.yaml
```diff
@@ -15,15 +15,15 @@ repos:
...
-  - repo: local
-    hooks:
-      - id: pytest
-        name: pytest
-        entry: pytest  # Or your specific test command, e.g., poetry run pytest
-        language: system
-        types: [python] # Specify relevant file types for your tests
-        pass_filenames: false
-        always_run: true
+  # - repo: local
```
jirhiker: you don't want tests to pass before committing?

jacob-a-brown: This was preventing me from doing incremental git commits, so I commented it out. I meant to put it back. I'll uncomment it out when I push back to this PR.

transfers/waterlevels_transfer.py
jacob-a-brown: We haven't encountered duplicates, but I put this in here as a safeguard; there are no restrictions currently for  uniqueness in the `contact` model. I'll  coordinate with @ksmuczynski about restraints to put in the DB.

jirhiker: @jacob-a-brown how much does this slow down the transfer of contacts?

jacob-a-brown:
> suggestion: consider `zoneinfo` in place of `tzset`. We use python 3.13, so it should work.

@kbighorse: according to the docs

> [ZoneInfo](https://docs.python.org/3/library/zoneinfo.html#zoneinfo.ZoneInfo) is a concrete implementation of the [datetime.tzinfo](https://docs.python.org/3/library/datetime.html#datetime.tzinfo) abstract base class, and is intended to be attached to tzinfo, either via the constructor, the [datetime.replace](https://docs.python.org/3/library/datetime.html#datetime.datetime.replace) method or [datetime.astimezone](https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone)

You already set the timezones to `UTC` via `.astimezone(timezone.utc)` in the tests. Do you recommend using `zoneinfo` instead of `timezone.utc`?

transfers/waterlevels_transfer.py
kbighorse: > there are no restrictions currently for uniqueness in the contact model

We want the model code to define uniqueness validation and fail before the database can catch it

jacob-a-brown: It didn't really slow it down that much locally, but @ksmuczynski gave me some code to ensure uniqueness on `name` and `organization`. If this proves to be an issue later on (another Jacob Brown at the Bureau! I've met a few of us now) we can change that unique constraint in the DB

transfers/data/owners_organization_mapper.json
jacob-a-brown: I opted to keep them all there to avoid extra logic, but I can write some simple if/else statement if that'd be easier/clearer

jirhiker: No change required now just something to think about

jirhiker (approved):

PR #214, opened 2025-10-29
fix: update metrics logging and improve installation date handling in transfer process

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


.gitignore
```diff
@@ -29,6 +29,8 @@ transfer*.log
 transfers/data/nma_csv_cache/*
 !transfers/data/nma_csv_cache/.gitkeep
 tests/features/*.feature
+transfers/metrics/*
+transfers/logs/*
```
jacob-a-brown, line 34: Since these two directories are now required, but we don't want to have the metrics or logs in git, you can add a file named `.gitkeep` to each of them and then add `!transfers/metrics/.gitkeep` and `!/transfers/logs/.gitkeep` to `.gitignore` so the directories will exist in git but none of the files

jirhiker: These directories are not part of the codebase and are just outputs from execution.  The directories are automatically created if they don't exist so no need to track them with git

jacob-a-brown (approved):

PR #225, opened 2025-11-04
BDMS-217: make lexicon read-only via API by deprecating modification endpoints

jirhiker:
### **Why**
- the lexicon.json is the single source of truth for ocotillo controlled vocabulary.  However it is use from a data integrity perspective to also have the lexicon terms/categories in the database to provide constraints.

### **How**
- make all patch, post, delete routes as `deprecated`
- add the `disabled_endpoint dependency`
- skipped all lexicon tests involving patch, post, delete

### **Notes**
- Use bullet points here

### DIFF
--- a/api/lexicon.py
+++ b/api/lexicon.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
-from fastapi import APIRouter, Query
+from fastapi import APIRouter, Query, HTTPException, Depends
 from fastapi_pagination.ext.sqlalchemy import paginate
 from sqlalchemy import select, func
 from sqlalchemy.exc import ProgrammingError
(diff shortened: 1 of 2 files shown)


api/lexicon.py
```diff
@@ -93,12 +93,18 @@ def database_error_handler(
     raise PydanticStyleException(status_code=HTTP_409_CONFLICT, detail=[detail])
 
 
+def disabled_endpoint():
+    raise HTTPException(status_code=404, detail="Endpoint disabled")
```
jacob-a-brown: Is this change going to be permanent? If so, we should consider using a [410 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410). Otherwise 404 is the correct one.

jacob-a-brown (approved):

PR #226, opened 2025-11-04
BDMS-220

jirhiker:
### **Why**
- Use path parameters to get Thing so that we can use thing specific responses e.g. WellResponse.  (seems like this still could be done with query parameters tho?)

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/api/thing.py
+++ b/api/thing.py
@@ -299,7 +299,6 @@ async def get_things(
     user: viewer_dependency,
     session: session_dependency,
     # thing_id: int = None,
-    thing_type: str = None,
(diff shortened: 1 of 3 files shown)


jacob-a-brown (commented):
If we are going to maintain this query parameter would you add a test to `tests/test_thing.py` to maintain coverage? Also, should there be a test per thing type that we have defined? Currently I think we only have `water well` and `spring`

jacob-a-brown (approved):

PR #227, opened 2025-11-04
[BDMS-264] Expand transfer seeder

TylerAdamMartinez:
### **Why**
- CI Testing will fail without a local DB with some data, and using the transfer script is not practical

### **How**
- Use Faker to mock dummy data

### **Notes**
- Please take note of the numbers in the code comments. This is how I think each model should be created in order.


transfers/seed.py
```diff
@@ -1,29 +1,204 @@
...
+from db.thing import Thing
+from db.sensor import Sensor
+from db.deployment import Deployment
+from db.sample import Sample
+from db.observation import Observation
+from db.parameter import Parameter
+from db.analysis_method import AnalysisMethod
+from db.regulatory_limit import RegulatoryLimit
+from db.transducer import TransducerObservation
+from db.status_history import StatusHistory
```
jacob-a-brown, line 26: All of the models are imported into `db/__init__.py`. To make access to them simpler you could do `from db import *` (I think). That way we won't have to keep track of them in multiple places.

transfers/seed.py
```diff
@@ -1,29 +1,204 @@
...
+        # 3. Parameters & Methods
+        param_names = ["pH", "Temperature", "Conductivity", "TDS", "DO"]
+        for p in param_names:
+            prm = Parameter(
+                parameter_name=p,
+                matrix="Water",
+                default_unit="pH units" if p == "pH" else "mg/L",
+            )
+            s.add(prm)
+            parameters.append(prm)
```
jacob-a-brown: This isn't needed if `MODE=development` since setting that environment variable initializes both the parameter and lexicon tables. See `core/appy.py`

transfers/seed.py
```diff
@@ -1,29 +1,204 @@
...
+                well_construction_notes=fake.sentence(),
+                well_casing_diameter=random.uniform(4, 8),
+                well_casing_depth=random.uniform(10, 50),
+            )
+            # link to random location
+            loc = random.choice(locations)
+            # mimic LocationThingAssociation (proxy)
+            t.location_associations = []  # placeholder if ORM auto-handles
+            t.locations.append(loc)
+            things.append(t)
```
jacob-a-brown: it would be good to associate contacts with things, too, via the `ThingContactAssociation` model

jirhiker (approved):

PR #230, opened 2025-11-04
well core information behave test development

jacob-a-brown:
Thoughts? Am I returning `longitude/latitude` and `easting/northing` in a correct fashion? Should I call them `point_gcs` and `point_pcs` and then include metadata for the geographic and projected coordinate systems?


tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,218 @@
...
+        == "NAD83"
+    )
+
+
+@then("the response should include the elevation in feet with vertical datum NAVD88")
+def step_impl(context):
+    data = context.response.json()
+    assert data["current_location"]["elevation"] == 2464.9
+    assert data["current_location"]["vertical_datum"] == "NAVD88"
+
```
chasetmartin: @jacob-a-brown I just want to check that elevation will be returned in feet?

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,226 @@
...
+# TODO: a new endpoint named /thing/{thing_id}/group needs to be added to the API
+# TODO: this needs to be added to the ThingResponse
+@then("the response should include the project(s) or group(s) associated with the well")
+def step_impl(context):
+    data = context.response.json()
+    assert data["groups"] == ["Collabnet"]
+
+
+# TODO: this needs to be added to the model, schema, and test data
+# TODO: how do we rectify this with the name field? Is there a better way to name this?
```
ksmuczynski: I had the same question. Could it be considered an alternate id?

ksmuczynski (approved):
Looks good! I really appreciate the TODO comments you left about the items that need to be added to the models/schemas/tests.

> Am I returning longitude/latitude and easting/northing in a correct fashion? Should I call them point_gcs and point_pcs and then include metadata for the geographic and projected coordinate systems?

Regarding naming conventions, maybe `point_wgs84` and `point_utm`? I think we decided that the coordinate transformations would occur in the schema.

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,226 @@
...
+
+# ------------------------------------------------------------------------------
+# Well Purpose and Status and Monitoring Status
+# ------------------------------------------------------------------------------
+
+
+@then("the response should include the purpose of the well (current use)")
+def step_impl(context):
+    data = context.response.json()
+    assert sorted(data["well_purposes"]) == sorted(["Domestic", "Irrigation"])
```
jirhiker: Assert "Domestic" in data["well_purpose"]

Assert "Irrigation" ...

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+# TODO: this needs to be added to the ThingResponse and thing_helper via StatusHistory
+@then(
+    "the response should include the well status of the well as the status of the hole in the ground"
+)
+def step_impl(context):
+    data = context.response.json()
+    assert data["well_status"] == "Active"
```
jirhiker: Instead of calling context.response.json() for every step let's update the "response should be json format" to set context.json_data=context.response.json()

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+# TODO: this needs to be added to the model, schema, and test data
+# TODO: the monitoring frequency field needs to be added to lexicon
+# the monitoring status field from NM_Aquifer contains a multitude of information, like having three codes (6AC), so the transfer and model/schemas will need to take this into account
+# could create descriptor table like WellPurpose and CasingMaterial
+@then("the response should include the monitoring frequency (new field)")
+def step_impl(context):
+    data = context.response.json()
+    assert data["monitoring_frequency"] == "Monthly"
```
jirhiker: They way the step is written you should assert first that "montioring_frequency" in data. Then assert that it is the correct value

tests/features/steps/well-notes.py
```diff
@@ -26,17 +26,20 @@ def step_impl(context):
     context.response = context.client.get("thing/water-well/9999")
 
 
+# TODO: this is a commonly used step, should it be moved to a common steps file?
```
jirhiker: Yes this has been moved to common.py

tests/features/steps/well-core-information.py
jacob-a-brown: Since this is going to be a commonly used step, and there are going to be other data that are commonly used, I'll call it `context.water_well_data` to distinguish it from other data that are retrieved/returned

tests/features/steps/well-core-information.py
jacob-a-brown: Would the `alternate_organization` be `NMBGMR`? Since this is a name that we assign

jacob-a-brown:
The feature file says the the response should include `latitude` and `longitude`, which leads me to think that there should be fields with those names. I can check with folks at standup today to see if that's their intent

chasetmartin:
> The feature file says the the response should include `latitude` and `longitude`, which leads me to think that there should be fields with those names. I can check with folks at standup today to see if that's their intent

I'm updating the feature file to specify the use of geojson for consistency in the app.

tests/features/steps/well-core-information.py
jacob-a-brown: @chasetmartin if this gets moved to alternate ids should this feature be removed and "site_name" gets added to the feature `And the response should include any alternate IDs for the well like the USGS site number or the OSE well ID and OSE well tag ID`?

chasetmartin: That makes sense, let me update it.

chasetmartin: @jacob-a-brown Ok I moved site_name to the alternate identifiers

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,252 @@
...
+# the monitoring status field from NM_Aquifer contains a multitude of information, like having three codes (6AC), so the transfer and model/schemas will need to take this into account
+# could create descriptor table like WellPurpose and CasingMaterial
+@then(
+    "the response should include whether the well is currently being monitored with status text if applicable (from previous status field)"
+)
+def step_impl(context):
+    assert "is_being_monitored" in context.water_well_data
+    assert "monitoring_status" in context.water_well_data
+    assert context.water_well_data["is_being_monitored"] == True
+    assert context.water_well_data["monitoring_status"] == "Active"
```
chasetmartin: I will remove the "is being monitored" from the feature so that only monitoring_status is reflected

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,252 @@
...
+    assert "Irrigation" in context.water_well_data["well_purposes"]
+
+
+# TODO: this needs to be added to the ThingResponse and thing_helper via StatusHistory
+@then(
+    "the response should include the well status of the well as the status of the hole in the ground"
+)
+def step_impl(context):
+    assert "well_status" in context.water_well_data
+    assert context.water_well_data["well_status"] == "Active"
```
chasetmartin: I will update the feature file to reflect that this is the previous WellData.Status field

tests/features/steps/well-core-information.py
```diff
@@ -0,0 +1,252 @@
...
+def step_impl(context):
+    assert "monitoring_frequency" in context.water_well_data
+    assert context.water_well_data["monitoring_frequency"] == "Monthly"
+
+
+# TODO: this needs to be added to the model, schema, and test data
+# the monitoring status field from NM_Aquifer contains a multitude of information, like having three codes (6AC), so the transfer and model/schemas will need to take this into account
+# could create descriptor table like WellPurpose and CasingMaterial
+@then(
+    "the response should include whether the well is currently being monitored with status text if applicable (from previous status field)"
```
chasetmartin: I will update the feature file to that this reflects the previous MonitoringStatus field, not Status.

tests/features/steps/well-core-information.py
```diff
@@ -165,9 +215,13 @@ def step_impl(context):
     'the response should include a geometry object with type "Point" and coordinates array [longitude, latitude, elevation] in decimal degrees with datum WGS84'
 )
 def step_impl(context):
```
jirhiker, line 217: there should be an assertion that the x and y are actually latitude and longitude.  -90, 90, -180,180. 

there should also be an assertion that the horizontal datum is WGS84

jacob-a-brown: I saw this after merging into our shared pre-staging branch. Here are asserts I can use:

```python
assert context.water_well_data["current_location"]["crs"] == {
        "type": "name",
        "properties": {
            "name": "EPSG:4326",
        },
    }

assert longitude >= -180 and longitude <= 180
assert latitude >= -90 and latitude <= 90
```

From what I've read `crs` has been deprecated and is no longer used.[ GeoJSONs must be in WGS84](https://datatracker.ietf.org/doc/html/rfc7946#section-4). Because of this, is it worth having the `crs` field? Or should it just be assumed and the frontend will just know and display that as the horizontal datum? @chasetmartin do you want the string `WGS84` in the response somehow?

chasetmartin: If the proper way to define it in geojson properties is EPSG:4326 that seems fine.

jacob-a-brown: I can put `crs` in there, but according to the specs all GeoJSONs have to be in WGS84

>   The coordinate reference system for all GeoJSON coordinates is a
 geographic coordinate reference system, using the World Geodetic System 1984 (WGS 84) [[WGS84](https://datatracker.ietf.org/doc/html/rfc7946#ref-WGS84)] datum, with longitude and latitude units of decimal degrees.

Given that _must_ be the case, should I still include `crs`? It has been used in the past, but is now deprecated.

chasetmartin: Ah I see what you're saying. If crs is deprecated in GeoJSON spec I don't think we should include it. But since we have horizontal_datum for UTM coords in their properties, I think we should have horizontal_datum within these properties that identifies WGS84, just to be explicit with the client.

jacob-a-brown: @jirhiker because all GeoJSONs use WGS84 and the coordinates are decimal degrees the feature file was updated to

```
# Location Information
# GeoJSON spec format RFC 7946 (Aug 2016) requires coordinates to be decimal degrees in WGS84
And the response should include location information in GeoJSON spec format RFC 7946
And the response should include a geometry object with type "Point" and coordinates array [longitude, latitude, elevation]
```

The note was included for documentation purposes

PR #232, opened 2025-11-05
[BDMS-264] Expand transfer seeder

TylerAdamMartinez:
### Why
This PR addresses the following problem/context:
• CI Testing will fail without a local DB with some data, and using the transfer script is not practical
### How
Implementation summary - the following was changed/added/removed:
• Use Faker to mock dummy data
### Notes
• Please take note of the numbers in the code comments. This is how I think each model should be created in order.

### DIFF
--- a/transfers/seed.py
+++ b/transfers/seed.py
@@ -18,6 +18,7 @@ from db.location import Location, LocationThingAssociation
 from db.thing import Thing
 from db.sensor import Sensor
 from db.deployment import Deployment
+from db.field import FieldEvent, FieldActivity
 from db.sample import Sample
 from db.observation import Observation
 from db.parameter import Parameter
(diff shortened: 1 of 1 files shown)


TylerAdamMartinez:
NO, NOT DELETE THE BRANCH SINCE THE FRONTEND STILL CALLS IT FOR CI/CD AND WILL CONTINUE TO DO SO UNTIL MERGED INTO STAGING.

ksmuczynski (approved):
The order in which the models are created looks good to me. I think the Contact or Location models could be created first, either way should work. The model numbers skip from 2 to 5, but I'm assuming that's just a typo; I don't think any additional core models are needed.

chasetmartin (approved):
lgtm

PR #234, opened 2025-11-06
NO TICKET: make pytest and behave interoperable

jacob-a-brown:
### **Why**

I'm running into errors seeding the testing data for `behave` tests. This occurs after I run `pytest` (either on its own or for `git commit`), which erases and repopulates the database for the `pytest` tests. It does not, however, teardown and build the database back up. Because of this, the `id`s of all tables get incremented, so when `behave` testing data are added that depend on a hardcoded `id` (sensor, deployment) an error occurs because that `id` is out of date.

### **How**
- use objects and their auto-generated IDs to prevent hardcoding values

### **Notes**
- Because of this flexible implementation, I can no longer check if an object exists via hardcoded `id`s. Due to this, unique constraints are broken. To address this, I tear down all of the testing data in `after_all`. This should be quicker and easier than running `erase_and_rebuild_db` at the start of every `behave` testing session

### DIFF
--- a/tests/features/environment.py
+++ b/tests/features/environment.py
@@ -15,6 +15,7 @@
 # =============== ================================================================
 import random
 from datetime import datetime, timedelta
+from sqlalchemy import text
 
 from core.initializers import erase_and_rebuild_db, init_lexicon, init_parameter
 from db import (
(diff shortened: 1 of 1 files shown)


jirhiker:
I don't understand why this is preferable. What issues were you having seeding the database. The current solution is idempotent

jacob-a-brown:
When `pytest` is run it creates a bunch of data, thereby auto-incrementing the `id` in each table. Then, when `behave` is run it adds the different objects. Because of the auto-incrementing from `pytest` the hardcoded `id`s are out of date. This causes two issues with the data that is being seeded:

1. if `sh run_bdd.sh` is run more than once, it will try to add a second sensor with the same `serial_no`, which breaks the unique constraint.
2. the hardcoded `thing_id` and `sensor_id` in the deployment are no longer valid.

jacob-a-brown:
By keeping it flexible and tearing down the data after each test we won't have to worry about `id`s and the order in which tests are run (`pytest` or `behave`), and during development we can call `sh run_bdd.sh` any number of times and the unique constraints won't be broken.

jirhiker:
I don't see the point of removing seed data. Removing data added by a test is acceptable, but removing seed data solely to re-add it does not make sense. The seed functions need to be idempotent (and for the most part, they are) without requiring the removal of data add the end of a test suite.

jirhiker (approved):
I will approve this to keep this moving forward. I don't see that as a necessary solution now but it seems to address the issues you are having

PR #235, opened 2025-11-06
feat: add transducer observation handling and improve database initialization

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/services/observation_helper.py
+++ b/services/observation_helper.py
@@ -53,7 +53,10 @@ def get_transducer_observations(
     filter_: str = Query(alias="filter", default=None),
 ):
     if thing_id:
-        simple_get_by_id(session, Thing, thing_id)
+        item = session.get(Thing, thing_id)
+        if item is None:
+            empty_query = select(TransducerObservation).where(False)
+            return paginate(query=empty_query, conn=session)
 
     # Subquery to get latest block for each observation
     block_subq = (
--- a/services/observation_helper.py
+++ b/services/observation_helper.py
@@ -52,6 +52,9 @@ def get_transducer_observations(
     order: str | None = None,
     filter_: str = Query(alias="filter", default=None),
 ):
+    if thing_id:
+        simple_get_by_id(session, Thing, thing_id)
+
     # Subquery to get latest block for each observation
     block_subq = (
         select(TransducerObservationBlock.id)
--- a/tests/features/steps/transducer.py
+++ b/tests/features/steps/transducer.py
@@ -24,7 +24,7 @@ from db.engine import session_ctx
 def step_impl(context):
     with session_ctx() as session:
         sql = select(Thing).where(Thing.thing_type == "water well")
-        wells = session.execute(sql).scalars().all()
+        wells = session.execute(sql).unique().scalars().all()
         assert len(wells) > 0, "No wells found in db"
 
         sql = select(TransducerObservation)
@@ -43,7 +43,7 @@ def step_impl(context):
 def step_impl(context):
     context.response = context.client.get(
         "/observation/transducer-groundwater-level",
-        params={"thing_id": context.well.id},
+        params={"thing_id": context.objects["wells"][0].id},
     )
 
 
@@ -56,7 +56,7 @@ def step_impl(context):
 @then("each transducer data entry should include a timestamp, value, status")
 def step_impl(context):
     data = context.response.json()
-    items = data[0]
+    items = data["items"][0]
     item = items["observation"]
     block = items["block"]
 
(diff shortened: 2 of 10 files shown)


tests/features/steps/transducer.py
jacob-a-brown, line 1: With the fixture-like implementation you can retrieve the database objects from the `context` object, like `context.objects["wells"][0]` to get the first well that was added or `context.objects["blocks"][0]` to get the first block of transducer observations that were added for the step `"the system has valid well and transducer data in the database"`

tests/features/steps/transducer.py
```diff
@@ -85,7 +87,7 @@ def step_impl(context):
 
 @then('the status should be one of "Draft", "Corrected"')
 def step_impl(context):
-    assert context.status in ("not reviewed",)
+    assert context.status in ("not reviewed",), f"Unexpected status: {context.status}"
```
jacob-a-brown: shouldn't this be `context.status in ("Draft", "Corrected")` to correspond with the the step definition?

jirhiker: it should be. You've identified an issue that Ive been thinking about. How do you keep step implementations from "lying" about their assertions.

jacob-a-brown: I think your revision takes care of it. It asserts that it's one of two values, otherwise it sense the message of unexpected status 👍

services/observation_helper.py
```diff
@@ -52,6 +52,9 @@ def get_transducer_observations(
     order: str | None = None,
     filter_: str = Query(alias="filter", default=None),
 ):
+    if thing_id:
+        simple_get_by_id(session, Thing, thing_id)
+
```
jacob-a-brown, line 60: This raises a 404 error if `thing_id` does not exist, but since `thing_id` is a query parameter if there are no observations for that `thing` (one of the reasons could be that the `thing_id` does not exist), it should just return an empty list.

I think that 404 errors should only be returned for path parameters since the full path is a resource.

jirhiker: That is not how the feature file is written. https://github.com/DataIntegrationGroup/OcotilloBDD/blob/main/features/backend/transducer-data-response.feature

```
When the user requests transducer data for a non-existing well
      Then the system should return a 404 status code
```

jirhiker: Remaining faithful to the feature file, it seems like making thing_id a path parameter is the best approach. In the case of using a query parameter, I don't think it makes sense to get an empty list if the thing_id is invalid.  

I think internal int ids should be path parameters
`get observation/transducer-groundwater-level/123`

Human-readable identifiers like name, pointid, site-name, etc, should be query parameters. e.g.

`get observation/transducer-groundwater-level?thing.name == 'WL-001' `
but we don't have a need for this functionality at this time.

jacob-a-brown: I wrote up a discussion about this a while ago to record my thoughts: https://github.com/orgs/DataIntegrationGroup/discussions/14

The empty list vs 404 error came up when @TylerAdamMartinez and I were working on AMP API and the water level entry form a while ago. I can look more into the appropriate response when a user supplies an erroneous query parameter.

I also agree that path parameter `id`s should pertain to that table

jirhiker:
@jacob-a-brown can we get this approved and merged

jacob-a-brown:
Would you be okay to change the feature file to return an empty list when a `thing_id` is provided as a query parameter but it doesn't exist in the database? Then filter by `thing_id` instead of using `simple_get_by_id`, which raises a 404 error. From what I've read that's the standard practice. 

https://softwareengineering.stackexchange.com/questions/203492/when-to-use-http-status-code-404-in-an-api
https://stackoverflow.com/questions/30217761/http-200-or-404-for-empty-list
https://www.reddit.com/r/learnprogramming/comments/1lirdwc/for_a_rest_api_fetch_with_parameters_should_you/
https://www.servicenow.com/community/developer-forum/404-error-in-case-of-rest-api-not-returning-any-data/m-p/3087325
https://chatgpt.com/share/e/6915fe67-288c-800a-a9c7-bef9f0c548c8

jirhiker:
I'm good with that pattern for other endpoints. I removed the functionality of using thing_id as a query parameter for transducer observations for this iteration as there isn't a use case for it yet

jacob-a-brown:
I missed that, I was still reading the `thing_id` as a query parameter. Does it make sense to have `thing_id` be the path parameter for a table for which it is not the ID? I thought that the IDs for path parameters pertained to the table itself, not related tables

jirhiker:
It's a good point. It's not something that matters in practice but for consistency I will change back to a query parameter

jacob-a-brown (approved):

PR #237, opened 2025-11-07
BDMS 227: well additional information tests

jacob-a-brown:
this is a first draft, i'm sure more will change. i also need to add to the testing data, but i figured it'd be good to get eyes on this in the meantime

PR #238, opened 2025-11-10
BDMS 221 & 227: function to retrieve polymorphic table record

jacob-a-brown:
### **Why**
- Records from the `parameter` and `status_history` tables need to be easily retrieved for schemas and tests

### **How**
- A function to retrieve a record was put into `services/util.py`

### **Notes**
- This only retrieves one function in a sorted list by `start_date`. I don't think that we need one to retrieve _all_ records since that is easily done via the attribute from the mixins.

### DIFF
--- a/services/util.py
+++ b/services/util.py
@@ -116,15 +116,16 @@ def get_epqs_elevation_from_point(lon: float, lat: float) -> float | None:
     return data["value"]
 
 
-def retrieve_polymorphic_table_record(
+def retrieve_latest_polymorphic_table_record(
     target_record: Base,
     polymorphic_relationship: str,
     polymorphic_type: str,
-    latest=True,
 ) -> Base:
     """
-    Retrieve a record from a polymorphic table. This function assumes that the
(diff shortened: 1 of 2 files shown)


jirhiker (commented):
where is this being used

jacob-a-brown:
I'll use it in tests. I initially thought I'd use it throughout, but it may not actually be needed (I'm still developing responses, so TBD). From [your comment on the well additional info PR](https://github.com/DataIntegrationGroup/NMSampleLocations/pull/237#discussion_r2505837705) I figured I'd put it in `services/util.py`, but if it's just used in tests should it go in `tests/__init__.py`?

jacob-a-brown:
I need to add a check in there, too, that `end_date is null`

jacob-a-brown:
I updated this to make sure `end_date` is not null, but since `status_history` is eagerly loaded for each `thing` object I don't need to use `polymorphic_relationship` in the API. So I've moved it to `tests/__init__.py` for use in tests only.

jirhiker (approved):

PR #240, opened 2025-11-10
BDMS 221: updates to pass tests

jacob-a-brown:
the updates herein enable the tests to pass all tests except for `@then("the response should include the source of the well depth information")` since the polymorphic source table is still in development. I've made changes in a variety of places. The only changes I made to models is to fix some lexicon terms and add `@property` fields to the `thing` model. Otherwise the work has been done in schemas and the testing directory


schemas/thing.py
```diff
@@ -153,6 +164,10 @@ class WellResponse(BaseThingResponse):
     well_casing_depth_unit: str = "ft"
     well_casing_materials: list[CasingMaterial] = []
     well_construction_notes: str | None = None
+    well_status: str | None
+    measuring_point_height: float
```
TylerAdamMartinez, line 196:
```suggestion
    measuring_point_height: float | None
```

jacob-a-brown: `measuring_point_height` is required for every water level measurement to get depth of water below ground surface. Why do you think it should be optional?

jacob-a-brown:
@ksmuczynski and @jirhiker, @TylerAdamMartinez came across an error that I didn't anticipate:

I changed [line 183](https://github.com/DataIntegrationGroup/NMSampleLocations/pull/240/files#diff-b1c745f4d17f56d8b49d188f963dc26dbb37e449f233e3d7dd0fc5e8700fdf49R138) to `alternate_organization: Organization` (a restricted list of enums). I made this update because of the name of the field. In the script `transfers/link_id_transfer.py`, however, `PLSS` data is being added as alternate IDs and the `alternate_organization` is being set to `PLSS`. This is not an organization in our lexicon term table. My questions are:

1. Should the field `alternate_organization` be restricted to the enums? If so, should `PLSS` be added to as an organization to the lexicon (even though it isn't really)?
2. Should the name of the field `alternate_organization` change so that `PLSS` can be added? If so, should the field be a string or an enum?

jacob-a-brown:
@jirhiker @chasetmartin @ksmuczynski 

One of the feature steps is

services/util.py
```diff
@@ -26,6 +26,10 @@ def transform_srid(geometry, source_srid, target_srid):
     return transform(transformer.transform, geometry)
 
 
+def convert_m_to_ft(meters: float) -> float:
+    return meters * 3.28084
+
```
TylerAdamMartinez, line 44:
```
METERS_TO_FEET = 3.28084

def convert_m_to_ft(meters: Optional[float]) -> Optional[float]:
    """Convert a length from feet to meters."""
    if meters is None:
        return None
    return round(meters * METERS_TO_FEET, 6)

def convert_ft_to_m(feet: Optional[float]) -> Optional[float]:
    """Convert a length from feet to meters."""
    if feet is None:
        return None
    return round(feet / METERS_TO_FEET, 6)

```

Rounding isn't required, but it's a way to avoid floating-point artifacts like `3.2808399999999997`.

jirhiker:
> @ksmuczynski and @jirhiker, @TylerAdamMartinez came across an error that I didn't anticipate:
> 
> I changed [line 183](https://github.com/DataIntegrationGroup/NMSampleLocations/pull/240/files#diff-b1c745f4d17f56d8b49d188f963dc26dbb37e449f233e3d7dd0fc5e8700fdf49R138) to `alternate_organization: Organization` (a restricted list of enums). I made this update because of the name of the field. In the script `transfers/link_id_transfer.py`, however, `PLSS` data is being added as alternate IDs and the `alternate_organization` is being set to `PLSS`. This is not an organization in our lexicon term table. My questions are:
> 
> 1. Should the field `alternate_organization` be restricted to the enums? If so, should `PLSS` be added to as an organization to the lexicon (even though it isn't really)?
> 2. Should the name of the field `alternate_organization` change so that `PLSS` can be added? If so, should the field be a string or an enum?

For the sake of expediency add PLSS as an `alternate_organization` to the lexicon

services/util.py
jacob-a-brown: good call. this change has been pushed to the PR.

ksmuczynski (approved):

schemas/thing.py
TylerAdamMartinez: Because the transfer script results in some rows having a NULL in the `measuring_point_height` field, leading to a 500 error when I request it.

jacob-a-brown: Got it 👍 . The updates I have made to the transfer scripts now require a `measuring_point_height` for every well. That's in another PR... (PR #242 )

PR #241, opened 2025-11-10
BDMS-195:  jir-kas-notes-model

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


db/notes.py
```diff
@@ -0,0 +1,78 @@
...
+    # These are viewonly relationships used by the 'target' property below.
+    _thing_target: Mapped["Thing"] = relationship(
+        "Thing",
+        primaryjoin="and_(foreign(Notes.notable_id) == Thing.id, Notes.notable_type == 'Thing')",
+        viewonly=True,
+    )
+    _location_target: Mapped["Location"] = relationship(
+        "Location",
+        primaryjoin="and_(foreign(Notes.notable_id) == Location.id, Notes.notable_type == 'Location')",
+        viewonly=True,
```
jacob-a-brown: The `NotesMixin` provides a way for each table that inherits from the mixin to access its notes easily. Should we maintain these separate relationships? Or would it make it more complicated and difficult to maintain if we did?

schemas/notes.py
```diff
@@ -0,0 +1,46 @@
...
+
+# -------- RESPONSE ----------
+class NoteResponse(BaseResponseModel, BaseNote):
+    """
+    Response schema for Note details.
+    """
+
+    notable_id: int
+    notable_type: str
+
```
jacob-a-brown, line 37: should this contain the content?

schemas/notes.py
```diff
@@ -0,0 +1,46 @@
...
+# -------- CREATE ----------
+class CreateNote(BaseCreateModel, BaseNote):
+    # TODO: this was a suggestion by AI, but based on our other schemas it
+    # seems like more should be added here...
+    """
+    Schema for creating a new Note. The parent object's ID and type will be
+    taken from the URL path, not the request body.
+    """
+    pass
```
jacob-a-brown, line 26: should this contain the content and notable type?

services/crud_helper.py
```diff
@@ -35,10 +36,22 @@ def model_adder(session, table, model, user=None, **kwargs):
         md["created_by_id"] = user["sub"]
         md["created_by_name"] = user["name"]
 
+    notes = None
+    if issubclass(table, NotesMixin):
```
jacob-a-brown, line 40: The `NotesMixin` is to add the `notes` attribute to a table. Should this be `if issubclass(table, Notes)`?

services/crud_helper.py
```diff
@@ -60,7 +73,17 @@ def model_patcher(
     """
 
     for key, value in payload.model_dump(exclude_unset=True).items():
-        setattr(item, key, value)
+        if isinstance(item, NotesMixin) and key == "notes":
```
jacob-a-brown, line 77: The `NotesMixin` is to add the notes attribute to a table. Should this be `if issubclass(table, Notes)`?

jirhiker: no. this is to check whether you are trying to update a table with the `notes` via e.g. a model that inherits NotesMixin

services/crud_helper.py
jirhiker: no. this is to check whether you are trying to add notes to a table with notes e.g. a model that inherits NotesMixin

schemas/notes.py
jirhiker: inherited from BaseNote

schemas/notes.py
jirhiker: currently the api figures out the notable_type

db/notes.py
jirhiker: @ksmuczynski ☝️

db/notes.py
```diff
@@ -0,0 +1,78 @@
...
+    # notable_type: Mapped[str] = lexicon_term(
+    #     nullable=False,
+    #     comment="The type of the note associated with this record.",
+    # )
+
+    # --- Columns ---
+    note_type: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="A controlled vocabulary field that defines the specific category of the note (e.g. 'Access Instructions`, ",
+    )
```
jacob-a-brown: @ksmuczynski and I had a discussion about code style and how to make it easily understandable and usable across the different polymorphic tables. We thought that instead of `statusable_id`, `permissible_id`, `notable_id` ... we should use a standard style for every table. What we landed on for naming convention was:

`target_id`
`target_table`

Do you agree with this?

jacob-a-brown: Thoughts? This is being implemented for `StatusHistory` and `Permissions` (which is going to be renamed `PermissionHistory`

db/base.py
```diff
@@ -210,6 +215,56 @@ def permissions(self):
...
+        # All parent tables use 'id' as their primary key.
+        pk_name = "id"
+
+        return relationship(
+            "Notes",
+            primaryjoin=f"and_({cls.__name__}.{pk_name}==foreign(Notes.notable_id), "
+            f"Notes.notable_type=='{cls.__name__}')",
+            lazy="selectin",
+            viewonly=True,
+        )
```
jacob-a-brown: Because the model names are different in SQL Alchemy and the database, `CamelCase` and `snake_case`, this relationship will break for tables with more than one word in their names (e.g. `FieldEvent`, `FieldActivity`). To fix this, change the relationship to:

```python
@declared_attr
def notes(cls):
    return relationship(
        "Notes",
        primaryjoin=and_(
            cls.id == foreign(Notes.target_id),
            Notes.target_table == pascal_to_snake(cls.__name__),
        ),
        cascade="all, delete-orphan",
        lazy="selectin",
    )
```

Because this uses the `Notes` table this mixin will need to be moved to `db/notes.py` to prevent circular imports.

jacob-a-brown (approved):

PR #242, opened 2025-11-11
BDMS 221: updates to transfer scripts

jacob-a-brown:
this PR is to update transfer scripts for the changes made to models/schemas/API as we implement feature tests. This is dependent on PR #240 being merged.


db/thing.py
```diff
@@ -274,6 +280,48 @@ def current_location(self):
             else None
         )
 
+    @property
```
jirhiker, line 308:
```
def _get_status(self, status_type):
        entries = [status for status in self.status_history 
                   if status.status_type == status_type and status.end_date is None]
        if entries:
            entry = sorted(entries, key=lambda x: x.start_date, reverse=True)[0]
            return entry.status_value

    @property
    def well_status(self) -> str | None:
        """
        Returns the well status from the most recent status history entry
        where status_type is "Well Status".

        Since status_history is eagerly loaded, this should not introduce N+1 query issues.
        """
        return self._get_status("Well Status")

    @property
    def monitoring_status(self) -> str | None:
        """
        Returns the monitoring status from the most recent status history entry
        where status_type is "Monitoring Status".

        Since status_history is eagerly loaded, this should not introduce N+1 query issues.
        """
        return self._get_status("Monitoring Status")
```

schemas/thing.py
```diff
@@ -46,6 +48,26 @@ def check_depths(self):
             raise ValueError(
                 "well casing depth must be less than or equal to hole depth"
             )
+        elif (
```
jirhiker: improve readability 
```python
if self.measuring_point_height is not None:
    if self.hole_depth is not None and self.measuring_point_height>=self.hole_depth:
        raise ValueError("measuring point height must be less than hole depth")
    if self.well_casing_depth is not None and self.measuring_point_height>=self.well_casing_depth:
        raise ValueError("measuring point height must be less than well casing depth")
    if self.well_depth is not None and self.measuring_point_height>=self.well_depth:
        raise ValueError("measuring point height must be less than well depth")```

File: db/thing.py (hunk 1 above)
Comment (jirhiker) [reply db/thing.py:308]: or use `retrieve_latest_polymorphic_table_record`

**Review by jirhiker — CHANGES_REQUESTED** [review]
looking pretty good

File: db/thing.py (hunk 1 above)
Comment (jacob-a-brown) [reply db/thing.py:308]: Good call - that simplifies things. I made it so that `None` is returned if there are no associated polymorphic records (I ran into errors because of this). I also moved the function into `services/util.py` since it's now being used throughout the repo, not just tests.

File: transfers/group_transfer.py [hunk 3]
```diff
@@ -20,6 +20,7 @@
 from db.engine import session_ctx
 from transfers.util import read_csv
 from transfers.logger import logger
+from tests import retrieve_latest_polymorphic_table_record
```
TylerAdamMartinez: Where is this coming from?

jacob-a-brown: I moved this to `services/util.py` and forgot to update. Good catch

TylerAdamMartinez (approved):
lgtm

transfers/util.py
```diff
@@ -318,22 +335,56 @@ def make_location(row: pd.Series) -> Location:
...
-    )
-    return location
+    if row.CoordinateMethod or row.CoordinateAccuracy:
+        coordinate_method = (
+            lexicon_mapper.map_value(f"LU_CoordinateMethod:{row.CoordinateMethod}")
+            if not pd.isna(row.CoordinateMethod)
+            else None
+        )
+
+        if row.CoordinateAccuracy == "5m":
```
jirhiker: Do not use long if elif blocks for this purpose. A dict is a more elegant, maintainable, and readable solution

transfers/well_transfer.py
```diff
@@ -259,14 +294,139 @@ def transfer_wells(session: Session, flags: dict = None, limit: int = 0) -> None
...
+                start_date=datetime.now(tz=UTC),
+                target_id=target_id,
+                target_table=target_table,
+            )
+            session.add(status_history)
+            logger.info(
+                f"  Added monitoring status for well {well.name}: {status_value}"
+            )
+
+            if "6" in row.MonitoringStatus:
```
jirhiker: This should be a loop

jirhiker (approved):
2 non blocking / technical debt/style comments

ksmuczynski (approved):

transfers/util.py
jacob-a-brown: this update will get pushed to the PR

PR #244, opened 2025-11-12
BDMS-221-225: New measuring_point_history model

ksmuczynski:
### **Why**
- This PR addresses the need to track a well's official measuring point height over time.
- The purpose of this PR is to make these intermediate updates available to @jacob-a-brown as we continue to co-work on the main BDMS-221 task.

### **How**
- Created a new model, `measuring_point_history`, to store the official measuring point and description for a Thing.
- A new One-To-Many relationship named `measuring_points` was added to the `Thing` model. This addresses the fact that a Thing may have multiple measuring points over time.

### **Notes**
- This PR does NOT address updates to the schemas, tests, or transfer scripts.

### DIFF
--- a/db/measuring_point_history.py
+++ b/db/measuring_point_history.py
@@ -1,0 +1,61 @@
+"""
+SQLAlchemy model for the MeasuringPointHistory table.
+
+This table stores the authoritative MP height of a Thing from
+construction or modification events. It provides a complete, auditable
+history of the official, surveyed measuring point (MP) descriptions
+and heights for a Thing.
+
+This table is not for storing routine field checks of the
+MP height (which are stored on the `Observation` table). This table should
+only be updated when a well is first installed, physically modified
+(e.g., a new wellhead is installed), or officially re-surveyed.
+"""
+
+from typing import TYPE_CHECKING
+
+from sqlalchemy import Integer, ForeignKey, Date, Text, Numeric
+from sqlalchemy.orm import relationship, Mapped, mapped_column
+
+from db.base import Base, AutoBaseMixin, ReleaseMixin
+
+if TYPE_CHECKING:
+    from db.thing import Thing
+
+
+class MeasuringPointHistory(Base, AutoBaseMixin, ReleaseMixin):
+    """
+    Represents a single, authoritative, time-stamped record of a
+    Thing's measuring point description and height.
+    """
+
+    # --- Foreign Keys ---
+    thing_id: Mapped[int] = mapped_column(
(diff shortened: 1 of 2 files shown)


db/measuring_point_history.py
```diff
@@ -0,0 +1,67 @@
...
+        Integer, ForeignKey("thing.id", ondelete="CASCADE"), nullable=False
+    )
+
+    # --- Columns ---
+    measuring_point_height: Mapped[float] = mapped_column(
+        Numeric,
+        nullable=False,
+        comment="The official, surveyed height of the measuring point relative to ground surface (in feet).",
+    )
+    mp_description: Mapped[str] = mapped_column(
```
jacob-a-brown: can this be called `measuring_point_description` since it's called `measuring_point_height` above?

db/measuring_point_history.py
```diff
@@ -0,0 +1,67 @@
...
+
+    reason: Mapped[str] = mapped_column(
+        Text,
+        nullable=True,
+        comment="Describes the reason for the new or updated measuring point (e.g., 'A new wellhead was installed').",
+    )
+
+    # --- Relationships ---
+    # Many-To-One: A description history record belongs to one Thing. Many history records may belong to a single Thing.
+    thing: Mapped["Thing"] = relationship("Thing", back_populates="mp_history")
```
jacob-a-brown: `back_populates` should be `measuring_points` to correspond with how it is defined in the `thing` table.

db/measuring_point_history.py
jacob-a-brown, line 1: since this model only pertains to the `thing` model I don't think that it needs to go into its own file. It can go into `db/thing.py`

ksmuczynski: Happy to do this. Just curious, how are we deciding which models get their own files? When I'm looking for info about a table/model, I'm expecting it to be listed in the `db` directory.

jacob-a-brown: I don't know if we have a standard, but when making a new model I put it into an existing file if it only pertains to other models in that file (e.g. `WellPurpose` and `WellCasingMaterial`). [Now that I think of it, I don't know if I've created a standalone model, but it makes sense for me to have the polymorphic models be in their own files for ease of understanding.]

ksmuczynski: Curious what @jirhiker thoughts are.

jacob-a-brown: @jirhiker I merged this so I can use develop off of it - where the table is defined in the codebase won't affect how it's used in the schemas/tests. If you think it should be in the `db/thing.py` file I can move it there easily.

PR #245, opened 2025-11-12
BDMS-221-225: Polymorphic DataProvenance model

ksmuczynski:
### **Why**
- Centralizes provenance metadata for foundational/static records to improve data traceability.
- Enables models to associate provenance information efficiently using reusable mixins.
- Refines location and thing models to support provenance tracking and standardized metadata.
- The purpose of this PR is to make these intermediate updates available to @jacob-a-brown as we continue to co-work on the main BDMS-221 task.

### **How**
- Added new DataProvenance model in db/data_provenance.py to centralize provenance metadata for foundational/static records.
- Introduced DataProvenanceMixin in db/base.py for reusable polymorphic relationships, enabling models to associate provenance efficiently.
- Added DataProvenanceMixin to the Location and Thing models to support provenance tracking for location and thing records.
- Enhanced field definitions, relationships, and documentation in all affected files for clarity and maintainability.

### **Notes**
- Lexicon categories and terms for provenance fields need to be added to cover all relevant controlled vocabularies.
- Further integration with other models may be required as provenance tracking is extended.


db/data_provenance.py
```diff
@@ -0,0 +1,103 @@
...
+    _thing_target: Mapped["Thing"] = relationship(
+        "Thing",
+        primaryjoin="and_(foreign(DataProvenance.target_id) == Thing.id, DataProvenance.target_table == 'Thing')",
+        viewonly=True,
+    )
+    _location_target: Mapped["Location"] = relationship(
+        "Location",
+        primaryjoin="and_(foreign(DataProvenance.target_id) == Location.id, DataProvenance.target_table == 'Location')",
+        viewonly=True,
+    )
```
jacob-a-brown, line 87: because the table uses `snake_case` in the database I think that it should be `DataProvenance.target_table == 'thing')` (and the same with `location`)

db/base.py
```diff
@@ -210,6 +210,25 @@ def permissions(self):
...
+        # One-to-Many polymorphic relationship
+        return relationship(
+            "DataProvenance",
+            primaryjoin=f"and_({self.__name__}.id==foreign(DataProvenance.target_id), "
+            f"DataProvenance.target_table=='{self.__name__}')",
+            lazy="selectin",
+            viewonly=True,
+        )
+
+
```
jacob-a-brown: To future proof this and allow for `CamelCase` to `snake_case` models --> table names this should be moved to `db/data_provenance.py` and be

```python
from db.base import pascal_to_snake
from sqlalchemy import and_
from sqlalchemy.orm import declared_attr, relationship, foreign


class DataProvenanceMixin:
    """
    Mixin for models that can have data provenance records (e.g., Thing, Location).
    It automatically creates a polymorphic One-to-Many relationship to the
    DataProvenance table.
    """

    @declared_attr
    def data_provenance(cls):
        # One-to-Many polymorphic relationship
        return relationship(
            "DataProvenance",
            primaryjoin=and_(
                cls.id == foreign(DataProvenance.target_id),
                DataProvenance.target_table == pascal_to_snake(cls.__name__)
            ),
            lazy="selectin",
            viewonly=True,
        )
```

db/data_provenance.py
```diff
@@ -0,0 +1,103 @@
...
+    # TODO: Values from the following NMAquifer tables should be included as terms in the lexicon: 'LU_CoordinateAccuracy'.
+    accuracy_value: Mapped[float] = mapped_column(
+        nullable=True, comment="A numeric value representing the data's accuracy."
+    )
+    # TODO: Values from the following NMAquifer tables should be included as terms in the lexicon: 'LU_CoordinateAccuracy'.
+    accuracy_unit: Mapped[str] = lexicon_term(
+        nullable=True,
+        comment="The unit for the `accuracy_value` (e.g., 'meters', 'feet').",
+    )
```
jacob-a-brown, line 73: these fields are used for a small subset of fields in a subset of tables. should these move to those tables, rather than be here? since they won't apply to a number of fields for which `DataProvenance` will be used.

ksmuczynski: Maybe, but I saw this as one of the advantages of the Provenance model - all the sparse, optional, and evolving metadata is organized in one central place. If we move these fields to the Location table we'd have to add even more fields (`coordinate_accuracy`, `coordinate_collection_method`, `coordinate_accuracy_value`, `coordiante_accuracy_unit`, plus the same ones for elevation). I think storing this type of metadata is more efficient with the `DataProvenance` table, but will let @jirhiker weigh in, too.

jirhiker: It seems appropriate to store these fields here.  We can reevaluate later if user requirements dictate

PR #248, opened 2025-11-17
BDMS-227-231-aquifer-geology-models

ksmuczynski:
### **Why**

**Aquifer model:**
- Schema needs to be able to model the real-world, many-to-many relationship where a single Thing (well) can interact with multiple AquiferSystems.
- Schema should be able to distinguish between a well penetrating an aquifer and a well screen targeting a specific aquifer.
- Wells can encounter multiple aquifer characteristics simultaneously (e.g., both fractured and confined zones). The creation of a separate `AquiferType` model separates aquifer identity (name) from its characteristics (types), following proper database normalization.

**Geology model:**
- Adds support for tracking geologic formations and their associations with other entities (e.g., wells)
- Enables controlled vocabulary for formation codes and terms
- Improves data integrity and reference structure for geologic information

The purpose of this PR is to make these intermediate updates available to @jacob-a-brown as we continue to co-work on the main BDMS-227 task.

### **How**

**AQUIFER MODEL**
- Created new `AquiferSystem` model with controlled vocabulary fields and spatial boundary support
- Created new `AquiferType` model to store observed aquifer characteristics at specific well locations (e.g., Fractured, Confined, Perched)
- Created new ThingAquiferAssociation model to manage many-to-many relationships between Thing and AquiferSystem. This relationship focuses on a well's penetration of an aquifer.
- Updated Thing model to include aquifer_associations relationship and association proxy for direct access to related aquifer systems
- Updated the WellScreens model to include a new `aquifer_system_id` foreign key. This is important for identifying which aquifer a specific screen is monitoring

**GEOLOGY MODEL**
- Created `GeologicFormation` model for master reference of geologic formations
- Added `ThingFormationAssociation` model to link formations with things (e.g., wells)
- Updated `Thing` model to support formation associations
- Extended lexicon file to include new formation_code category and terms

### **Notes**
- This PR incorporates updates from BDMS-221: Core well information.
- Controlled vocabularies for lithology pending further review and cleanup
- Ensure migration scripts are applied to update database schema


core/lexicon.json
```diff
@@ -673,6 +665,34 @@
...
+    {"categories": ["aquifer_type"], "term": "Fractured", "definition": "Fractured"},
+    {"categories": ["aquifer_type"], "term": "Mixed", "definition": "Mix of confined and unconfined aquifers"},
+    {"categories": ["aquifer_type"], "term": "Perched", "definition": "Perched"},
+    {"categories": ["aquifer_type"], "term": "Semi-confined", "definition": "Semi-confined"},
+    {"categories": ["aquifer_type"], "term": "Unconfined, single", "definition": "Unconfined single aquifer"},
+    {"categories": ["aquifer_type"], "term": "Unconfined, multiple", "definition": "Unconfined multiple aquifers"},
+    {"categories": ["aquifer_type"], "term": "Unsaturated", "definition": "Unsaturated (dry)"},
+    {"categories": ["geographic_scale"], "term": "Major", "definition": "Major"},
+    {"categories": ["geographic_scale"], "term": "Intermediate", "definition": "Regional"},
+    {"categories": ["geographic_scale"], "term": "Local", "definition": "Local"}
```
jacob-a-brown: since `major` is used above should this be `minor`? or are those accepted/common terms?

db/aquifer_system.py
```diff
@@ -0,0 +1,77 @@
...
+    aquifer_type: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="A controlled vocabulary field to classify the aquifer's hydrologic properties (e.g., 'Unconfined', 'Confined', 'Perched').",
+    )
+    geographic_scale: Mapped[str] = lexicon_term(
+        nullable=False,
+        comment="A controlled vocabulary field to classify the aquifer's geographic scale (e.g., 'Major', 'Regional', 'Local').",
+    )
+    boundary: Mapped[Geometry] = mapped_column(
+        Geometry(geometry_type="MULTIPOLYGON", srid=4326, spatial_index=True),
```
jacob-a-brown: rather than hardcoding the srid you can import `SRID_WGS84` from constants and use that here. That'd help the code be more readable and ensure the same value is used throughout

db/aquifer_system.py
```diff
@@ -0,0 +1,77 @@
...
+        back_populates="aquifer_system",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
+
+    # --- Association Proxies ---
+    # Proxy to directly access Things (wells) associated with this AquiferSystem.
+    things: AssociationProxy[List["Thing"]] = association_proxy(
+        "thing_associations", "thing"
+    )
```
jacob-a-brown, line 77: should you add the same proxy for well screens?

db/thing_aquifer_association.py
```diff
@@ -0,0 +1,42 @@
...
+    # --- Relationship Definitions ---
+    # Many-To-One: This association links to one Thing.
+    thing: Mapped["Thing"] = relationship(
+        "Thing", back_populates="aquifer_associations"
+    )
+
+    # Many-To-One: This association links to one AquiferSystem.
+    aquifer_system: Mapped["AquiferSystem"] = relationship(
+        "AquiferSystem", back_populates="thing_associations"
+    )
```
jacob-a-brown, line 43: because these records need to be loaded for the `ThingResponse` should these be eagerly loaded?

db/aquifer_system.py
jirhiker: What feature requirement necessitates adding a proxy for well screens?

jacob-a-brown: there are no feature requirements, but since the description of the PR indicates this applies to well screens too i figured that may be a good feature to add (though not necessary right now)

core/lexicon.json
jacob-a-brown, line 1: the formation codes are out of date and need to be cleaned up. we may not want to transfer them until AMMP can get work on them

db/geologic_formation.py
```diff
@@ -0,0 +1,73 @@
...
+
+class GeologicFormation(Base, AutoBaseMixin, ReleaseMixin):
+    __versioned__ = {}
+
+    # TODO: Should `name` use a controlled vocabulary?
+    name: Mapped[str] = mapped_column(
+        nullable=False,
+        unique=True,
+        comment="The full, human-readable name of the geologic formation (e.g., 'Navajo Sandstone').",
+    )
```
jacob-a-brown: should this be a lexicon term?

db/geologic_formation.py
```diff
@@ -0,0 +1,73 @@
...
+        nullable=True,
+        comment="A detailed description of the geologic formation, its characteristics, and its significance.",
+    )
+    # TODO: Implement controlled vocabularies for `lithology` using NMAquifer's 'LU_Lithology' table.
+    #  This should be implemented after AMMP reviews and cleans up their formation terms and codes.
+    lithology: Mapped[str] = mapped_column(
+        nullable=True,
+        comment="A controlled vocabulary for the primary, dominant rock type"
+        "(e.g., 'Tuff', 'Sandstone', 'Alluvium', 'Shale').",
+    )
```
jacob-a-brown: lexicon term?

db/geologic_formation.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # --- Relationships ---
+    # One-To-Many (Association Object): A GeologicFormation can be associated with many Things (e.g., wells) via the
+    # ThingFormationAssociation join table.
+    thing_associations: Mapped[List["ThingFormationAssociation"]] = relationship(
+        "ThingFormationAssociation",
+        back_populates="geologic_formation",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
```
jacob-a-brown: because these records need to be loaded for the ThingResponse should these be eagerly loaded?

db/thing_geologic_formation_association.py
```diff
@@ -0,0 +1,60 @@
...
+    # --- Relationship Definitions ---
+    # Many-To-One: This association links to one Thing.
+    thing: Mapped["Thing"] = relationship(
+        "Thing", back_populates="formation_associations"
+    )
+
+    # Many-To-One: This association links to one GeologicFormation.
+    geologic_formation: Mapped["GeologicFormation"] = relationship(
+        "GeologicFormation", back_populates="thing_associations"
+    )
```
jacob-a-brown, line 60: because these records need to be loaded for the ThingResponse should these be eagerly loaded?

db/thing_formation_association.py
```diff
@@ -0,0 +1,60 @@
...
+from sqlalchemy.orm import relationship, Mapped, mapped_column
+
+from db.base import Base, AutoBaseMixin, ReleaseMixin
+
+if TYPE_CHECKING:
+    from db.thing import Thing
+    from db.geologic_formation import GeologicFormation
+
+
+class ThingFormationAssociation(Base, AutoBaseMixin, ReleaseMixin):
```
jacob-a-brown: should this be called `ThingGeologicFormationAssociation` to correspond with the `GeologicFormation` table?

db/thing_geologic_formation_association.py
jacob-a-brown, line 1: should the file be called `thing_geologic_formation_association.py` to be consistent with the `GeologicFormation` table?

core/lexicon.json
ksmuczynski: I changed "Intermediate" to "Regional" and added "Minor" as a separate term. I also clarified the definitions.

db/aquifer_system.py
ksmuczynski: I did the same thing for the 'geologic_formation` model, too.

db/geologic_formation.py
jirhiker: Does the database need to store the formation `name`? 

maybe its better it let the API map formation codes to names using a formations.json file that can be periodically updated from the authoritative source (.e.g USGS)

i think we remove name for now and a a placeholder formations.json file for the API to use

db/aquifer_system.py
ksmuczynski: No, the `WellScreen` model does not need proxies for `AquiferSystem` or `GeologicFormation`. The purpose of a proxy is to hide the intermediate join table. The relationship between `WellScreen`  and `AquiferSystem` and between `WellScreen`  and `GeologicFormation` is a direct many-to-one link via a foreign key. There is no intermediate table.

I did notice that `WellScreen` is missing a direct FK to the `GeologicFormation` table. I will update the `WellScreen` and `GeologicFormation` models to include this relationship.

db/geologic_formation.py
ksmuczynski: Should we do the same for `formation_code`, too?

core/lexicon.json
ksmuczynski: Should I remove the formation codes from the lexicon, or leave them but not transfer related legacy data?

db/thing_formation_association.py
ksmuczynski: I found `ThingGeologicFormationAssociation` was too lengthy, but I'll defer to you and @jirhiker. Is there a standard we should follow?

db/geologic_formation.py
ksmuczynski: I think eager loading only needs to occur on the relationships within the join table (`ThingFormationAssociation`), no?

db/thing_formation_association.py
jirhiker: `ThingGeologicFormationAssociation` for consistency

core/lexicon.json
jirhiker: no action needed. lets resolve this later

db/geologic_formation.py
jirhiker: leave `formation_code` in the model

db/geologic_formation.py
jacob-a-brown: Whoops! That's correct. The `thing` model should eagerly load the association records, which should eagerly load the geologic formation records.

db/thing.py
```diff
@@ -230,6 +258,30 @@ class Thing(Base, AutoBaseMixin, ReleaseMixin, StatusHistoryMixin, PermissionMix
...
+        passive_deletes=True,
+    )
+
+    # Many-To-Many: A Thing can penetrate many GeologicFormations.
+    formation_associations: Mapped[List["ThingFormationAssociation"]] = relationship(
+        "ThingFormationAssociation",
+        back_populates="thing",
+        cascade="all, delete-orphan",
+        passive_deletes=True,
+    )
```
jacob-a-brown, line 305: for quick and efficient access these should be eagerly loaded (otherwise we'll run into the N+1 problem)

schemas/aquifer_system.py
```diff
@@ -0,0 +1,32 @@
...
+    Geometry schema for GeoJSON response.
+    """
+
+    type: str
+    coordinates: (
+        List[float]
+        | List[List[float]]
+        | List[List[List[float]]]
+        | List[List[List[List[float]]]]
+    )
```
jacob-a-brown: GeoJSONGeometry is defined in schemas/location.py and can be imported from there. It requires three coordinates: longitude, latitude, and elevation.

schemas/geologic_formation.py
```diff
@@ -0,0 +1,46 @@
...
+
+class GeologicFormationResponse(BaseResponseModel):
+    """
+    Response schema for geologic formation details.
+    """
+
+    formation_code: str | None = None
+    description: str | None = None
+    lithology: str | None = None
+    boundary: GeoJSONGeometry | None = None
```
jacob-a-brown: the boundary is defined as a multypolygon in the database model. should this just be a string?

schemas/aquifer_system.py
```diff
@@ -0,0 +1,32 @@
...
+class AquiferSystemResponse(BaseResponseModel):
+    """
+    Response schema for aquifer system details.
+    """
+
+    name: str
+    description: str | None = None
+    aquifer_type: str
+    geographic_scale: str
+    boundary: GeoJSONGeometry | None = None
```
jacob-a-brown: the boundary is defined as a multypolygon in the database model. should this just be a string?

jacob-a-brown: and the function `point_to_wkt` in `schemas/location.py` can be renamed `geometry_to_wkt`, moved to `schemas/__init__.py`, and be used here for serialization

schemas/geologic_formation.py
jacob-a-brown: and the function `point_to_wkt` in `schemas/location.py` can be renamed `geometry_to_wkt`, moved to `schemas/__init__.py`, and be used here for serialization

schemas/thing.py
```diff
@@ -173,6 +195,23 @@ def populate_well_casing_materials_with_strings(cls, well_casing_materials):
...
+                return [assoc.aquifer_system for assoc in aquifers]
+        return aquifers or []
+
+    @field_validator("formations", mode="before")
+    def populate_formations(cls, formations):
+        """Convert formation association objects to response objects."""
+        if formations is not None:
+            # formations should already be ThingFormationAssociation objects
+            return formations
+        return []
```
jacob-a-brown: I don't think these field validators are necessary since the `thing` model has `aquifers` and `formations` association proxies; the pydantic schemas will get populated from those.

jacob-a-brown (commented):
The updates from BDMS-221 have been incorporated into BDMS-227. If you merge BDMS-227 with this branch (and push back to this PR) the length of the PR will become much shorter (and it'll have the most current work).

db/thing.py
ksmuczynski: Aren't they being eagerly loaded within the `ThingAquiferAssociation` and `ThingGeologicFormationAssociation` models?

schemas/aquifer_system.py
ksmuczynski: I think that's only for point geometries. The aquifer and formation boundaries/extents are MULTIPOLYGON, which require this structure. I am open to alternative implementation strategies, but I think this nested structure is necessary for MULTIPOLYGON.

db/thing.py
jacob-a-brown: My understanding is that the eager loading needs to be done from the model where you want to obtain the related data. By defining the eager load in `ThingGeologicFormationAssocation` if you call `Thing.formation_associations` it will need to make another query to the database (N+1). If the eager load of the association model is in the `Thing` model, however, the data will be loaded in one query and calling `Thing.formation_associations` won't make further queries to the database.

With the eager loading in `ThingGeologicFormationAssociation` you can call `ThingGeologicFormationAssociation.thing` and the related `thing` model will already be loaded from the original SQL query, but since we want to get the associated data from the `Thing` model I think the eager loading should be done there (that is, have the `thing` eagerly load the association, and have the association eagerly load the formation)

schemas/geologic_formation.py
ksmuczynski: I thought it would be easier to render in maps as GeoJSON. Is a string better? @jirhiker thoughts?

db/thing.py
jacob-a-brown: From what I've read eager loading looks to be unidirectional, meaning that if it is defined on `ThingGeologicFormationAssocation` it is not also defined on `Thing`. Since we want to get the associated geologic formations for a thing I think they need to be defined 

`thing` --> `association` --> `formations` (and `aquifers`)

https://chatgpt.com/share/e/691e3831-ffe8-800a-9be5-3943884705a0
https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#configuring-loader-strategies-at-mapping-time

I'll look for some more documents about loading strategies and post them here if relevant

schemas/geologic_formation.py
jacob-a-brown: Because they are geographic, aquifer and geologic responses are now GeoJSONs themselves

schemas/aquifer_system.py
jacob-a-brown: We decided to return aquifer/geology as a full GeoJSON response because they are geographically located

schemas/aquifer_system.py
jacob-a-brown: We decided to return aquifer/geology as a full GeoJSON response because they are geographically located

jacob-a-brown (approved):

PR #249, opened 2025-11-17
[NO TICKET] Add db healthcheck to docker-compose file

TylerAdamMartinez:
### **Why**
- The FastAPI backend container frequently failed to start in GitHub Actions due to PostGIS taking longer to initialize than standard PostgreSQL.
- The application attempted to connect to the database before PostGIS finished booting, causing repeated connection failures (db:5432 - no response / Waiting for postgres.).
- The test workflow’s health check waited on FastAPI, not the underlying database, so even increasing timeouts did not fix the issue.

### **How**
- Added a proper healthcheck to the db service in docker-compose.yml to wait for PostGIS to become ready (pg_isready).
- Updated app service to use depends_on: condition: service_healthy so FastAPI starts only after the database is fully initialized.

### **Notes**
- Use bullet points here

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,11 @@ services:
       - 5432:5432
     volumes:
       - postgres_data:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER}"]
+      interval: 2s
+      timeout: 5s
+      retries: 20
 
   app:
     build: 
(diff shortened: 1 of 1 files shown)


jacob-a-brown:
The health check is present in `entrypoint.sh`. should it go there or here?

jacob-a-brown:
The only reason I really made `entrypoint.sh` was to do this health check. It makes sense to me to have it in the docker compose files for ease of understanding. In that case, we may want to add a step to populate the database (`alembic upgrade head`) and start the application (`uvicorn main:app --host 0.0.0.0 --port 8000 --reload`)

jirhiker:
I like having the health check in the docker-compose file. It seems pretty standard practice

jirhiker (approved):

jacob-a-brown:
If it's in docker-compose should we remove entrypoint? to keep a clean and consistent implementation

PR #250, opened 2025-11-17
source tests for bdms 221

jacob-a-brown:
implementation of the `DataProvenance` model to pass tests. This meant removing many fields from `Location` that are now found in `DataProvenance`. Since there are no feature files for these fields yet they are not being handled/added back


db/location.py
```diff
@@ -83,6 +79,17 @@ def latlon(self):
...
+    @property
+    def elevation_method(self) -> str | None:
+        data_provenance_records = self.data_provenance
+        elevation_method_record = [
+            r for r in data_provenance_records if r.field_name == "elevation"
+        ]
+        if elevation_method_record:
+            return elevation_method_record[0].collection_method
+        else:
+            return None
```
jirhiker: this should be a function provided by the DataProvenanceMixin

```python
def _get_data_provenance_attribute(self, attribute, record_name):
    data_provenance_records = self.data_provenance
    record = next(
            r for r in data_provenance_records if r.field_name == attribute
        ), None)
    if record:
        return getattr(record[0], record_name)
```

db/thing.py
```diff
@@ -360,6 +360,17 @@ def measuring_point_description(self) -> str | None:
...
+    @property
+    def well_depth_source(self) -> str | None:
+        data_provenance_records = self.data_provenance
+        well_depth_source_records = [
+            r for r in data_provenance_records if r.field_name == "well_depth"
+        ]
+        if well_depth_source_records:
+            return well_depth_source_records[0].origin_source
+        else:
+            return None
```
jirhiker: see above comment

jirhiker (approved):

ksmuczynski (approved):
Looks good!

PR #253, opened 2025-11-18
BDMS 221: well core information

jacob-a-brown:
### **Why**
- The feature file `well-core-information.feature` was written, requiring updates throughout the repository

### **How**
- Updated existing models
- Added new models
- Updated schemas
- Updated transfer scripts
- Added step tests
- Updated step tests
- Updated some pytest tests
- Skipped some pytest tests

### **Notes**
- Some pytest tests were easy to fix, but others required much more exhaustive changes to the API. Because of this, and because of the focus on the well details page, the more difficult ones have been skipped.
- `WellData.MonitoringStatus` is currently being used to transfer both monitoring status and frequency from **NM_Aquifer**, but this field has not been maintained. Once we receive a table Ethan made in May the transfer script will need to be updated again.
- Much of the work herein has been reviewed in smaller PRs, but it would be good to get eyes on this work again before it gets merged with staging.
- **IMPORTANT**: now that the `current_location` for a `thing` is in the form of a GeoJSON, should _all_ location responses be in the same GeoJSON format? It may be a good idea to have this be the standard, rather than have a `LocationResponse` that is different from the `LocationGeoJSONResponse`. (For example: the `current_location` in the `ThingResponse` uses a GeoJSON, whereas the response for `/location` uses the `LocationResponse`.


schemas/location.py
```diff
@@ -50,10 +53,10 @@ class CreateLocation(BaseCreateModel, ValidateLocation):
     notes: List[CreateNote] = []
     point: str  # point is required and should be in WKT format
     elevation: float
-    elevation_accuracy: float | None = None
-    elevation_method: ElevationMethod | None = None
-    coordinate_accuracy: float | None = None
-    coordinate_method: CoordinateMethod | None = None
+    # elevation_accuracy: float | None = None
```
TylerAdamMartinez, line 56: Why not delete these fields if they're not going to be used?

jacob-a-brown: I decided to comment them out for now because we may want to include them in the future, so they're here for notes. I can remove those comments if it's bad style to keep these in.

jirhiker (approved):

PR #255, opened 2025-11-19
Bdms 227 jab updates to pass tests

jacob-a-brown:
This PR implements model/schema updates to pass all well additional information tests except for the aquifer/geology tests. Those will get implemented after the models have been finalized.

@chasetmartin:

- the feature step `the response should include the source of the completion information` seems to apply to the completion date of the well, is that correct?
- the feature step `the response should include the source of the construction information` seems to apply to the construction method of the well, is that correct?
- should other fields, like the driller name, also have a source?


jacob-a-brown:
the comparison to bdms 227 is broken and outdated. i'm going to close this and open a new one and that will hopefully fix this bug

PR #257, opened 2025-11-25
Bdms 227 transfer updates

jacob-a-brown:
all transfers for additional well information except for aquifer/geology data. this depends on PR #256 being merged since it incorporates api/model/schema changes made for the tests to pass


db/thing.py
```diff
@@ -484,6 +545,10 @@ class WellScreen(Base, AutoBaseMixin, ReleaseMixin):
         "AquiferSystem", back_populates="well_screens", passive_deletes=True
     )
 
+    geologic_formation: Mapped["GeologicFormation"] = relationship(
+        "GeologicFormation", back_populates="well_screens", passive_deletes=True
+    )
+
```
ksmuczynski, line 533: Do these relationships belong in the `ThingIdLink` model?

schemas/thing.py
```diff
@@ -24,18 +24,20 @@
...
-from schemas.aquifer_system import AquiferSystemResponse
-from schemas.geologic_formation import (
-    GeologicFormationResponse,
-    ThingGeologicFormationAssociationResponse,
-)
+from schemas.aquifer_system import AquiferSystemGeoJSONResponse
+
+# from schemas.geologic_formation import (
+#     GeologicFormationResponse,
+# )
```
ksmuczynski: Don't we still need a GeologicFormationResponse?

schemas/thing.py
```diff
@@ -24,18 +24,20 @@
...
 from schemas import BaseCreateModel, BaseUpdateModel, BaseResponseModel, PastOrTodayDate
 from schemas.group import GroupResponse
 from schemas.location import LocationGeoJSONResponse
 from schemas.notes import NoteResponse, CreateNote
-from schemas.aquifer_system import AquiferSystemResponse
-from schemas.geologic_formation import (
-    GeologicFormationResponse,
-    ThingGeologicFormationAssociationResponse,
-)
+from schemas.aquifer_system import AquiferSystemGeoJSONResponse
```
ksmuczynski: I thought we decided to forego the GeoJSON response?

schemas/thing.py
```diff
@@ -229,18 +238,23 @@ class WellResponse(BaseThingResponse):
...
+    well_construction_method_source: str | None
     well_pump_type: WellPumpType | None
     well_pump_depth: float | None
     well_pump_depth_unit: str = "ft"
-    aquifers: list[AquiferSystemResponse] = []
-    formations: list[ThingGeologicFormationAssociationResponse] = []
+    allow_water_level_samples: bool | None
+    allow_water_chemistry_samples: bool | None
+    allow_datalogger_installation: bool | None
+    is_suitable_for_datalogger: bool | None
```
ksmuczynski: These look like fields from the previous PermissionHistory table design. Should these fields be removed and replaced with `permission_type` and/or `permission_allowed`?

ksmuczynski (commented):
Overall looks good! Left a few comments.

db/thing.py
jacob-a-brown: I don't think that they would. These two are in the `WellScreen` model

schemas/thing.py
jacob-a-brown: And I'll just include the `name` of hte aquifer system in the `WellScreenResponse`. It's not currently used, but that shouldn't be an issue for the time being

db/thing.py
ksmuczynski: Oh, I thought I saw them in the ThingIdLink model, which is why I was confused. All good!

schemas/thing.py
jacob-a-brown: Sorry if you are getting this twice. I somehow started a review of my own PR, which I just deleted so now I need to comment again:

The steps in the feature file are

```
And the response should include whether repeat measurement permission is granted for the well
And the response should include whether sampling permission is granted for the well
And the response should include whether datalogger installation permission is granted for the well
```

so I'm just returning boolean values based off of the latest response where `end_date` is `None`. This data is, however, stored in the `PermisionHistory` table.

@chasetmartin do you want the response to be single fields like this, or would you prefer that it more closely reflect the database design? If so I can change it to

```python
{
...
"permissions": [
  {"permission_type": "Water Level Sample", "permission_allowed": True/False},
  {"permission_type": "Water Chemistry Sample", "permission_allowed": True/False},
  {"permission_type": "Datalogger Installation", "permission_allowed": True/False}
]
}

File: schemas/thing.py (hunk 3 above)
Comment (jacob-a-brown) [reply schemas/thing.py]: This was used in the `WellResponse` and I forgot it was there. I've now changed it so that a string of the aquifer system name is returned

File: schemas/thing.py (hunk 2 above)
Comment (jacob-a-brown) [reply schemas/thing.py]: I don't think we do. The feature file states that it should just return the formation codes, so I parse all of the geologic formation and just return the codes as plain strings

File: schemas/thing.py (hunk 4 above)
Comment (chasetmartin) [reply schemas/thing.py]: @jacob-a-brown I like the later, the list of permission objects, I think it's almost more intuitive to read/parse. What do you think @TylerAdamMartinez ?

Comment (TylerAdamMartinez) [reply schemas/thing.py]: Same here, let's make it more closely resemble the database design.

Comment (jacob-a-brown) [reply schemas/thing.py]: Do you want it to return _all_ permissions, even if they have ended? Or just the latest permissions (that is, latest `start_date` where `end_date` is `None`)?

**Review by ksmuczynski — APPROVED** [review]
Nice work!

File: schemas/thing.py (hunk 4 above)
Comment (jacob-a-brown) [reply schemas/thing.py]: the permissions are now returned in a list

```python
{
...
"permissions": [
  {"permission_type": "Water Level Sample", "permission_allowed": True/False, "start_date": date, "end_date": None},
  {"permission_type": "Water Chemistry Sample", "permission_allowed": True/False, "start_date": date, "end_date": None},
  {"permission_type": "Datalogger Installation", "permission_allowed": True/False, "start_date": date, "end_date": None}
]
}
```

- Only the latest record (based off of `start_date`) where `end_date` is `None` is returned for each type
- If the permission record does not exist for a particular `permission_type`then `permission_allowed`, `start_date`, and `end_date` are `None`.
- while `start_date` is non-nullable in the database, it must be nullable in the response incase there are no permissions of  particular type
- `end_date` will always be `None` for active permissions. I'm including it in the `PermissionHistory` response, though, so that if we want to list out the full permission history for an object it'll be there

PR #259, opened 2025-11-25
BDMS 296: increase limit size for search

jacob-a-brown:
### **Why**
- A pagination size of 25 is too restrictive

### **How**
- Increased the size to 100

### **Notes**
- I haven't tested a pagination in the search tests, but from the research I've done setting a query parameter called `size` should do the trick.

### DIFF
--- a/api/search.py
+++ b/api/search.py
@@ -175,7 +175,8 @@ async def search_api(
     user: viewer_dependency,
     session: session_dependency,
     q: str,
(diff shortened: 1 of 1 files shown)


TylerAdamMartinez (approved):
lgtm

chasetmartin (approved):

ksmuczynski (approved):

PR #261, opened 2025-11-25
BDMS 233: transfer updates

jacob-a-brown:
### **Why**
- notes need to be transferred to the polymorphic notes table

### **How**
- updated well and location transfer scripts/code to transfer the following note types:
  - well: `Other`, `Water`, `Construction`, `Measuring`
  - location: `Other`, `Coordinate`
- refactored legacy notes fields to go into the into the polymorphic notes table:
  - `thing.well_construction_notes` --> `note_type = "Construction"`
  - `location.nma_notes_location` -- > `note_type = "Other"`
  - `location.nma_coordinate_notes` --> `note_type = "Coordinate"`
- update `well-notes.py` feature test for the refactoring of `well_construction_notes` to the polymorphic notes table
  

### **Notes**
- `Measuring` notes for a `thing` come from `WaterLevels.SiteNotes` in **NM_Aquifer**. The documentation for that field states "Optional notes concerning the nature of the well site or measuring event."
- This PR utilizes work done for **well-additional-information** so it should not be merged until PR #257  and PR #258 are merged into staging. I am opening a PR, though, to preserve the work done and get feedback when able. That being said, once those two PRs are merged into staging the number of files (and lines) changed will decrease dramatically.
  - I'll re-request reviews once those two PRs are merged.


tests/features/steps/well-additional-information.py
```diff
@@ -0,0 +1,210 @@
...
+
+@then(
+    "the response should include the aquifer class code to classify the aquifer into aquifer system."
+)
+def step_impl(context):
+    for aquifer in context.water_well_data["aquifers"]:
+        assert "aquifer_system" in aquifer
+    assert sorted(
+        [a.get("aquifer_system") for a in context.water_well_data["aquifers"]]
+    ) == sorted([system.name for system in context.objects["aquifer_systems"]])
```
jirhiker: comparing sets might be a little more elegant

`{a.get("aquifer_system") for a in ... } == {system.name for system in ... }`

transfers/permissions_transfer.py
```diff
@@ -0,0 +1,87 @@
...
+            )
+            continue
+        else:
+            # Assuming the first contact is the relevant one
+            contact_id = well.contacts[0].id
+
+        allow_water_level_samples = wdf.loc[
+            wdf["PointID"] == well.name, "MonitorOK"
+        ].values
+        if len(allow_water_level_samples) > 0 and allow_water_level_samples is not None:
```
jirhiker: you want to test if `allow_water_level_samples` is not None before testing its length.  this will fail if `allow_water_level_samples` is None because you cannot `len(None)`

transfers/permissions_transfer.py
```diff
@@ -0,0 +1,87 @@
...
+
+def transfer_permissions(session: Session):
+    """
+    The transferred wells and contacts need to be queried to know who gave
+    permission to which well since contact_id is required for PermissionHistory
+    """
+    wdf = read_csv("WellData", dtype={"OSEWelltagID": str})
+    wdf = replace_nans(wdf)
+
+    transferred_wells = (
```
jirhiker, line 27: all the other transfers use `filter_to_valid_point_ids` for this functionality

transfers/thing_transfer.py
```diff
@@ -57,12 +57,17 @@ def transfer_thing(session: Session, site_type: str, make_payload, limit=None) -
...
+                row, cached_elevations
+            )
             session.add(location)
             session.flush()
             data_provenances = make_location_data_provenance(
                 row, location, elevation_method
             )
+            for note_type, note_content in location_notes.items():
+                if not isna(note_content):
+                    location.add_note(note_content, note_type)
```
jirhiker, line 70: Doesn't add_note just create a note object? you still need to add it to the session

transfers/waterlevels_transfer.py
```diff
@@ -385,6 +385,19 @@ def transfer_water_levels(session):
             logger.info(
                 f"{SPACE_4}Created observation: ID {observation.id} | DT {observation.observation_datetime} | Value {observation.value} | MPHeight {observation.measuring_point_height} | nma_pk_waterlevels {observation.nma_pk_waterlevels}"
             )
+
+            # WaterLevels.SiteNotes --> notes where note_type = "measuring_notes"
+            if not pd.isna(row.SiteNotes):
+                note = thing.add_note(
```
jirhiker: how is this note going to be associated with this water level measurement?

transfers/well_transfer.py
```diff
@@ -117,6 +118,21 @@ def _extract_casing_materials(row) -> list[str]:
     return materials
 
 
+def _extract_well_pump_type(row) -> str | None:
+    construction_notes = row.ConstructionNotes.lower()
+    if "pump" in construction_notes:
```
jirhiker: a regex would be better here. for the most part this will work but what about the note.   "Handle on steel tape broke. Needed to jettison our tools. Pump is a submersible".  The function would incorrectly identify the pump as "Jet". 

```
import re


pattern = re.compile(
    r"\b(?P<term>jet|hand|submersible)\b|\b(?P<phrase>line[-\s]+shaft)\b",
    re.IGNORECASE
)

def first_matched_term(text: str):
    m = pattern.search(text)
    if not m:
        return None
    return m.group("term") or m.group("phrase")

PUMP_MAPPING = {
    "jet": "Jet",
    "hand": "Hand",
    "submersible": "Submersible"
}

def _extract_well_pump_type(row) -> str | None:
    construction_notes = row.ConstructionNotes.lower()
    return PUMP_MAPPING.get(first_matched_term(construction_notes), None)
```

transfers/well_transfer.py
```diff
@@ -236,6 +255,9 @@ def transfer_wells(session: Session, flags: dict = None, limit: int = 0) -> None
             well_casing_materials = (
                 [] if isna(row.CasingDescription) else _extract_casing_materials(row)
             )
+            well_pump_type = (
+                _extract_well_pump_type(row) if row.ConstructionNotes else None
```
jirhiker: move the `if row.ConstructionNotes` test to `_extract_well_pump_type`

.pre-commit-config.yaml
```diff
@@ -16,15 +16,15 @@ repos:
...
-  - repo: local
-    hooks:
-      - id: pytest
-        name: pytest
-        entry: pytest  # Or your specific test command, e.g., poetry run pytest
-        language: system
-        types: [python] # Specify relevant file types for your tests
-        pass_filenames: false
-        always_run: true
+#  - repo: local
```
jirhiker, line 19: You should not be committing these temporary, dev changes to .pre-commit-config.yaml

jirhiker (commented):
Looks good. I will need to refactor slightly to match the latest transfer branch

tests/features/steps/well-additional-information.py
jacob-a-brown: This will get changed in PR #256 and then will be present in this PR once that one is merged.

transfers/permissions_transfer.py
jacob-a-brown: This will get changed in PR #257 and then will be present in this PR once that one is merged.

transfers/thing_transfer.py
jacob-a-brown: I just looked at it and it does - I'll fix to add it to the session

transfers/waterlevels_transfer.py
jacob-a-brown: It won't be associated with water level measurements in Ocotillo, but with wells. The step from `well-notes.feature` indicates `the response should include measuring notes (notes about measuring/visiting the well, on Access form)`. This is really the only field in the database that explicitly records notes about measures/visits, so it is now being associated with a well and not an observation.

The documentation is: "Optional notes concerning the nature of the well site or measuring event."

transfers/well_transfer.py
jacob-a-brown: This will get changed in PR #257 and then will be present in this PR once that one is merged.

transfers/waterlevels_transfer.py
chasetmartin: I have messages out to Ethan and Sianin to clarify "measuring notes" since this field they mention that they want to see and enter data into is not found specifically in NM Aquifer and needs clarification. Looking at some of the values in WaterLevels.SiteNotes, they are definitely observation level, containing info like battery levels, which seems most relevant when associated with a specific time/date of observation.
image
I assume this response field from the old AMP API is the WaterLevels.SiteNotes?

jacob-a-brown: That's correct - that response comes from `WaterLevels.SiteNotes`

jirhiker: as @chasetmartin mentioned WaterLevel.SiteNotes is related to observation level notes where time/date info is necessary. 

I think the feature is saying return all the observation notes for this well, i.e. all the WaterLevel.SiteNotes.  So WaterLevel.SiteNotes needs to be associated with a measurement. The API then collects all these notes and returns them with the other well metadata

jacob-a-brown: Got it. I'll then update this transfer, as well as update the feature file/test to reflect this

transfers/permissions_transfer.py
jacob-a-brown: The permissions require the `id` of the `thing` record to set the `target_id` field, which is why the wells are queried in `permissions_transfer.py` (after the wells have been transferred and assigned `id`s by the database). Using `filter_to_valid_point_ids` would get me the correct `thing.name` for all of the wells, but wouldn't fetch the `id`s.

jacob-a-brown: When I was figuring out this transfer it made me sense to me to do it this way, but I can move this to `well_transfer.py` after line 365 - the second after the comment "add things that need  well id"

jacob-a-brown: Upon reflection I think it makes more sense to go there. It'll be better organized and easier to follow

jacob-a-brown: I forgot another reason why `permission_transfer.py` is its own file... The permissions require an associated `contact`, so they need to be transferred after both the `things` and `contacts` are transferred (which is why they can't go into `well_transfer.py` (which negates the above two comments)

transfers/permissions_transfer.py
jacob-a-brown: See fix in PR #263

transfers/well_transfer.py
jacob-a-brown: See fix in PR #263

jacob-a-brown:
This PR is out of date. See PR #272

PR #262, opened 2025-11-26
Transfer refactor for future reusability improvements

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


transfers/transferer.py
```diff
@@ -0,0 +1,210 @@
...
+                "table": table,
+                "field": field,
+            }
+        )
+
+    def _transfer_hook(self, session: Session):
+        self._limit_iterator(session, self.flags.get("LIMIT", 0))
+
+    def _get_df_to_iterate(self) -> pd.DataFrame:
+        return self.cleaned_df
```
jacob-a-brown, line 67: the method `_get_df_to_iterate` adds an extra layer of complication. If all it does is `return self.cleaned_df` I think that you can get rid of the method and set `df = self.cleaned_df` where you call `self._get_df_to_iterate`. That'll make the code/process easier to understand and follow.

transfers/transferer.py
jacob-a-brown, line 1: There are a number of methods here (mainly in `Transferer` and `ChunkTransferer`) that need to be implemented for each transfer. Because they'll need to be developed multiple times it'd be helpful to have docstrings for each method describing what they do and (potentially) return.

jacob-a-brown (commented):
calling return rather than continue when we hit an error stops the process from running for subsequent records. Do we want that? Or should continue still be used so that all records can be evaluated (and errors identified for subsequent processing of data)

jacob-a-brown:
It will take me a few iterations of review to better understand this new implementation, but I just left some initial comments. I'll review again later today to get a better feel for this implementation and leave more comments if needed. Overall I like the class-based approach. I agree that it'll make the code easier to maintain in the future - it adds a lot of flexibility. I noted this in a comment, but because there are a number of methods implemented throughout (essentially ones that need to be rewritten for different transfer scripts), it would be super helpful to have docstrings describing what each method needs, does, and returns.

jirhiker:
I will work on some clean up this afternoon; adding doc strings and removing any unnecessary complexity

jirhiker:
> calling return rather than continue when we hit an error stops the process from running for subsequent records. Do we want that? Or should continue still be used so that all records can be evaluated (and errors identified for subsequent processing of data)

I don't believe the code is every using return within a loop. Return is used within loop step functions to exit the current function and return to the outer loop

jirhiker:
> It will take me a few iterations of review to better understand this new implementation, but I just left some initial comments. I'll review again later today to get a better feel for this implementation and leave more comments if needed. Overall I like the class-based approach. I agree that it'll make the code easier to maintain in the future - it adds a lot of flexibility. I noted this in a comment, but because there are a number of methods implemented throughout (essentially ones that need to be rewritten for different transfer scripts), it would be super helpful to have docstrings describing what each method needs, does, and returns.

It follows a similar pattern to DIE; classes, subclasses, and hooks

transfers/waterlevels_transducer_transfer.py
```diff
@@ -13,149 +13,196 @@
...
+    _partition_field = "PublicRelease"
+    _sensor_type = "Acoustic Sounder"
+
+
+def _find_deployment(ts, deployments):
+    for d in deployments:
+        start = Timestamp(d.installation_date)
+        if start > ts:
+            break  # because sorted by start
+        end = Timestamp(d.removal_date) if d.removal_date else Timestamp.max
```
jacob-a-brown: When `removal_date` gets converted to `pandas.Timestamp` it sets the time to midnight (e.g. `2018-01-01` --> `2018-01-01T00:00:00`). During comparison if there are measurements on that date, like `2018-01-01T12:00:00`, it will evaluate that the measurement occurs after removal and therefore won't be transferred. Is this the expected behavior? Or, should we compare the timestamp of the measurement to the day after removal?

jirhiker:
@codex review

transfers/waterlevels_transducer_transfer.py
```diff
@@ -13,149 +13,196 @@
...
+
+        except ValidationError as e:
+            logger.critical(f"Observation validation error: {e.errors()}")
+            self._capture_error(pointid, str(e), "DepthToWaterBGS")
+
+
+class WaterLevelsContinuousPressureTransferer(WaterLevelsContinuousTransferer):
+    source_table = "WaterLevelsContinuous_Pressure"
+    _partition_field = "QCed"
+    _sensor_types = ("Pressure Transducer", "Barometer", "DiverLink", "Diver Cable")
```
jacob-a-brown, line 188: `DiverLink` and `Diver Cable` sensors are not currently transferred. To get them transferred, and then have their observations transferred too, they should be added to the dictionary `EQUIPMENT_TO_SENSOR_TYPE_MAP` in **transfers/sensor_transfer.py**

transfers/util.py
```diff
@@ -53,35 +54,190 @@
...
+            mphs = []
+            start_dates = []
+            mph_descs = []
+
+            if len(df) == 0:
+                logger.warning(f"No measurements found for PointID: {row.PointID}.")
+            else:
+                # try to estimate mpheight from measurements
+                for m in df.itertuples():
+                    mphi = m.DepthToWater - m.DepthToWaterBGS
```
jacob-a-brown, line 83: This will run into an error if the row doesn't have either `DepthToWater` or `DepthToWaterBGS`. Can you put an if statement here? And if neither are found (and there's no mp height) should you raise an error?

transfers/well_transfer.py
```diff
@@ -320,165 +425,374 @@ def transfer_wells(session: Session, flags: dict = None, limit: int = 0) -> None
...
+                for code in NMA_MONITORING_FREQUENCY.keys():
+                    if code in row.MonitoringStatus:
+                        monitoring_frequency = NMA_MONITORING_FREQUENCY[code]
+                        monitoring_frequency_history = MonitoringFrequencyHistory(
+                            thing_id=well.id,
+                            monitoring_frequency=monitoring_frequency,
+                            start_date=datetime.now(tz=UTC),
+                            end_date=None,
+                        )
+                        session.add(monitoring_frequency_history)
```
jacob-a-brown: these are the monitoring frequencies that Ethan mentioned are likely bad data/out of date. Do we want to keep them? And just have folks update them as they access/work on wells

jirhiker: Keep them and let amp fix as they see fit.  This data is "bad" not invalid.

core/lexicon.json
jacob-a-brown, line 1: The new organizations will need to be added to `transfers/data/owners_organization_mapper.json` for the contact transfer to work

transfers/util.py
jirhiker: I have not experienced that issue. This is technical debt that may or may not need to be addressed

core/lexicon.json
```diff
@@ -367,6 +373,82 @@
...
+    {"categories": ["organization"], "term": "El Paso Water", "definition": "El Paso Water"},
+    {"categories": ["organization"], "term": "BLM, Socorro Field Office", "definition": "BLM, Socorro Field Office"},
+    {"categories": ["organization"], "term": "USFWS", "definition": "US Fish & Wildlife Service"},
+    {"categories": ["organization"], "term": "NPS", "definition": "National Park Service"},
+    {"categories": ["organization"], "term": "Sile MDWCA", "definition": "Sile Municipal Domestic Water Assn."},
+    {"categories": ["organization"], "term": "Pena Blanca Water & Sanitation District", "definition": "Pena Blanca Water & Sanitation District"},
+    {"categories": ["organization"], "term": "Town of Questa", "definition": "Town of Questa"},
+    {"categories": ["organization"], "term": "Lamy MDWCA", "definition": "Lama MDWCA"},
+    {"categories": ["organization"], "term": "Town of Cerro", "definition": "Town of Cerro"},
+    {"categories": ["organization"], "term": "Farr Cattle Company", "definition": "Farr Cattle Company (Farr Ranch)"},
```
jacob-a-brown, line 398: Farr Cattle Company already exists in the lexicon

core/lexicon.json
jirhiker: I dont think that is true

core/lexicon.json
```diff
@@ -367,6 +373,82 @@
...
+    {"categories": ["organization"], "term": "Riverbend Hotsprings", "definition": "Riverbend Hotsprings"},
+    {"categories": ["organization"], "term": "Armendaris Ranch", "definition": "Armendaris Ranch"},
+    {"categories": ["organization"], "term": "El Paso Water", "definition": "El Paso Water"},
+    {"categories": ["organization"], "term": "BLM, Socorro Field Office", "definition": "BLM, Socorro Field Office"},
+    {"categories": ["organization"], "term": "USFWS", "definition": "US Fish & Wildlife Service"},
+    {"categories": ["organization"], "term": "NPS", "definition": "National Park Service"},
+    {"categories": ["organization"], "term": "Sile MDWCA", "definition": "Sile Municipal Domestic Water Assn."},
+    {"categories": ["organization"], "term": "Pena Blanca Water & Sanitation District", "definition": "Pena Blanca Water & Sanitation District"},
+    {"categories": ["organization"], "term": "Town of Questa", "definition": "Town of Questa"},
+    {"categories": ["organization"], "term": "Lamy MDWCA", "definition": "Lama MDWCA"},
```
jacob-a-brown, line 396: Lamy MDWCA is in the lexicon twice. I forgot to post this in the previous review...

jacob-a-brown: Valle Vista Water Utility is also in there three different ways

jacob-a-brown: And Santa Fe Downs is in there twice

jirhiker: "Valle Vista Water Utility is also in there three different ways" amp will need to resolve this

core/lexicon.json
jacob-a-brown: I didn't notice that `org_mapper.json` was added 👍. It looks to do the same thing and should work out. Should we log a similar warning or message for when `owners_organization_mapper.json` is used?

jirhiker: technical debt that can be addressed at any point

jacob-a-brown (approved):
I think it looks good! And I like the movement to class-based transfers. With a large PR it can be difficult to be sure I gave everything its due in a review, but if we run into something in the future we can always open a fix PR

transfers/transferer.py
jirhiker: its here for future flexibility.

PR #263, opened 2025-11-26
BDMS 227: permission & well pump type transfer fix

jacob-a-brown:
### **Why**
- this addresses comments about well pump type transfers made in PR #261
- this addresses comments about permissions transfer made in PR #261 

### **How**
- use regex to get the well pump type from the construction notes
- ensure that the permissions exist and are not null before converting 0/1 to boolean values

### **Notes**


ksmuczynski (approved):
This looks good.

PR #266, opened 2025-12-01
BDMS-227-254: add schema validations to aquifer and geology schemas

ksmuczynski:
### **Why**
- Data Integrity: Currently, there are no checks to ensure that depth intervals are logical (i.e., bottom_depth > top_depth) or that spatial boundaries are valid WKT strings.
- Missing Functionality: There was no Pydantic schema available to handle the creation of ThingGeologicFormationAssociation records.
- Loose Typing: Input strings for fields like primary_aquifer_type or lithology were not strictly validated against their respective controlled vocabularies, leading to potential data inconsistencies (e.g., casing mismatches).

### **How**

**New Validators Module**: Created schemas/validators.py containing reusable mixins:
    - DepthIntervalMixin: Enforces that bottom_depth > top_depth and depths are non-negative.
    - GeometryMixin: Validates boundary fields
    - validate_enum_input: A helper function to validate inputs against Enums case-insensitively.

**Geologic Formations:**
- Added CreateThingGeologicFormationAssociation schema to schemas_geologic_formation.py using the DepthIntervalMixin.
- Updated CreateGeologicFormation to enforce uppercasing on formation_code and Enum validation on lithology.

**Aquifer Systems:**
- Updated CreateAquiferSystem in schemas_aquifer_system.py to enforce AquiferType and GeographicScale Enum validation.
- Added GeometryMixin to CreateAquiferSystem for WKT boundary validation.

### **Notes**
- **Design decisions:** Shared validation logic was placed in a dedicated schemas/validators.py file rather than __init__.py to maintain clean separation of concerns and prevent circular import issues.


schemas/validators.py
```diff
@@ -0,0 +1,76 @@
...
+    # 2. Case-insensitive fallback (for string inputs)
+    if isinstance(v, str):
+        v_lower = v.lower()
+        for member in enum_cls:
+            if str(member.value).lower() == v_lower:
+                return member.value
+
+    # 3. Fail if no match found
+    valid_options = [str(e.value) for e in enum_cls]
+    raise ValueError(f"Invalid value '{v}'. Must be one of: {', '.join(valid_options)}")
```
jacob-a-brown: By setting a field to an enum Pydantic will automatically check if that field is one of the values in the enum's items. Because of this I don't think the function `validate_enum_input` is needed; let Pydantic check the input and raise errors if necessary. I also think that inputs should be strict for case-sensitivity as we are targeting specific strings.

schemas/validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class DepthIntervalMixin(BaseModel):
```
jacob-a-brown, line 11: This would be helpful for well screens, too, since depths are also checked there. Should we change the names of the well screen fields to match here? and reuse the validator

schemas/validators.py
```diff
@@ -0,0 +1,76 @@
...
+    @model_validator(mode="after")
+    def check_depth_logical_order(self) -> "DepthIntervalMixin":
+        if self.bottom_depth is not None and self.top_depth is not None:
+            if self.bottom_depth <= self.top_depth:
+                raise ValueError(
+                    f"Bottom depth ({self.bottom_depth}) must be greater "
+                    f"than top depth ({self.top_depth})"
+                )
+        if self.top_depth < 0:
+            raise ValueError("Top depth cannot be negative.")
```
jacob-a-brown: numeric values should be checked and validated at the field itself using the [Field](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field) object. For the individual field you can set (for example)

```
top_depth: float = Field(gt=0)
```

You can also replicate this with bottom depth for greater than or equal

```
bottom_depth: float = Field(ge=0)
```

schemas/validators.py
```diff
@@ -0,0 +1,76 @@
...
+    Mixin to enforce that bottom_depth is greater than top_depth.
+    Assumes the model has 'top_depth' and 'bottom_depth' fields.
+    """
+
+    top_depth: float
+    bottom_depth: float
+
+    @model_validator(mode="after")
+    def check_depth_logical_order(self) -> "DepthIntervalMixin":
+        if self.bottom_depth is not None and self.top_depth is not None:
```
jacob-a-brown: since both `top_depth` and `bottom_depth` are required this is check isn't necessary

schemas/validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @field_validator("boundary")
+    @classmethod
+    def validate_wkt(cls, v: str | None) -> str | None:
+        if v is None:
+            return v
+
+        # Basic String Check
+        if not isinstance(v, str) or not v.strip():
+            raise ValueError("Boundary must be a valid WKT string.")
```
jacob-a-brown: this just checks that it's a string, but not a valid WKT. To do that you may need to coerce it into a geometry object and catch any errors. See **services/validation/geospatial.py** for how this is done (and it can be replicated). 

Also should that validator be moved to schemas?

schemas/geologic_formation.py
```diff
@@ -1,22 +1,51 @@
...
+
+
+class CreateThingGeologicFormationAssociation(BaseModel, DepthIntervalMixin):
+    """
+    Schema for linking a Thing (Well) to a GeologicFormation.
+    Uses DepthIntervalMixin to enforce bottom_depth > top_depth.
+    """
+
+    thing_id: int
+    geologic_formation_id: int | None = None
```
jacob-a-brown: shouldn't this be required?

ksmuczynski: Yes!!!

schemas/validators.py
ksmuczynski: What do you mean by move the validator to the schemas? I thought it was in schemas?

schemas/validators.py
ksmuczynski: Agreed. Since we want to treat these vocabulary terms as strict identifiers, I'll remove the custom case-insensitive logic and rely on Pydantic's native strict validation.

schemas/validators.py
jacob-a-brown: I meant move that validator function from **services/validation/geospatial.py** to the schemas directory somewhere.

schemas/validators.py
ksmuczynski: I'm not sure I'm following your suggestion. Are you saying the DepthIntervalMixin could be used to validate well screens and that the current well screen validation should inherit this mixin?

jacob-a-brown: That's what I'm saying - since it checks depths it can be used in various places where depths are defined.

jacob-a-brown: But we could add that to technical debt and keep it as-is for now

jacob-a-brown (approved):

PR #267, opened 2025-12-02
BDMS 227: well additional information

jacob-a-brown:
### **Why**
- This PR addresses the Gherkin file `well-additional-information.feature` by updating the API, schemas, models, and transfer scripts

### **How**
- Updated `Response` schemas: `ThingResponse`, `WellResponse`, `LocationGeoJSONResponse`
- Updated some `Create` schemas for the transfer scripts: `CreateWell`, `CreateLocation`
- Added `Response` schemas: `PermissionResponse`
- Added new database models: `GeologicFormation`, `AquiferSystem`, `AquiferType`
- Updated database models: `Thing`, `Permission`/`PermissionHistory`
- Updated transfer scripts
- Added new lexicon terms and enums

### **Notes**

- Used `cls.__tablename__` for polymorphic table relationships in the mixins
- The `ThingAquiferAssocation` model relates a `Thing` to an `AquiferSystem`. It also related `AquiferType`s (e.g. characteristics), to the system at that well.
- Formation and aquifer codes have been ported from NM_Aquifer, but they need to be cleaned up.
  - Formation names intentionally NOT stored in database; API will map formation codes to names using authoritative formations.json file that can be updated independently
- `start_date` is not nullable for `PermissionHistory` records, but it is in the response schema so that all types of permissions for a well can be returned. If a particular permission type does not exist, it will be in the returned list of permissions, but everything evaluates to `None`. This essentially allows the user to determine if the permission does or does not exist.
- `Permissions` was renamed `PermissionHistory` since it stores the history of permissions for an object (most often a `Thing`)
- permissions need to be transferred after both wells and contacts are transferred since it requires IDs from both tables

### **Work (Kind of) Reviewed**
We opened up PRs for smaller bits of work. They were reviewed, but this is the authoritative PR into staging. There are some notes and discussions in those PRs, so they're listed here for record-keeping:
- #248 
- #255 
- #256 
- #257
- #263
- #266

### **Jira Tasks**
Many Jira tasks, and subtasks, were addressed by this PR. They are:
- BDMS-227
- BDMS-229
- BDMS-230
- BDMS-231
- BDMS-232
- BDMS-254
- BDMS-258


jirhiker:
this changes were manuallly merged into PR #262

PR #268, opened 2025-12-04
transfer

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
-  ⚠️ 🚨 🦺  For safety, I changed the database port the testing suite uses.  port=54321. Updated docker-compose.yml


jacob-a-brown (approved):
Besides codex' comments about `chunk_by_size` I think that it looks good. I need to head out for the evening, and we are trying to get the transfer initiated tonight, so I'm approving it so it can be merged and transferred after codex' comments are addressed (whether changes need to be made or not)

PR #270, opened 2025-12-04
fix: specify database name in healthcheck for postgres service in docker

jacob-a-brown:
### **Why**
- If the database name is not specified it tries to connect to a database called **user**, which prevents the healthcheck from completing correctly, which causes problems down the line

### **How**
- specify `-d ${POSTGRES_DB}` in the health check

### **Notes**
- Use bullet points here


chasetmartin (approved):
lgtm

PR #274, opened 2025-12-05
fix: prevent sorting of measuring points when none are available for water wells

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/db/thing.py
+++ b/db/thing.py
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ===============================================================================
-from typing import List, TYPE_CHECKING
 from datetime import date
+from typing import List, TYPE_CHECKING
+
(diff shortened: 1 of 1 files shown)


jacob-a-brown:
Measuring points are required for each well. Are we transferring some where that's not the case?

jacob-a-brown:
It's a required field in `CreateWell`: https://github.com/DataIntegrationGroup/NMSampleLocations/blob/staging/schemas/thing.py#L133

Which is used to validate the incoming data: https://github.com/DataIntegrationGroup/NMSampleLocations/blob/staging/transfers/well_transfer.py#L343

I'll look to see if the `MeasuringPointEstimator` is setting it to `None` instead of raising an exception.

jirhiker:
removed hack from transfer branch to insert wells with no measuring point

PR #276, opened 2025-12-07
[NO TICKET]: Decrease load time for geospatil API 

jirhiker:
### **Why**
- Eager loading all Thing associations made the API too slow and non operational.

### **How**
- Comment out all use of eager loading for Thing associations. e.g. location_associations

### **Notes**
- with this simple fix performance increase 10X

### DIFF
--- a/db/thing.py
+++ b/db/thing.py
@@ -207,7 +207,7 @@ class Thing(
         cascade="all, delete-orphan",
         passive_deletes=True,
         order_by="LocationThingAssociation.effective_start.desc()",
-        lazy="joined",
+        # lazy="joined",
     )
 
     contact_associations = relationship(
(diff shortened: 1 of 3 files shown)


db/thing.py
```diff
@@ -349,7 +349,7 @@ class Thing(
     search_vector = Column(TSVectorType("name"))
 
     # for temporary backwards compatibility
-    well_construction_notes = mapped_column(String(1000), nullable=True)
+    # well_construction_notes = mapped_column(String(1000), nullable=True)
```
jacob-a-brown, line 352: `well_construction_notes` was removed from the `thing` model during the implementation of **well-notes.feature**. Rather than comment it out I think that it can be removed altogether.

jirhiker:
eager loading addressed by PR #277

PR #280, opened 2025-12-09
BDMS 414: updates to the contact response for searches

jacob-a-brown:
### **Why**
- The frontend 's needs the contact's `id` to allow the user to click on the contact and navigate immediately to the contact's detail page
- The frontend intends to display the associated things for a contact when it is searched

### **How**
- Added the contact's `id` to its `properties` field
- Added the associated things to its `properties` field. Because the `thing_type` is unknown until they have been retrieved the only fields returned in the array of things are:
  - the `name` (in the `label` field to correspond with how the names are returned elsewhere)
  - the `id`
  - the `thing_type`

### **Notes**
- When the `thing` model was refactored a while ago to allow multiple well purposes we forgot to update the `/search` endpoint. To fix this I removed `well_purpose` from the search's `well_response` and now return a field called `well_purposes` that evaluates to a list of purposes (i.e. a list of one or more strings)
  - e.g. `"well_purposes": ["Domestic", "Irrigation", "Observation"]` or `"well_purposes": ["Open, unequipped well"]`
- eager loading is used to get a contact's emails, phones, addresses, and things since it's unknown how many will be returned, and since they are used in every response from the search query

### DIFF
--- a/api/search.py
+++ b/api/search.py
@@ -26,6 +26,7 @@ from db import (
     Email,
     Phone,
     Address,
+    ThingContactAssociation,
     Thing,
     WellCasingMaterial,
     WellPurpose,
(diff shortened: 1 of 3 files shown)


TylerAdamMartinez (approved):
It's perfect 😍

PR #282, opened 2025-12-09
BDMS 389: well inventory csv updates

jacob-a-brown:
🚨 I added the feature file **well-inventory-csv.feature** to this PR so that the tests can be run against it 🚨 

### **Why**
- Fields that have been added for the implementation of **well-core-information.feature**, **well-additional-information.feature**, and **well-notes.feature** need to be handled by the `/well-inventory-csv` endpoint and the feature tests in **well-inventory-csv.feature**. 

### **How**
- New fields were mapped in both `CreateWell` and `CreateThing` schemas. They were placed in the former if water well-specific and added to the latter if they apply to all `thing_types`
- All models that relate to a `Thing`, such as `MonitoringFrequencyHistory` and `MeasuringPointHistory`, `Notes`, or `DataProvenance` are now handled in the function **services/thing_helper.py::add_thing**. All of this functionality was moved there to keep the code DRY since this is also used in POST `/thing/water-well` (and eventually other `thing_types`).
  - Models that are dependent on the creation of objects from other models, like `PermissionHistory`, are not handled by this function.
  - `add_thing` handles the creation of `LocationThingAssociation` and `GroupThingAssociation` already, so that logic was removed from **api/well_inventory.py**
- The `WellTransferer` was updated to exclude fields as necessary.
- The csv tables used for the tests were updated to use restricted vocabulary from the lexicon.

### **Notes**
- Not every field in the `WellInventoryRow` schema has a home in a model. I marked them with the inline comment `TODO: needs a home`. Should I create new fields for these? If so, some of them can probably be in the polymorphic notes table (`target_table=contact`), but others I'm not too sure. Thoughts on those fields?
  - Please make comments about them in **schemas/well_inventory.py** where I made the inline comments.
- The default port for the DB engine was changed to 54321 to correspond with the new Docker setup.
- Am I handling `LocationThingAssociation.effective_start` correctly in `add_thing`?
- @ksmuczynski, @chasetmartin and I determined that inventorying a well is a field activity, so a `FieldActivity` record is added too (and if a water level is determined/chemistry sample taken at the same time they can all be part of the same field event)


db/engine.py
```diff
@@ -109,7 +109,7 @@ def getconn():
     # elif driver == "postgres":
     password = os.environ.get("POSTGRES_PASSWORD", "")
     host = os.environ.get("POSTGRES_HOST", "localhost")
-    port = os.environ.get("POSTGRES_PORT", "5432")
+    port = os.environ.get("POSTGRES_PORT", "54321")
```
jirhiker: 54321 is only for testing and should not be default

jirhiker (changes requested):
looks good. Only issue is the port assignment

db/engine.py
jacob-a-brown: Because `5432` is used within Docker, but `54321` is used for testing, I set `POSTGRES_PORT=5432` in docker-compose.yml and then updated my environment file to use the correct port for testing.

I also updated `.env.example` to have the other environment variables in use (`POSTGRES_PORT` and `REBUILD_DB`)

jirhiker (approved):

PR #291, opened 2025-12-11
BDMS 389: historic well notes

jacob-a-brown:
### **Why**
- The Gherkin file **well-inventory-csv.feature** has a field for historic water levels. Since this isn't a measured water level, it may just be from the memory of the owner, it shouldn't go into the `observation` table. It should go into the polymorphic notes table instead with a `note_type` of `Historic`.

### **How**
- If `historic_depth_to_water_ft` is defined a note is added with the content being `f"Historic depth to water: {model.historic_depth_to_water_ft} ft"`.
  - This field doesn't indicate how/where the historic depth to water is from so I can't put that in the note. Should there be another field to indicate that? Or does this note suffice?

### **Notes**
- Use bullet points here


jirhiker:
Please follow up with @chasetmartin and Marissa to verify your assumption about the historic water level.  This seems reasonable; but " Since this isn't a measured water level, it may just be from the memory of the owner," I can't tell if that is your reasoning or based on feedback from AMP

chasetmartin:
> Please follow up with @chasetmartin and Marissa to verify your assumption about the historic water level. This seems reasonable; but " Since this isn't a measured water level, it may just be from the memory of the owner," I can't tell if that is your reasoning or based on feedback from AMP

One AMP member:
"Sometimes it’s what is in the drilling log
Sometimes it’s what the person just says to you"

I'll check with a couple others to make sure.

TylerAdamMartinez (approved):
lgtm

chasetmartin:
> I'll check with a couple others to make sure.

Confirmed during the Ocotillo Coordination meeting with AMP. @jacob-a-brown Also noted during the meeting that WellData.DepthSource from NM Aquifer refers to both well depth and historic depth to water.

jacob-a-brown:
Thoughts on the new note? Should I change the wording?

PR #298, opened 2025-12-12
jir-data-visibility-review-feature

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/tests/features/data-visibility-and-review.feature
+++ b/tests/features/data-visibility-and-review.feature
@@ -3,7 +3,7 @@
 
 @backend @BDMS-XXX
 Feature: Manage data visibility separately from review and approval in the backend
-  As an Ocotillo user
+  As an Ocotillo data manager
   I want visibility/privacy and review/approval to be stored as two separate backend attributes
   So that the system can independently control who can see the data and which data is reviewed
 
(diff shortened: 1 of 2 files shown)


chasetmartin (approved):
I like that this keeps some of the scenarios for potential future work as comments, but simplifies the feature for work now.

One question: is it challenging to test "according to the documented mapping" - or is that just a matter of having the mapping referenced somewhere in the code implementation?

tests/features/data-visibility-and-review.feature
```diff
@@ -16,78 +16,107 @@ Feature: Manage data visibility separately from review and approval in the backe
...
+    And all database models have a review_status field that supports "provisional" and "approved"
+    And visibility and review_status are required fields when creating new records
+    # hard to test?
+    And new records must use visibility and review_status as the source of truth
     And legacy combined visibility/review fields are deprecated
 
-  Scenario: Require visibility and review status when creating a new dataset
-    When the user attempts to create a new dataset via the api without specifying visibility or review_status
+
+  Scenario Outline: Require visibility and review status when creating a new record
```
jacob-a-brown, line 27: should these be required fields in the schemas? If so, the response will be 422 from Pydantic

marissafichera:
> I like that this keeps some of the scenarios for potential future work as comments, but simplifies the feature for work now.
> 
> One question: is it challenging to test "according to the documented mapping" - or is that just a matter of having the mapping referenced somewhere in the code implementation?

That's what I was thinking - I'm not sure what exactly the mapping will actually be at this moment but I figured Kelsey/Jacob would be able to implement that when they do the mapping

tests/features/data-visibility-and-review.feature
```diff
@@ -16,78 +16,107 @@ Feature: Manage data visibility separately from review and approval in the backe
...
   Background:
     Given a functioning api
-    And the dataset model has a visibility field that supports "internal" and "public"
-    And the dataset model has a review_status field that supports "provisional" and "approved"
-    And visibility and review_status are required fields when creating new datasets
-    And new datasets must use visibility and review_status as the source of truth
+    And all database models have a visibility field that supports "internal" and "public"
+    And all database models have a review_status field that supports "provisional" and "approved"
+    And visibility and review_status are required fields when creating new records
+    # hard to test?
```
jacob-a-brown, line 22: I don't think it'll be too hard to test. We can just make sure in `BaseCreateSchema` that those fields are always required for new records, so if they are not submitted Pydantic will raise errors.

tests/features/data-visibility-and-review.feature
```diff
@@ -16,78 +16,107 @@ Feature: Manage data visibility separately from review and approval in the backe
...
-  Scenario: Migrate legacy combined visibility/review values to separate fields
-    Given the system has legacy records with a combined visibility/review field
+  Scenario Outline: Migrate legacy combined visibility/review values to separate fields
+    Given the system has legacy records with a combined visibility/review field for <legacy_table>
     And there is a documented mapping from each legacy value to a visibility and review_status pair
-    When the migration job runs to populate visibility and review_status for legacy records
+    When the migration job runs to populate visibility and review_status for <legacy_table> records
     Then the system should set visibility according to the documented mapping
     And the system should set review_status according to the documented mapping
+    # hard to test?
```
jacob-a-brown, line 50: We can put these mappings into a test and then ensure that the function works correctly in some unit tests

tests/features/data-visibility-and-review.feature
jirhiker: So for this step you would assert that BaseCreateSchema has these fields and they are required?

tests/features/data-visibility-and-review.feature
jirhiker: hard to test? refers to "the system should stop using the legacy combined field as the source of truth
And subsequent updates to migrated records should modify visibility and review_status fields only".  How would you test for that

tests/features/data-visibility-and-review.feature
jirhiker: what test could be used to assert "And legacy combined visibility/review fields are deprecated"

tests/features/data-visibility-and-review.feature
jacob-a-brown: Off the top of my head we can go through every model in the database and ensure that none of the legacy fields exist and all that is there are `visibility` and `review_status`

tests/features/data-visibility-and-review.feature
jacob-a-brown: For the step `visibility and review_status are required fields when creating new records` I would assert that `BaseCreateSchema` has these fields and that every `Create...` schema inherits from `BaseCreateSchema`. Or, I suppose, we could test that every `Create...` schema has these fields, though it'd be nice to test the former to ensure that all `Create...` schemas have parity

PR #299, opened 2025-12-12
BDMS 389 inspired: status history refactor for wells

jacob-a-brown:
### **Why**
- The `thing` table had a field to store `is_suitable_for_datalogger`, but no way to store if it is open. After multiple discussions we came to the conclusion that both suitability for a datalogger at a well, as well as if it is open, are actually statuses since they are mutable. They can change over time. Because of this, those two pieces of information should be stored in the `status_history` table
- All fields in **WellData.CurrentUse** are being mapped to the `well_purpose` table. This includes the legacy `Open, unequipped well`. This is a status to indicate if a well is open or not, so should be recorded in the `status_history` table and not the `well_purpose` table.

### **How**
- Removed `is_suitable_for_datalogger` from the `Thing` model
- Added the `status_type` `Datalogger Suitability Status` to the lexicon, as well as the `status_values` `Datalogger can be installed` and `Datalogger cannot be installed` as the possible values
- Added the `status_type` `Open Status` to the lexicon, as well as the `status_values` `Open` and `Closed` as possible values
- Updated **transfers/well_transfer.py** to write to the `status_history` table for `Datalogger Suitability Status`.
- If **WellData.CurrentUser** evaluates to `Open, unequipped well` then a record is made in the `status_history` table where `status_type="Open Status"` and `status_value="Open"`. There is no way to determine if a well is closed, though.

### **Notes**
- After these changes are made to `staging` I'll merge with `well-inventory-csv` and make requisite updates.


jirhiker (approved):
looks good. seems like the project structure and style is starting to settle into place

jacob-a-brown:
Should we close this PR since it is now in `well-inventory-csv`?

PR #300, opened 2025-12-13
🤖 cli-water-levels

jirhiker:
### **Why**
- CLI for bulk uploading water levels via a csv. Well must already exist in database

### **How**
- Mostly AI (codex) implemented. Driven by `water-level-csv.feature`

### **Notes**
- The updates to the database models are cosmetic and organizational. I do not think this PR requires a database migration


jacob-a-brown (dismissed):
I think that this looks good for the current implementation of the water levels upload. Some changes will be coming, like making a new field for `data_quality` and mapping `level_status` to the `groundwater_level_reason` field, but when these updates occur these tests will be updated.

tests/features/environment.py
```diff
@@ -687,12 +688,16 @@ def before_all(context):
...
-                obj = session.get(record.__class__, record.id)
-                if obj:
-                    session.delete(obj)
+        for table in reversed(Base.metadata.sorted_tables):
+            if table.name in ("alembic_version", "parameter"):
+                continue
+            elif table.name.startswith("lexicon"):
+                continue
+            elif table.name == "sample":
+                continue
```
jacob-a-brown, line 697: if this was implemented because of a foreign key/non-nullable constraint from `sample` to `field_activity` that is going to be addressed in the DB models. @ksmuczynski and I talked about how we need to add `cascade="DELETE"` to the foreign key definition. I'm going to open a PR with this. 

So, if that is the case, I think that handling samples separately from other tables can be removed from this PR.

jirhiker:
I am merging this PR. I don't want PRs to remain unresolved for this long. The review comments are peripheral to the main point of this PR so it seems reasonable to merge into staging so that we are ready with cli functionality once the data and logic layers are complete

PR #307, opened 2025-12-16
NO TICKET: Update well inventory csv feature field names

chasetmartin:
### **Why**
- measuring_person and water_level_date_time field names were updated on the well inventory csv template to reduce user confusion and more easily differentiate the person performing the measurement from the field staff recorded at field event.

### DIFF
--- a/tests/features/well-inventory-csv.feature
+++ b/tests/features/well-inventory-csv.feature
@@ -125,9 +125,9 @@ Feature: Bulk upload well inventory from CSV
       | sample_possible                   |
      And the csv includes optional water level entry fields when available:
       | water_level_entry fields          |
-      | sampler                           |
+      | measuring_person                  |
       | sample_method                     |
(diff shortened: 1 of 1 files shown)


jacob-a-brown (approved):
These look good to me. On the backend we're going to need to ensure that the `measuring_person` is a participant in the field event (that's more of a note for me, I don't think it needs to change here).

This change should also go into the water level csv feature file.

chasetmartin:
@jacob-a-brown @jirhiker @ksmuczynski I updated this feature PR after talking with Jacob to account for the fact that users are entering data into the csv template as timezone naive strings, which are mountain time (the time they record in the field), but the system should handle assigning the proper timezone offset and conversion for storage based on the date.

jacob-a-brown (approved):

jacob-a-brown (approved):

PR #308, opened 2025-12-16
NO TICKET: update water level feature field names

chasetmartin:
### **Why**
- Add the ability for multiple field staff at field event, just like well inventory, and update the field names for measuring_person and water_level_date_time to match the csv template.

### DIFF
--- a/tests/features/water-level-csv.feature
+++ b/tests/features/water-level-csv.feature
@@ -63,8 +63,8 @@ Feature: Bulk upload water level entries from CSV via CLI
     And the CSV includes required fields:
       | required field name   |
       | well_name_point_id    |
-      | measurement_date_time |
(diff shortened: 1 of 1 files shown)


chasetmartin:
This PR is for updating the standalone water-level-csv feature for uploading water level measurements taken at an existing well. The required fields and field names are largely the same as the optional water level section of the well-inventory-csv feature documented here: https://github.com/DataIntegrationGroup/NMSampleLocations/pull/307

jacob-a-brown (commented):
Overall I think that it looks good, but per our work on the `well inventory csv upload` feature file, we should make it so that `measurement_date_time` is timezone-naive and that the backend will add a timezone offset.

jacob-a-brown (approved):

PR #311, opened 2025-12-17
NO TICKET: enable user to submit tz-naive datetime objects 

jacob-a-brown:
### **Why**
- All datetime fields are necessarily in Mountain Time since all locations must be in New Mexico. To facilitate the upload of data and ensure the user doesn't need to keep track of MST/MDT they now submit timezone-naive objects.
- These changes address the following updates to `well-inventory-csv.feature` that were made in PR #307:

> - And the required "date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T10:30:00")
> - And the optional "water_level_date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T10:30:00") when provided
> - Then all datetime objects are assigned the correct Mountain Time timezone offset based on the date value.

### **How**
- Add the timezone to the provided datetime fields and handle MST/MDT correctly with Python's `ZoneInfo` library
- Update the testing data to remove timezone offsets

### **Notes**
- Use bullet points here


chasetmartin (approved):
This looks good to me. It seems like if a future version of this well-inventory code is used to handle records sent individually from OcotilloUI (not from a csv), we'll have to circle back on the date/time format required.

jacob-a-brown:
That's definitely a good thing to keep in mind. Two things come to mind:

1. If users are only submitting water levels in New Mexico we can continue to use timezone-naive dt strings and continue as-is.
2. If users are submitting water levels from anywhere in the world, we can use the well's current location to determine the timezone and then add the timezone offset to the record appropriately

jirhiker:
I think for this case it's ok to assume that all datetimes will be in mountain time. However in general we should not assume all locations will be in New Mexico. Bureau researchers collect and analyze samples from all over Earth and even beyond.

PR #292, opened 2025-12-11
BDMS-372: data visibility and review feature

chasetmartin:
### **Why**
_This PR addresses the following problem / context:_
- Implementing new data visibility / review state management - Marissa

### **Notes**
- commented out well-sensor-deployment.py because there is no associated feature file currently. this may be revisited in the future

PR #312, opened 2025-12-17
BDMS-402: data visibility/review feature review

chasetmartin:
### **Why**
 - Some potential updates to the `data-visibility-and-review.feature` given the now documented legacy business logic in `release_status.feature`

### **How**
- rewrote `data-visibility-and-review.feature` scenarios to match the business logic covered in `release_status.feature` while preserving the new feature’s description and rules
- expanded public and staff access coverage so visibility and review_status behavior mirrors legacy workflows
- documented workflow, management, integrity, and special cases to clearly separate visibility and review in business logic mirroring current `release_status`

### **Notes**
**- This PR is just meant as a discussion point - Should we change the proposed feature like this to ensure that business logic currently documented in the `release_status.feature` is properly covered by the `data-visibility-and-review` proposal?**
 - Would merge into https://github.com/DataIntegrationGroup/NMSampleLocations/pull/292 if we want to make these changes

### DIFF
--- a/tests/features/data-visibility-and-review.feature
+++ b/tests/features/data-visibility-and-review.feature
@@ -2,9 +2,9 @@
 # file: tests/features/data_visibility_and_review_backend.feature
 
 @backend @BDMS-XXX
-Feature: Manage data visibility separately from review and approval in the backend
+Feature: Manage data visibility separately from review in the backend
   As an Ocotillo data manager
-  I want visibility/privacy and review/approval to be stored as two separate backend attributes
+  I want visibility and review to be stored as two separate backend attributes
   So that the system can independently control who can see the data and which data is reviewed
 
   # Business rules:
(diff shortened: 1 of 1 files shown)


tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
-#    Then the response should include a disclaimer field derived from the review_status
-#    And the disclaimer should indicate that the data is approved
-#
+# Created by marissa at 12/11/2025
+# file: tests/features/data_visibility_and_review_backend.feature
+
+@backend @BDMS-XXX
+Feature: Manage data visibility separately from review in the backend
+  As an Ocotillo data manager
+  I want visibility and review to be stored as two separate backend attributes
```
jacob-a-brown: Should this read `...and review status...` to reflect the verbiage used later on?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+  # - visibility has two states: "internal" or "public"
+  # - review_status has two states: "provisional" or "approved"
+  # - visibility and review_status are REQUIRED when creating new data
+  # - new data must use visibility and review_status (no new use of legacy fields)
+  # - legacy combined values will be migrated into visibility and review_status using a documented mapping
+  # - visibility and review must cover previous release_status business logic
+
+  Background:
+    Given a functioning api
+    And all database models have a visibility field that supports internal and public
```
jacob-a-brown: I think that `supports internal and public` is somewhat vague. What about the following?

> And all database models have a visibility field that determines if a record can or cannot be viewed by the public

or

> And all database models have a visibility field that if a record is for internal purposes only, or if it can always be viewed by the public

jacob-a-brown: If `internal` and `public` are the restricted vocabulary, I think that should be another step:

`And the only possible values for the visibility field are "internal" and "public"`

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And all database models have a review_status field that supports provisional and approved
```
jacob-a-brown: Akin to my comments above, I think that this can be split into two steps:

1. `And all database models have a review_status field that determines if a record is provisional or approved`
2. `And the only possible values for the review_status` field are "provisional" and "approved"

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And legacy combined visibility/review fields are deprecated
+
+  # ---------------------------------------------------------------------------
+  # PUBLIC ACCESS
+  # ---------------------------------------------------------------------------
+
+  @public_access @visibility
+  Scenario Outline: Public users can only access public data
+    Given I am a public API consumer
+    And there is <data_type> data stored with visibility public or internal
```
jacob-a-brown: To be more clear this can be updated to `And there is <data_type> data stored with a visibility field that evaluates to either "public" or "internal"`

(or something like that)

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    When I request <data_type> data through unauthenticated endpoints
+    Then I should only see records that are marked visibility public
```
jacob-a-brown: To be more specific I think that this can be updated to

`Then I should only see records where the visibility field is set to "public"`

(or something like that)

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And records marked internal should not be returned
```
jacob-a-brown: Akin to my above comments, this can be made a bit more specific

`And records whose visibility field is set to "internal" should not be returned`

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And the download should list the review_status for each record
+
+  # ---------------------------------------------------------------------------
+  # AUTHENTICATED STAFF ACCESS
+  # ---------------------------------------------------------------------------
+
+  @staff_access @visibility
+  Scenario Outline: Staff can access all data and its review status
+    Given I am an authenticated staff member
+    When I view <data_type> data through internal endpoints
```
jacob-a-brown: We do, in a sense, have internal endpoints with `/docs-auth`, but they are still "public" if someone knows of that URL. I think that this should indicate if the user is authenticated or not, since that is what will ultimately determine if they have access to internal or public records

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And the error response should indicate that review_status is required
+
+    Examples:
+      | data_type    |
+      | thing        |
+      | contact     |
+      | observation |
+
+  @workflow @defaults
+  Scenario: Safe defaults when both attributes are provided
```
jacob-a-brown: This scenario describes the user submitting the values "internal" and "provisional," not that they are the defaults. Should the scenario describe that if these values are not set that they are defaulted to "internal" and "provisional"?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    Given data is being submitted to the system with visibility internal and review_status provisional
+    When the record is created
+    Then the system should persist those explicit values as provided
+
+  @workflow @privacy_protection
+  Scenario: Private contact data is protected by default
```
jacob-a-brown: If _all_ contact data is private by default,  I think that this scenario should be "Contact data is private and protected by default," and then the rest of the scenario describes the workflow of setting the default values.

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    Then each record should have visibility populated according to the mapping
+    And each record should have review_status populated according to the mapping
+
+  # ---------------------------------------------------------------------------
+  # VISIBILITY AND REVIEW MANAGEMENT
+  # ---------------------------------------------------------------------------
+
+  @management @status_change
+  Scenario: Making internal data public without re-review
+    Given a thing is currently visibility internal and review_status approved
```
jacob-a-brown: Does this apply to other tables, too?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,202 @@
...
+    And appropriate authorization has been obtained
+    When staff changes the visibility to public
+    Then the thing should become visible in public endpoints
```
jacob-a-brown: I think that this should be "Then the thing should become visible by unauthenticated users" because we don't really have a distinction between public and private endpoints. That is determined by the user's authorization/authentication

jacob-a-brown (commented):
Overall I think it looks good! And describes these situations well. I think that some of the steps could be made a bit more clear, such as indicating that the `visibility` field is set to "internal" or "public."

I often made a comment in the first appearances of these situations and if you agree with my feedback then they should be applied elsewhere too.

tests/features/data-visibility-and-review.feature
chasetmartin: Is this better?

tests/features/data-visibility-and-review.feature
chasetmartin: let me know if this is better

tests/features/data-visibility-and-review.feature
chasetmartin: Good idea, let me know if these makes sense

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    Then each record should have visibility populated according to the mapping
+    And each record should have review_status populated according to the mapping
+
+  # ---------------------------------------------------------------------------
+  # VISIBILITY AND REVIEW MANAGEMENT
+  # ---------------------------------------------------------------------------
+
+  @management @status_change
+  Scenario: Making internal data public without re-review
+    Given a <data_type> is currently visibility internal and review_status approved
```
chasetmartin, line 151: @jacob-a-brown let me know if this makes more sense with the system?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+      | observation |
+
+  @workflow @defaults
+  Scenario: Safe defaults when both attributes are provided
+    Given data is being submitted to the system with visibility and review_status omitted
+    When the record is created
+    Then the system should persist visibility internal and review_status provisional by default
+
+  @workflow @privacy_protection
+  Scenario: Contact data is private and protected by default
```
chasetmartin, line 129: @jacob-a-brown updated with your suggestions, does this look good?

jacob-a-brown: I think that looks good

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+  # - visibility has two states: "internal" or "public"
+  # - review_status has two states: "provisional" or "approved"
+  # - visibility and review_status are REQUIRED when creating new data
+  # - new data must use visibility and review_status (no new use of legacy fields)
+  # - legacy combined values will be migrated into visibility and review_status using a documented mapping
+  # - visibility and review must cover previous release_status business logic
+
+  Background:
+    Given a functioning api
+    And all database models have a visibility field that can or cannot be viewed by the public
```
jacob-a-brown: How about "that determines if a record can or cannot be viewed by the public"?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+      | samples      |
+      | observations |
+
+  # might not be relevant yet in NMSampleLocations
+  @public_access @reports
+  Scenario: Public reports include review status disclaimers
+    Given there are public observations with review_status provisional and approved
+    When a public user generates a report
+    Then all returned observations should have visibility public
+    And provisional observations should include a disclaimer derived from review_status
```
jacob-a-brown, line 54: Would this disclaimer just be that `review_status` is `provisional`? Or is that TBD?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+
+  # ---------------------------------------------------------------------------
+  # PUBLIC ACCESS
+  # ---------------------------------------------------------------------------
+
+  @public_access @visibility
+  Scenario Outline: Public users can only access public data
+    Given I am a public API consumer
+    And there is <data_type> data stored with a visibility field that evaluates to either public or internal
+    When I request <data_type> data through unauthenticated endpoints
```
jacob-a-brown: We don't really have authenticated/unauthenticated endpoints - everything is determined via the user's permissions (or lack thereof). Since this scenario has the `Given` "I am a public API consumer," I think that this can just say 

> When I request <data_type> data through the API

Or something like that

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    Examples:
+      | data_type    |
+      | locations    |
+      | samples      |
+      | observations |
+
+  # might not be relevant yet in NMSampleLocations
+  @public_access @reports
+  Scenario: Public reports include review status disclaimers
+    Given there are public observations with review_status provisional and approved
```
jacob-a-brown: I think there can be an `And` after this `Given` that corresponds with the one above: `I am a public API consumer`. Then the step "When a public user generates a report" can be removed. This make the two scenarios specific to the same audience: the public/unauthenticated users

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    When a public user generates a report
+    Then all returned observations should have visibility public
+    And provisional observations should include a disclaimer derived from review_status
+    And approved observations should indicate that they are fully reviewed
+
+  # might not be relevant yet in NMSampleLocations
+  @public_access @data_download
+  Scenario: Public data downloads exclude internal datasets
+    Given a public user requests a CSV export
```
jacob-a-brown: Akin to my above comment, I think that the `Given` "I am a public API consumer" also applies here. So it could be something like

> Given I am a public API consumer
> And the dataset contains public and internal records with different review_status values
> When I request a CSV export
> Then the only the public records should be included
> And the download should list the review_status for each record

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    Then only the public records should be included
+    And the download should list the review_status for each record
+
+  # ---------------------------------------------------------------------------
+  # AUTHENTICATED STAFF ACCESS
+  # ---------------------------------------------------------------------------
+
+  @staff_access @visibility
+  Scenario Outline: Staff can access all data and its review status
+    Given I am an authenticated staff member
```
jacob-a-brown, line 72: I think that we should specify that the staff member has the correct permissions to view the data

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+
+    Examples:
+      | data_type    |
+      | things       |
+      | locations    |
+      | samples      |
+      | observations |
+
+  @staff_access @management
+  Scenario: Staff can change visibility without affecting review status
```
jacob-a-brown, line 86: I think that we should specify that the staff member is authenticated like the previous scenario. I also think that we should specify that the staff member has the correct permissions to edit the visibility field

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    Given data is stored with visibility internal and review_status provisional
+    When an authenticated staff member updates the visibility to public
+    Then the data visibility should be persisted as public
+    And the review_status should remain provisional
+    And retrieving the data should show the updated visibility and original review_status
+
+  @staff_access @review_status
+  Scenario: Staff can approve data without changing visibility
```
jacob-a-brown, line 96: I think that we should specify that the staff member is authenticated like the previous scenario. I also think that we should specify that the staff member has the correct permissions to edit the review_status field

tests/features/data-visibility-and-review.feature
jacob-a-brown: I just realized that the scenario above indicates these two fields are required. Does that contradict this scenario?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+  # ---------------------------------------------------------------------------
+  # VISIBILITY AND REVIEW MANAGEMENT
+  # ---------------------------------------------------------------------------
+
+  @management @status_change
+  Scenario: Making internal data public without re-review
+    Given a <data_type> is currently visibility internal and review_status approved
+    And appropriate authorization has been obtained
+    When staff changes the visibility to public
+    Then the <data_type> should become visible by unauthenticated users
```
jacob-a-brown: I think this should be

> Then the <data_type> should become visible to unauthenticated users

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,210 @@
...
+    Then only observations with visibility public should appear
+    And each observation should clearly display whether it is provisional or approved
+
+  # ---------------------------------------------------------------------------
+  # SPECIAL DATA REPRESENTATION
+  # ---------------------------------------------------------------------------
+
+  @special_cases @defaults
+  Scenario: Certain data types default to public and provisional
+    Given certain categories of data are configured to default to visibility public and review_status provisional
```
jacob-a-brown, line 206: Should these categories be enumerated?

tests/features/data-visibility-and-review.feature
chasetmartin: seems worth keeping for now since it mirrors the release_status.feature, but I agree it may conflict. It is a product decision we need to make: if a client submits a record without these two required fields, do we default them automatically to meet the requirements?

tests/features/data-visibility-and-review.feature
chasetmartin: TBD I think

tests/features/data-visibility-and-review.feature
chasetmartin: Used `Given I am a public API consumer` for both scenarios

tests/features/data-visibility-and-review.feature
chasetmartin: Made this two scenarios use the same staff given.

tests/features/data-visibility-and-review.feature
chasetmartin: Probably, but I'll keep this the same for now to compare to the release_status.feature

tests/features/data-visibility-and-review.feature
jacob-a-brown: I think this looks good! Will the staff member already have the permissions, or will be obtained for the actions? If they already possess the permissions then I think it could read `And staff has the correct permissions to change visibility`, or something to that effect. If they need to be given the permissions then I think it looks good as-is.

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,216 @@
...
+      | data_type    |
+      | thing        |
+      | contact     |
+      | observation |
+
+  @workflow @defaults
+  Scenario: Safe defaults when both attributes are omitted
+    Given data is being submitted to the system with visibility and review_status omitted
+    When the record is created
+    Then the system should persist visibility internal and review_status provisional by default
```
jacob-a-brown, line 126: should there be step that says what the default values are?

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,216 @@
...
+    Examples:
+      | data_type    |
+      | locations    |
+      | samples      |
+      | observations |
+
+  # might not be relevant yet in NMSampleLocations
+  @public_access @reports
+  Scenario: Public reports include review status disclaimers
+    Given a am a public API consumer
```
jacob-a-brown: there's a minor typo, `Given a...` should be changed to `Given I...`

jacob-a-brown (commented):
I think it looks quite good! Just a few minor comments.

tests/features/data-visibility-and-review.feature
```diff
@@ -1,122 +1,216 @@
...
+    And the download should list the review_status for each record
+
+  # ---------------------------------------------------------------------------
+  # AUTHENTICATED STAFF ACCESS
+  # ---------------------------------------------------------------------------
+
+  @staff_access @visibility
+  Scenario Outline: Staff can access all data and its review status
+    Given I am an authenticated staff member
+    And I have permission to view internal data
```
jirhiker, line 73: We will need to be more granular than this. E.g only amp users can few amp data, etc.

jirhiker (changes requested):

kbighorse:
I think if the goal of this PR is to make `review_status.feature` redundant, AND you think that has been accomplished, then I would delete `review_status.feature` in this PR and then merge it (safe, since it's not merging to `staging`). If you want to address comments here so they're not lost, that is also possible. Otherwise, you could re-address them in the parent branch, I might recommend this to avoid polluting the purpose of this PR.

Once you merge this, you can continue negotiating the resultant `data-visibility-and-review.feature` file on its existing PR (and wait to merge that until @jirhiker returns), which you are already doing here. But I would merge this as soon as possible once the objective is achieved, which it looks to me, is.

PR #313, opened 2025-12-17
NO TICKET: m to ft fix

jacob-a-brown:
### **Why**
- There were duplicate functions to convert from meters to feet
- The number of digits in the rounding was hardcoded. It is now an argument in the function and defaults to 6, which is what was used previously
- util functions need to be tested

### **How**
- Removed one of the duplicate functions
- Enabled the number of digits in the round function to be set as an argument
- Started a test file for **services/util.py** and added tests for `convert_ft_to_m` and `convert_m_to_ft`

### **Notes**
- Use bullet points here

### DIFF
--- a/services/util.py
+++ b/services/util.py
@@ -45,25 +45,18 @@ def transform_srid(geometry, source_srid, target_srid):
     return transform(transformer.transform, geometry)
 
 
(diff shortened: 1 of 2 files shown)


jirhiker (approved):

chasetmartin (approved):
lgtm

PR #314, opened 2025-12-18
BDMS 350: add groundwater_level_accuracy to observation model

jacob-a-brown:
### **Why**
- When groundwater level observations are added to `Ocotillo` we need a way to track their accuracy, formerly known as `DataQuality` in `NM_Aquifer`
- AMP requested that these value continue to be stored and retrievable, particularly because of their correspondence with USGS

### **How**
- Updated the lexicon to call these data `groundwater_level_accuracy` instead of `data_quality` because every string describes the accuracy level. If, in the future, we want to change it into a numeric field we can certainly do that
- Added `groundwater_level_accuracy` to the `observation` model and both the `CreateGroundwaterLevelObservation` and  `GroundwaterLevelObservationResponse` schemas
- Updated tests for this field
- Updated the water level transfer script to map from `WaterLevels.DataQuality` to `observation.groundwater_level_accurady`

### **Notes**
- This update was instigated by the field `data_quality` in the water level upload
- This PR is being opened into the `water-level-csv` because that is where the work is needed. Eventually, when ready, we'll open a PR from that branch into `staging` (then, as I wrote in Slack, this work will get applied to the well inventory csv upload if they contain water level data)


transfers/waterlevels_transfer.py
```diff
@@ -197,7 +197,17 @@ def _make_observation(
         else:
             value = row.DepthToWater
 
-            # TODO: after sensors have been added to the database update sensor_id (or sensor) for waterlevels that come from db sensors (like e probes?)
+        if not pd.isna(row.DataQuality):
```
jirhiker: This needs a guard for invalid DataQuality values ie ones not in the lexicon

jirhiker (changes requested):

transfers/waterlevels_transfer.py
jacob-a-brown: The check is now being done in the outer for loop and if it can't be mapped the error is logged and the loop continues to the next iteration

docker-compose.yml
```diff
@@ -28,6 +28,7 @@ services:
       - POSTGRES_DB=${POSTGRES_DB}
       - POSTGRES_HOST=db
       - MODE=${MODE}
+      - POSTGRES_PORT=5432
```
chasetmartin, line 31: @jacob-a-brown Was this change intentional?

jacob-a-brown: This change was intentional. Outwardly the database is connected to the API through the port 54321, but in the docker environment it is connected through the port 5432. To allow these different ports to be used in different environments, it should be hardcoded internally but then set outside docker via an environment variable. If it's not hardcoded within the docker environment then the environment variable `POSTGRES_PORT` will get applied and the connection cannot be made as needed.

jirhiker (approved):

chasetmartin (approved):
lgtm, thanks for the clarification

PR #332, opened 2026-01-05
BDMS 350: water level csv feature test update

jacob-a-brown:
### **Why**
- `measuring_person` is the field staff that took the measurement, not a lexicon value

### **How**
- Removed `measuring_person` from the scenario `Upload fails due to invalid lexicon values`
- Added a scenario to test when `measuring_person` does not correspond with `field_staff`, `field_staff_2`, or `field_staff_3` 

### **Notes**
- Use bullet points here

### DIFF
--- a/tests/features/water-level-csv.feature
+++ b/tests/features/water-level-csv.feature
@@ -45,7 +45,9 @@ Feature: Bulk upload water level entries from CSV via CLI
       """
       oco water-levels bulk-upload --file ./water_levels.csv --output json
       """
-    Then the command exits with code 0
+    # assumes users are entering datetimes as Mountain Time becuase well location is restricted to New Mexico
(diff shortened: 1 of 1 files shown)


jirhiker (approved):

chasetmartin (approved):

jacob-a-brown:
@jirhiker and @chasetmartin I made another minor update for review. Since it pertains to the feature file I added it to this PR:

In the summary I changed `validation_errors_or_warnings` to `total_validation_errors_or_warnings` because the summary just reports the total numbers, not the actual errors/warnings. Calling it `validation_errors_or_warnings` led to confusion since I was expecting it to list out those errors/warnings. Prepending `total_` to the field name clarifies what the field communicates.

The actual errors/warnings are reported in the payload, the standard out, and the standard error.

chasetmartin (approved):

PR #336, opened 2026-01-07
Implement backfill for legacy WaterLevelsContinuous_Pressure_Daily table 

jirhiker:
### Why

- Need to store legacy daily pressure water level records in our schema.
- Need a script to backfill those records from the existing CSV data source.

### How

- Added Alembic migration b6c986eff2a8 to create the legacy NMA_WaterLevelsContinuous_Pressure_Daily table with all expected
  columns.
- Added a backfill helper under transfers/backfill/ that loads the WaterLevelsContinuous_Pressure_Daily CSV, parses key dates,
  builds model instances, and bulk-inserts in batches via the Transferer flow, with a run(batch_size=...) entrypoint.

### Notes

- No automated tests were run; backfill and migration should be exercised in a suitable environment before deployment.


jirhiker:
@codex review

PR #349, opened 2026-01-08
BDMS-453

jirhiker:
### Why

- Add legacy HydraulicsData support for backfill workflows and admin visibility

### How

- Added NMA_HydraulicsData legacy model in db/nma_legacy.py and corresponding Alembic migration
- Implemented a HydraulicsData backfill script and wired it into transfers/backfill/staging.py
- Added a read-only admin view for Hydraulics data and registered it in the admin config

### Notes

- Backfill is idempotent via ON CONFLICT DO UPDATE on GlobalID

### DIFF
--- a/admin/config.py
+++ b/admin/config.py
@@ -36,11 +36,11 @@ from admin.views import (
     GroupAdmin,
     NotesAdmin,
     SampleAdmin,
+    HydraulicsDataAdmin,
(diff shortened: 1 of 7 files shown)


jirhiker:
there is no hydraulics data in the HydraulicsData table

PR #350, opened 2026-01-08
BDMS-455

jirhiker:
### Why

  - Introduce legacy SurfaceWaterData support for ingestion and admin visibility

  ### How

  - Added SurfaceWaterData legacy model in db/nma_legacy.py with UUID surface_id
  - Added Alembic migration to create NMA_SurfaceWaterData
  - Added SurfaceWaterData backfill script in transfers/backfill/surface_water_data.py
  - Added admin view for SurfaceWaterData and registered it in the admin config

  ### Notes

  - If the table already exists, use alembic stamp 1680a4a7cb77 to align migration history

### DIFF
--- a/admin/config.py
+++ b/admin/config.py
@@ -43,6 +43,7 @@ from admin.views import (
     FieldEventAdmin,
     FieldActivityAdmin,
     ParameterAdmin,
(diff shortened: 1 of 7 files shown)


jirhiker:
@codex review

PR #364, opened 2026-01-10
NO-TICKET: Consolidate backfill into transfers

jirhiker:
### What ###

  - Moved legacy transfer scripts for Chemistry SampleInfo, NGWMN views, SurfaceWaterData,
    WaterLevelsContinuous_Pressure_Daily, and WeatherData from transfers/backfill/ into the main transfers/ package and
    renamed their classes to standard *Transferer.
  - Wired these legacy transfers into the main transfer pipeline (transfers/transfer.py) in both parallel and
    sequential flows, with new feature flags to toggle them.
  - Added metrics reporting for the newly-integrated legacy tables in transfers/metrics.py.
  - Updated backfill orchestrator imports to point at the new module locations so the backfill pipeline still runs.
  - Added unit tests for the moved legacy models mirroring the style of tests/test_weather_data_legacy.py.
  - Added an ERD generator script and produced current schema ERD outputs.

  ### Why ###

  - These legacy datasets are now part of the standard transfer workflows and should live in the same module structure
    as the rest of the transfer logic.
  - Integrating them into transfers/transfer.py ensures consistent orchestration, logging, and metrics.
  - New tests verify legacy schema mappings and CRUD behaviors for the moved models, preventing regressions.
  - ERD output provides an up-to-date view of the schema for review and documentation.

  ### Key Changes ###

  - Module moves + renames
      - transfers/backfill/chemistry_sampleinfo.py → transfers/chemistry_sampleinfo.py
      - transfers/backfill/ngwmn_views.py → transfers/ngwmn_views.py
      - transfers/backfill/surface_water_data.py → transfers/surface_water_data.py
      - transfers/backfill/waterlevelscontinuous_pressure_daily.py → transfers/waterlevelscontinuous_pressure_daily.py
      - transfers/backfill/weather_data.py → transfers/weather_data.py
      - Class names updated to ChemistrySampleInfoTransferer, NGWMN*Transferer, SurfaceWaterDataTransferer,
        NMAWaterLevelsContinuousPressureDailyTransferer, and WeatherDataTransferer.
  - Transfer orchestration
      - Added transfer flags (default True):
          - TRANSFER_SURFACE_WATER_DATA
          - TRANSFER_CHEMISTRY_SAMPLEINFO
          - TRANSFER_NGWMN_VIEWS
          - TRANSFER_WATERLEVELS_PRESSURE_DAILY

### Testing ###
  - python -m pytest tests/test_chemistry_sampleinfo_legacy.py tests/test_surface_water_data_legacy.py tests/

### Notes ###
  - Running tests requires a local Postgres instance reachable at localhost:54321 with env vars loaded from .env.
  - ERD PNG generation requires Graphviz (dot) installed.
  - **Backfill will be used in the future for refactoring not for legacy transfer**


run_backfill.sh
chasetmartin, line 1: Should this file just be removed entirely if it won't work given the transfer setup?

chasetmartin (approved):
Looks good to me given my previous knowledge of the transfer process, I do have one comment about removing the unused `run_backfill.sh` file if it is not relevant.

transfers/chemistry_sampleinfo.py
```diff
@@ -48,7 +49,7 @@ def _get_dfs(self) -> tuple[pd.DataFrame, pd.DataFrame]:
     def _transfer_hook(self, session: Session) -> None:
         rows = self._dedupe_rows(
             [self._row_dict(row) for row in self.cleaned_df.to_dict("records")],
-            key="OBJECTID",
+            key="SamplePtID",
```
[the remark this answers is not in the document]
jirhiker: SamplePtID is never null

tests/conftest.py
```diff
@@ -1,10 +1,61 @@
...
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _setup_test_db():
+    _reset_schema()
+    command.upgrade(_alembic_config(), "head")
+    _sync_search_vectors()
+    init_lexicon()
+    init_parameter()
+    yield
```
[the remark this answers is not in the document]
jirhiker: The schema reset is currently session-scoped (runs once per session), and per-test cleanup relies on deletes/
  fixtures. We could switch to a transaction-per-test pattern (nested transaction/rollback) to speed things up, but it
  requires routing all DB access through the same connection/session (including the FastAPI test client).

alembic/versions/b7d4c6a1b2c3_create_chemistry_sampleinfo.py
```diff
@@ -25,24 +26,48 @@ def upgrade() -> None:
...
-            sa.Column("AnalysesAgency", sa.String(length=100), nullable=True),
-            sa.Column("SampleType", sa.String(length=100), nullable=True),
-            sa.Column("SampleMaterialNotH2O", sa.Boolean(), nullable=True),
-            sa.Column("WaterType", sa.String(length=100), nullable=True),
-            sa.Column("StudySample", sa.Boolean(), nullable=True),
+            sa.Column("SamplePtID", postgresql.UUID(as_uuid=True), primary_key=True),
+            sa.Column("WCLab_ID", sa.String(length=18), nullable=True),
+            sa.Column(
+                "SamplePointID", sa.String(length=10), nullable=False, unique=True
+            ),
```
jacob-a-brown: The unique constraint doesn't exist in NM_Aquifer. Do we want to enforce it? For data fidelity and transfer.

db/nma_legacy.py
```diff
@@ -164,45 +165,52 @@ class ChemistrySampleInfo(Base):
...
     __tablename__ = "NMA_Chemistry_SampleInfo"
 
-    object_id: Mapped[int] = mapped_column("OBJECTID", Integer, primary_key=True)
+    sample_pt_id: Mapped[uuid.UUID] = mapped_column(
+        "SamplePtID", UUID(as_uuid=True), primary_key=True
+    )
+    wclab_id: Mapped[Optional[str]] = mapped_column("WCLab_ID", String(18))
+    sample_point_id: Mapped[str] = mapped_column(
+        "SamplePointID", String(10), nullable=False, unique=True
+    )
```
jacob-a-brown, line 174: Akin to the alembic comment: this is not unique in NM_Aquifer. Do we want to enforce it here?

services/util.py
jacob-a-brown, line 1: I like how in some warnings the lat/long is in the message. I think it would be good for all of them to communicate `POINT {lon} {lat}` in the error messages.

jacob-a-brown (commented):
Overall I think it looks good! I have some minor comments scattered throughout. Most of my comments are about Starlette-Admin, so I've compiled them below.

**Comments on Starlette-Admin**
1. I don't think that [Starlette-Admin](https://jowilf.github.io/starlette-admin/user-guide/configurations/modelview/) has a separate list view and form view. Looking at its current implementation I think that you can get rid of all of the attributes in the "List View" sections and then keep the following:

- `fields`
- `sortable_fields`
- `fields_default_sort`
- `searchable_fields`
- `page_size`
- `page_size_options`
- `exclude_fields_from_create`
- `exclude_fields_from_edit`

You can also define export types as desired (if you want all views to have the same export types the `export_type` attribute can be defined in `OcotilloModelView`). For example:

```python
rom starlette_admin import BaseField
from starlette_admin import ExportType
from admin.views.base import OcotilloModelView

class ExampleModelAdmin(OcotilloModelView):

    fields = [...]
    sortable_fields = [...]
    fields_default_sort = [...]
    searchable_fields = [...]
    page_size = ...
    page_size_options = [...]
    exclude_fields_from_create = [...]
    exclude_fields_from_edit = [...]
    export_type = [ExportType.CSV, ExportType.EXCEL]
```

2. There may be some discrepancies between what is enumerated in `list_fields` and `fields`. These should be resolved and then put in the intended order in the `fields` attribute (e.g. putting `sample_pt_id` first for `ChemistrySampleInfo` since that's the PK)
3. Unfortunately there isn't a way to list out all of the labels in a dictionary or similar object. To have labels, each field in the `fields` list needs to be defined as a [BaseField](https://jowilf.github.io/starlette-admin/api/fields/) or a particular field type (like `IntegerField` or `StringField`, but I think that'd be overkill) and then each label needs to be defined one-at-a-time. The same goes for help text. For example:

```python
from starlette_admin import BaseField
from admin.views.base import OcotilloModelView

class ExampleModelAdmin(OcotilloModelView):

    fields = [
        BaseField("field_one", label="FieldOne", help_text="Help for FieldOne"),
        BaseField("field_two", label="FieldTwo", help_text="Help for FieldTwo")
    ]

PR #377, opened 2026-01-14
BDMS-463: Docker Compose Checkup

chasetmartin:
### **Why**
- Prevent dev startup from wiping and rebuilding the database on every run
- Ensure lexicon/parameter data is available before seeding
- Avoid duplicate seed inserts when data already exists

### **How**
- Removed erase-and-rebuild from the dev lifespan flow in core/app.py
- Moved lexicon/parameter initialization into the seed path since seed.py errors if lexicon does not exist
- Added a “skip if data exists” guard to seeding

### **Notes**
- Seeding now no-ops if a single contact exist; adjust the guard if a different table is preferred

### DIFF
--- a/core/app.py
+++ b/core/app.py
@@ -37,7 +37,6 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     Application lifespan event handler to initialize the database and lexicon.
     """
     if settings.get_enum("MODE") == "development":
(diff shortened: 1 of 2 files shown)


jirhiker (approved):

chasetmartin:
@jirhiker Ok print statements added and I changed the function name as suggested.

PR #378, opened 2026-01-14
BDMS-464

jirhiker:
Add nma_formation_zone to Thing

Why

- The legacy WellData FormationZone (nvarchar(25)) needs to be preserved on Thing.
- We already store a normalized formation code, but the raw legacy value is needed for audit and fidelity.

What changed

- Added nma_formation_zone to Thing (nullable String(25)) and exposed it in Thing schemas for create/update/response.
- Updated well transfer to copy FormationZone into nma_formation_zone while keeping existing normalization logic.
- Added a migration to add nma_formation_zone to thing and thing_version.
- Added a small schema unit test for nma_formation_zone.

Files touched

- db/thing.py
- schemas/thing.py
- transfers/well_transfer.py
- alembic/versions/f1a2b3c4d5e6_add_nma_formation_zone_to_thing.py
- tests/test_thing.py

Testing

- pytest tests/test_thing.py -k nma_formation_zone

### DIFF
--- a/alembic/versions/f1a2b3c4d5e6_add_nma_formation_zone_to_thing.py
+++ b/alembic/versions/f1a2b3c4d5e6_add_nma_formation_zone_to_thing.py
@@ -1,0 +1,52 @@
+"""Add nma_formation_zone to Thing.
+
+Revision ID: f1a2b3c4d5e6
+Revises: f3b4c5d6e7f8
+Create Date: 2026-03-01 00:00:00.000000
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy import inspect
+
+# revision identifiers, used by Alembic.
+revision: str = "f1a2b3c4d5e6"
+down_revision: Union[str, Sequence[str], None] = "f3b4c5d6e7f8"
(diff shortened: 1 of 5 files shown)


ksmuczynski:
The `GeologicFormation` model has a `formation_code` field that ultimately should/will be used to store NMA `FormationZone` values. If formation zone info needs to be stored in the Thing model, too, for the puposes of migrating these legacy tables that's fine, just wanted to point out that Ocotillo does currently have a place for FormationZone info.

jacob-a-brown:
The formation zone of the well's completion is found in the `formation_completion_code` field of the [WellResponse model](https://github.com/DataIntegrationGroup/OcotilloAPI/blob/3f907017048f0a7c2dbe2d0f077b3e2ad1d3f7c9/schemas/thing.py#L250)

jirhiker:
@ksmuczynski @jacob-a-brown although Ocotillo has a place for FormationZone, a large number of the formation zones in NMA are not in the Ocotillo lexicon, and in many cases there are multiple formation zones for a given well, e.g. 211DKOT/217PRGR/110AVMB. nma_formation_zone is to facilitate bringing this data over as is.  Refactoring into the GeologicFormation model will be done later

PR #390, opened 2026-01-15
BDMS-468: Preserve legacy empty strings in NMA transfer pipelines

jirhiker:
### Why

- Preserve empty-string values from legacy sources during migration for specific NMA datasets.
- Ensure blank strings are not coerced into NULLs for legacy string fields.

### How

- Read targeted legacy CSVs with keep_default_na=False to keep empty strings.
- Added blank‑safe numeric/date parsing to map empty strings to NULL only for non-string fields.
- Added unit tests to prove "" is preserved for string fields in the specified transfers.

### Notes

- Full test suite: source .venv/bin/activate && set -a && source .env && set +a && pytest

### DIFF
--- a/tests/test_transfer_preserve_empty_strings.py
+++ b/tests/test_transfer_preserve_empty_strings.py
@@ -1,0 +1,217 @@
+# ===============================================================================
+# Copyright 2026 ross
+#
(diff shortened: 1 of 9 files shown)


chasetmartin (approved):
Makes sense to me for all of these nma 1:1 transfers.

PR #392, opened 2026-01-16
chore: update pre-commit configuration to disable pytest hook

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


chasetmartin (approved):
lgtm

PR #393, opened 2026-01-16
BDMS-470: Migrate Stratigraphy

jirhiker:
### Why

- Staging lacks a first-class legacy stratigraphy table, so Stratigraphy.csv rows aren’t preserved outside the association transfer.
- Without a dedicated transfer + tests, stratigraphy CSV ingestion can regress silently.
- Alembic/test/transfer flows don’t share a single schema reset path for stratigraphy legacy data.

### How

- Added Alembic revision 1d2c3b4a5e67_create_nma_stratigraphy_table.py plus the Stratigraphy model (NMA_Stratigraphy table) and wired Thing’s relationship so legacy intervals can be queried off wells.
- Built transfers/stratigraphy_legacy.py with a batch upsert pipeline, hooked it into the main transfer worker (guarded by TRANSFER_NMA_STRATIGRAPHY), and exposed metrics for the new table.
- Created tests/test_stratigraphy_legacy.py to cover the model FK linkage and the transfer’s CSV ingestion path end to end.

### Notes

- Run the new transfer after wells load so PointID→Thing resolutions succeed.
- The Alembic revision must be applied before running transfers/tests to ensure NMA_Stratigraphy exists.


transfers/transfer.py
chasetmartin, line 1: @jirhiker I'm getting a Typeerror when I try to run the transfer on this branch, I think because the `dtype` is passed twice to read_csv in the `transferer.py` file. Let me know if that's just an issue I'm seeing on my end.
https://github.com/DataIntegrationGroup/OcotilloAPI/blob/76dcc29e936d189be1372ebd62f68425f55458e3/transfers/transferer.py#L141C9-L141C49

PR #394, opened 2026-01-16
BDMS-430 Migrate legacy FieldParameters chemistry table

ksmuczynski:
### **Why**
  - Legacy FieldParameters data needed an ORM model, migration, tests, and transfer path to backfill into the new schema
  - Ensure FieldParameters records are validated against existing ChemistrySampleInfo entries

### **How**
- Added NMAFieldParameters model to db/nma_legacy.py
  - Created Alembic migration alembic/versions/c1d2e3f4a5b6_create_nma_field_parameters.py to build the legacy table
  - Added unit tests in tests/test_field_parameters_legacy.py to validate mappings and CRUD behavior
  - Implemented transfer script transfers/field_parameters_transfer.py with FK validation and batch upsert

### **Notes**
  - Transfer depends on ChemistrySampleInfo being populated first


chasetmartin (changes requested):
This is looking great! The two copilot comments do seem relevant. 

Migration ran ok for me, and tests passed. I think the one missing item is updating the transfer.py file so that that we can run and test the transfer of field parameters. I would look to this Stratigraphy PR for how to update `transfer.py`:
https://github.com/DataIntegrationGroup/OcotilloAPI/pull/393/changes#diff-d2d33d2e5666e926931898df688d5d66a04e7d8148800dec2399b1de947e1fbf

PR #419, opened 2026-01-23
[BDMS-492] Create new admin view for the legacy stratigraphy table

TylerAdamMartinez:
### **Why**
- Legacy stratigraphy data was not accessible in the admin UI

### **How**
- Added `StratigraphyAdmin` using the shared `OcotilloModelView` base

### **Notes**
- This change assumes the `NMA_Stratigraphy` relation exists in the connected database

### DIFF
--- a/admin/views/__init__.py
+++ b/admin/views/__init__.py
@@ -19,59 +19,61 @@ Admin views package for NMSampleLocations.
 Provides MS Access-like interface for CRUD operations on database models.
 """
 
-from admin.views.location import LocationAdmin
-from admin.views.thing import ThingAdmin
-from admin.views.observation import ObservationAdmin
-from admin.views.contact import ContactAdmin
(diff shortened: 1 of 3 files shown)


admin/views/__init__.py
TylerAdamMartinez, line 1: Updated the imports and the `__all__` list to be in alphabetical order.

admin/views/stratigraphy.py
```diff
@@ -0,0 +1,84 @@
+"""
+StratigraphyAdmin view for legacy Chemistry_SampleInfo.
```
TylerAdamMartinez: Don't know the legacy table name

TylerAdamMartinez:
Captura de pantalla 2026-01-23 a la(s) 10 36 53

chasetmartin (approved):
lgtm, nice!

PR #422, opened 2026-01-23
[NO TICKET ] Mv guard & logger up before list() to prevent crashes in transfers/profiling.py

TylerAdamMartinez:
### **Why**
- `upload_profile_artifacts()` could crash with `TypeError: 'NoneType' object is not iterable` when profiling artifacts were not present.

### **How**
- Updated `upload_profile_artifacts` to accept `Optional[Iterable[ProfileArtifact]]` instead of requiring an iterable
- Added an early-return guard for `None`/empty artifacts and logged a helpful message when there’s nothing to upload.
- Moved `list(artifacts)` conversion to occur only after the `None`/empty check to avoid exceptions.

### **Notes**
- Please see this PR before merging #420

### DIFF
--- a/transfers/profiling.py
+++ b/transfers/profiling.py
@@ -20,7 +20,7 @@ import pstats
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import Callable, Iterable, Any
+from typing import Callable, Iterable, Any, Optional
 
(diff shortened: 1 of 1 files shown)


chasetmartin (approved):
looks good. @jirhiker Just tagging you since the profiler is your recent addition so you see this.

PR #425, opened 2026-01-24
[BDMS-495] Update sidebar label on surface water data page

TylerAdamMartinez:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


jirhiker (dismissed):
Its standard practice in this repo to include the License header and EOF

chasetmartin (approved):
Thanks!

PR #429, opened 2026-01-24
bugfix: increase length of LithologicModifier field to accommodate larger values

jirhiker:
### **Why**
transfer was failing to transfer NMA_Stratigraphy because of inaccurate ORM/alembic definition

```CRITI [transfers.logger] Parallel task StratigraphyLegacy failed: (pg8000.exceptions.DatabaseError) {'S': 'ERROR', 'V': 'ERROR', 'C': '22001', 'M': 'value too long for type character varying(100)', 'F': 'varchar.c', 'L': '640', 'R': 'varchar'}```


chasetmartin (approved):
Nice, looking forward to trying this in the transfer

We should continue updating an AGENTS.MD as needed, good to add it.

PR #432, opened 2026-01-26
BDMS-503: Implement read-only admin view for NMAMajorChemistry

ksmuczynski:
### **Why**
  - Provide a read-only admin view for legacy NMA Major Chemistry records
  - Ensure the admin UI exposes all legacy columns with correct ordering and labels

### **How**
- Added MajorChemistryAdmin view with read-only permissions, list/sort/search config, and labels
  - Included all model columns in list_fields and fields in legacy order, plus the HasOne parent link
  - Registered the new view in admin/views/__init__.py and admin/config.py

### **Notes**
- None noted

### DIFF
--- a/admin/config.py
+++ b/admin/config.py
@@ -38,6 +38,7 @@ from admin.views import (
     LexiconCategoryAdmin,
     LexiconTermAdmin,
     LocationAdmin,
+    MajorChemistryAdmin,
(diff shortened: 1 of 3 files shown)


chasetmartin (approved):
nice, lgtm

PR #435, opened 2026-01-27
BDMS-512: AOUIWL Well Intel Equipment information to be preserved during the migration

jirhiker:
### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


chasetmartin (changes requested):
Can you just update the alembic migration to revise the now current staging head (I think 3a9c1f5b7d2e)

After that you should merge in and trigger a transfer during your AUS day. Otherwise I'll do this tomorrow.

PR #464, opened 2026-02-05
Data migrations framework + new site-type transfers, Typer CLI, and NMA notes backfill

jirhiker:
**Summary**
- Added a repeatable data migration framework tied to Alembic revision state with CLI tooling and tests.
- Added new non‑well site-type transfers and lexicon entries.
- Migrated CLI from Click to Typer.
- Added a data migration to move `Location.nma_location_notes` into the polymorphic `Notes` table.

**Key Changes**
- **Data migrations**
  - `data_migrations/` framework with registry, runner, history table, and Typer CLI.
  - Enforces applied Alembic revision before running a data migration.
  - Added template under `data_migrations/migrations/_template.py`.
  - New migration: `20260205_0001_move_nma_location_notes`.

- **Transfers**
  - Added per‑site-type transfer functions for new `LU_SiteType` entries.
  - Added env toggles for each new transfer.
  - Updated non‑well task selection logic.

- **CLI**
  - Switched CLI to Typer.
  - Added `data-migrations` command group and `alembic-upgrade-and-data`.

- **Tests**
  - Added unit tests for new site-type transfer helpers.
  - Added tests for data migration CLI and the notes backfill.

**Behavioral Notes**
- Data migrations now require the referenced Alembic revision to be applied.
- New note backfill is idempotent (skips duplicates) and clears `nma_location_notes`.

**How to Test**
1. `source .venv/bin/activate`
2. `pytest tests/test_thing_transfer.py tests/test_data_migrations.py tests/test_data_migrations_cli.py tests/test_cli_commands.py`

**Follow-ups**
- Decide on a canonical note_type for location legacy notes if “General” isn’t desired.
- Add additional data migrations to registry as needed.
- Add future migration to drop Location.nma_location_notes

### DIFF
--- a/data_migrations/migrations/_template.py
+++ b/data_migrations/migrations/_template.py
@@ -1,0 +1,38 @@
+# ===============================================================================
+# Copyright 2026 ross
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===============================================================================
+from sqlalchemy.orm import Session
+
+from data_migrations.base import DataMigration
+
+
+def run(session: Session) -> None:
+    """
+    Implement migration logic here.
+
+    Use SQLAlchemy core for large batches:
+      session.execute(insert(Model), rows)
+    """
+    return None
+
+
+MIGRATION = DataMigration(
+    id="YYYYMMDD_0000",
+    alembic_revision="REVISION_ID",
+    name="Short migration name",
+    description="Why this data migration exists.",
+    run=run,
+    is_repeatable=False,
+)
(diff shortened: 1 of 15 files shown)


data_migrations/migrations/_template.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+MIGRATION = DataMigration(
+    id="YYYYMMDD_0000",
+    alembic_revision="REVISION_ID",
+    name="Short migration name",
+    description="Why this data migration exists.",
+    run=run,
+    is_repeatable=False,
+)
```
jacob-a-brown, line 38: I like the template a lot! It'll help moving forward. Is `alembic_revision` the ID of the alembic head script at the moment the migration is defined? So that we can identify at what stage in the DB's life the migration was made

jirhiker: yep thats it exactly. https://github.com/DataIntegrationGroup/OcotilloAPI/blob/a829322854a25e83fd75dfc56e5c84f5dd89f096/data_migrations/migrations/20260205_0001_move_nma_location_notes.py#L89

jacob-a-brown (commented):
Overall I think it looks good! And like the data migration tracking. If `nma_location_notes` is being moved to the polymorphic `Notes` table should that field be removed from the `Location` table?

jirhiker:
Nma_location_notes will be removed in a subsequent alembic migration

jacob-a-brown (approved):

PR #247, opened 2025-11-14
🚨 ⚠️ 💀 [NO TICKET]: Well inventory csv Feature Tests

jirhiker:
🚨 ⚠️ 💀 DO NOT MERGE UNTIL VERIFYING DB SCHEMA IS COMPATIBLE WITH THIS PR

### **Why**
_This PR addresses the following problem / context:_
- Use bullet points here

### **How**
_Implementation summary - the following was changed / added / removed:_
- Use bullet points here

### **Notes**
_Any special considerations, workarounds, or follow-up work to note?_
- Use bullet points here

PR #466, opened 2026-02-05
BDMS 412/529/530: water level csv implementation

jacob-a-brown:
### **Why**
- The `water level csv` upload needs to be implemented correctly per the feature tests and DB design
- Common steps between `water-level-csv.feature` and `well-inventory-csv.feature` need to be resolved so that the same step functions can be reused 

### **How**

#### **Implementation**
- The `water level csv` upload implementation was refactored to leverage Pydantic for data validations and code reuse
- The different fields in the `water level csv` file were mapped to the correct tables and fields in Ocotillo

#### **Tests**
- The common steps between `water-level-csv.feature` and `well-inventory-csv.feature` were moved to `tests/features/common.py` so that they can be reused for both feature files.
  - This also meant renaming `context.csv_rows` to `context.rows` in `water-level-csv.feature` for the style and implementation to be the same between the two feature files
- A unit test was added to `tests/services/water_level_csv.py` to ensure that the function used to add bulk data to the DB works and persists data as expected
- The cli tests were updated for the updated implementation

### **Notes**
- Once this PR has been accepted the final thing to do is implement optional water levels for the `well inventory csv` upload. This will be achievable by reusing the function `services/water_level_csv.py::bulk_upload_water_levels` in the `well inventory csv` upload. That function requires the wells to already exist in the DB, so it should be called _after_ the wells have been persisted. Some things to consider as this final functionality is implemented/developed:
  - Even if the`bulk_upload_water_levels` function cannot be used for `well inventory csv` the Pydantic schemas can and should be used.
  - If a well is uploaded correctly, but there is an error in its associated (optional) water level, should that well still be uploaded? What will the error message(s) look like?
  - If everything is successfully uploaded, what will the stdout look like?
  - The step test `the optional "water_level_date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T10:30:00") when provided` in `tests/features/well-inventory-csv.py` needs to be uncommented when this is implemented. 
  - The well inventory unit tests need to be updated to ensure that the optional water level upload is working as expected. Correct data validation, correct data persistence, etc...
- This uses some functionality from `well-inventory-csv` that has been vetted and approved. PR #247 should get merged into `staging` before `water-level-csv` gets merged into `staging` (note: this is to merge into `water-level-csv`, which also has the latest work from `well-inventory-csv`)


jacob-a-brown:
@codex review

api/observation.py
```diff
@@ -90,8 +90,8 @@ async def add_water_chemistry_observation(
 
 @router.post(
     "/groundwater-level/bulk-upload",
-    response_model=WaterLevelBulkUploadResponse,
-    status_code=HTTP_200_OK,
+    response_model=WaterLevelBulkUploadPayload,
+    status_code=HTTP_201_CREATED,
```
TylerAdamMartinez, line 94: 😄

tests/conftest.py
```diff
@@ -1178,3 +1178,24 @@ def second_lexicon_triple(third_lexicon_term, fourth_lexicon_term):
...
+        "measuring_person": contact.name,
+        "sample_method": "Electric tape measurement (E-probe)",
+        "mp_height": "1.5",
+        "level_status": "Water level not affected",
+        "depth_to_water_ft": "7",
+        "data_quality": "Water level accurate to within two hundreths of a foot",
+        "water_level_notes": "Initial measurement",
+    }
+    yield data
+    del data
```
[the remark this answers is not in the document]
jacob-a-brown: @jirhiker and @ksmuczynski 

This typo exists in `LU_DataQuality`. Should it be fixed here or at a late time?

jacob-a-brown:
@codex review

tests/features/steps/well-inventory-csv.py
```diff
@@ -67,23 +46,38 @@ def step_impl(context: Context):
...
+    if (
+        not hasattr(context, "datetime_fields")
+        or "date_time" not in context.datetime_fields
+    ):
+        context.datetime_fields = ["date_time"]
+    else:
+        context.datetime_fields.append("date_time")
+
 
+# TODO: implement when optional water levels are added to the well inventory csv testing data
```
TylerAdamMartinez, line 58: Why is this test here before it's implemented? It's kinda silly to have a test with just pass.

tests/features/data/water-levels.csv
```diff
@@ -1,3 +1,3 @@
 field_staff,well_name_point_id,field_event_date_time,measurement_date_time,sampler,sample_method,mp_height,level_status,depth_to_water_ft,data_quality,water_level_notes
-Alice Lopez,AR0001,2025-02-15T08:00:00-07:00,2025-02-15T10:30:00-07:00,Groundwater Team,electric tape,1.5,stable,45.2,approved,Initial measurement after irrigation shutdown
-Bernardo Chen,AR0002,2025-03-05T09:15:00-07:00,2025-03-05T11:10:00-07:00,Consultant,steel tape,1.8,rising,47.0,provisional,Follow-up visit; pump was off for 24h
+Alice Lopez,AR0001,2025-02-15T08:00:00,2025-02-15T10:30:00,Groundwater Team,electric tape,1.5,stable,45.2,approved,Initial measurement after irrigation shutdown
```
TylerAdamMartinez, line 2: So, the CSVs that we use cannot have the UTC offset, or this feature will not work?

tests/features/steps/well-inventory-csv.py
jacob-a-brown: Good question! Since it is in the feature file if I don't implement the step then a warning/error will be flagged and the tests will not pass. 

As I wrote at the end of the PR description this is the next and final step in getting the csv uploads to be fully implemented

tests/features/data/water-levels.csv
jacob-a-brown: The intension is for users to submit timezone-naive and then we add the offset manually, either MST or MDT (assuming that the data is submitted in Mountain Time)

1. [And "water_level_date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T10:30:00")](https://github.com/DataIntegrationGroup/OcotilloAPI/blob/7debe732146fa08957eb53e89f35705eab9c8854/tests/features/water-level-csv.feature#L38)
2. [Then all datetime objects are assigned the correct Mountain Time timezone offset based on the date value.](https://github.com/DataIntegrationGroup/OcotilloAPI/blob/7debe732146fa08957eb53e89f35705eab9c8854/tests/features/water-level-csv.feature#L49)

api/observation.py
jacob-a-brown: The naming can change as we see fit 😅 . I tried to mimic the nomenclature from when I refactored/updated the `water level csv` implementation. I think originally we were planning to have it work as an API endpoint but now it's being implemented by the CLI, whose "response" we need to define manually.

tests/conftest.py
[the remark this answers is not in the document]
jirhiker: @jacob-a-brown open an github issue to record that this needs to be fixed (at a later time)

core/lexicon.json
```diff
@@ -1,8274 +1,1180 @@
...
-      "categories": [
-        "qc_type"
-      ],
-      "term": "Field Blank",
-      "definition": "A sample of certified pure water that is taken to the field, opened, and processed through the same sampling procedure as a normal sample (e.g., poured into a sample bottle)."
-    },
-    {
-      "categories": [
-        "qc_type",
-        "sample_type"
```
ksmuczynski, line 294: I know this is quite old, but why did we remove the `groundwater` matrix? While both are technically "water," laboratories treat groundwater as a distinct matrix to ensure accurate analytical results. `Groundwater` may not have been used in the legacy system, but I think we will want to use in Ocotillo.

tests/conftest.py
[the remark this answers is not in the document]
jacob-a-brown: Issue created: https://github.com/DataIntegrationGroup/OcotilloAPI/issues/473

ksmuczynski (approved):
One minor comment, otherwise looks good to me!

core/lexicon.json
jacob-a-brown: I think because we said they're both the same thing, but `groundwater` is just a descriptor of where the water came from. Should we update here or make an issue to re-implement and update later?

jacob-a-brown: It looks like both can still be matrices. Want me to revert the changes in **core/parameter.json**?

jacob-a-brown: the change was reverted to have the matrix be `groundwater`

PR #470, opened 2026-02-05
BDMS-540: Link NMA_WeatherPhotos to NMA_WeatherData

ksmuczynski:
### **Why**
- `NMA_WeatherPhotos` lacked a direct FK relationship to `WeatherData`, making joins and integrity checks harder.

### **How**
- Adds a `WeatherID` FK relationship between `NMA_WeatherPhotos` and `NMA_WeatherData`.
- Updates the `WeatherPhoto` transfer to use the new relationship and aligns transfer helpers.
- Extends legacy `WeatherPhoto` tests to cover the `WeatherID` relationship.
- Adds the Alembic migration to create the FK and supporting uniqueness constraint.

### **Notes**
**- !! This is the last PR in the chain of "BDMS-540:..." PRs (467-469).**

### DIFF
--- a/alembic/versions/a1b2c3d4e5f6_link_surface_water_photos_to_surface_water_data.py
+++ b/alembic/versions/a1b2c3d4e5f6_link_surface_water_photos_to_surface_water_data.py
@@ -33,8 +33,7 @@ def upgrade() -> None:
         ["SurfaceID"],
         ondelete="CASCADE",
     )
-    op.execute(
-        """
+    op.execute("""
         DELETE FROM "NMA_SurfaceWaterPhotos" p
         WHERE p."SurfaceID" IS NULL
            OR NOT EXISTS (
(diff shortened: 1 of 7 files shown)


jirhiker:
My hope was that these types of changes would be taken care of in the db refactor and data transformation step that is to follow the 1:1 migration. Given we are so close to having migrated all the data I'd like to limit the modifications to the transfer script

ksmuczynski:
Okay, I will revisit this on Monday.

ksmuczynski:
Do you suggest a rollback on all changes?

jirhiker:
no, I think the only changes that potentially need to be rolled back are the changes to the transfer scripts.  I think the outstanding question we all need to consider "is all the necessary data migrated from nma to ocotillo, for all models."  The changes this PR makes seem like they could be accomplished without having to run a fresh transfer (?). 

Basically, I want to move away from the "wholesale transfer model" and move into the "DB refactor and data transformation model" introduced in PR #464   

As I write this tho I remember that another transfer is required to get some previously orphaned chemistry data migrated.  so Im going to approve this

PR #478, opened 2026-02-09
refactor: consolidate transferer classes to use ChemistryTransferer a…

jirhiker:
…nd remove unused caching logic

### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here


transfers/minor_trace_chemistry_transfer.py
marissafichera, line 1: can you add the SamplePointID to this transfer?

marissafichera (approved):
approving but it'd be great if you went ahead and transferred the SamplePointID from the minor and trace chemistry table. can happen later though

PR #533, opened 2026-02-17
fix: move measuring_point_height_is_assumed from thing to measuring_p…

jirhiker:
…oint_history

### **Why**
- Use bullet points here

### **How**
- Use bullet points here

### **Notes**
- Use bullet points here

### DIFF
--- a/alembic/versions/c4d5e6f7a8b9_add_measuring_point_height_is_assumed.py
+++ b/alembic/versions/c4d5e6f7a8b9_add_measuring_point_height_is_assumed.py
@@ -1,4 +1,4 @@
-"""Move measuring_point_height_is_assumed from thing to measuring_point_history.
+"""Add measuring_point_height_is_assumed to measuring_point_history.
 
 Revision ID: c4d5e6f7a8b9
 Revises: e71807682f57
--- a/alembic/versions/c4d5e6f7a8b9_add_measuring_point_height_is_assumed.py
+++ b/alembic/versions/c4d5e6f7a8b9_add_measuring_point_height_is_assumed.py
@@ -1,0 +1,52 @@
+"""Move measuring_point_height_is_assumed from thing to measuring_point_history.
+
+Revision ID: c4d5e6f7a8b9
+Revises: e71807682f57
+Create Date: 2026-02-16 12:00:00.000000
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy import inspect
+
+# revision identifiers, used by Alembic.
+revision: str = "c4d5e6f7a8b9"
+down_revision: Union[str, Sequence[str], None] = "e71807682f57"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = inspect(bind)
+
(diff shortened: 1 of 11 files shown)


jirhiker:
@marissafichera @ksmuczynski

ksmuczynski:
@jirhiker There are two tables that capture measuring point height: `Observation` and `MeasuringPointHistory`.

The `MeasuringPointHistory` table is designed to be the source of truth. It stores official, surveyed heights from construction or re-survey events. Since updates to this table only happen during significant events (like a new wellhead installation), we want to force the user entering that data to provide the surveyed height immediately to ensure the record is complete. **A non-nullable `measuring_point_height` should be enforced here.**

The `Observation` table is designed to capture routine measurements taken during field events. This table is often messy and much more prone to human error. The `measuring_point_height` field here **SHOULD allow NULLs** because it allows the database to honestly reflect that a measurement was missed or overlooked.

A NULL in the `Observation` table isn't a dead end and does not mean depth to water (below ground surface) can't be calculated. We can use the `Observation` date to look up the authoritative height from the `MeasuringPointHistory` table and fill in the blanks during reporting.

My recommendation is to keep the `MeasuringPointHistory` table as-is and add an `observation_reason` field to the `Observation` table to document the reason for a missing value/measurement. I would make `observation_reason` a nullable, constrained lexicon term (the Notes field would be used to provide any additional context). Some example lexicon terms include: `Dry`, `Obstructed`, `Destroyed`, `Equipment Failure`, `Not Sampled/Measured`.

jirhiker:
@ksmuczynski I agree with everything except that any measuring_point_height should be nullable. defaulting the measuring_point_height to 0 and adding flags in both the Observation and MeasuringPointHistory reduces the complexity of the logic to calculate depth to water bgs.  Will you open a jira ticket for adding the observation_reason flag to Observation.

jirhiker:
@ksmuczynski I think for non legacy data and when implementing the Well Inventory/Water Level forms we should very much consider 

> " we want to force the user entering that data to provide the surveyed height immediately to ensure the record is complete. A non-nullable measuring_point_height should be enforced here."

jirhiker:
@ksmuczynski upon further consideration Im going to pass this off to you and let you move forward with your recommendation. I will close this PR as Ive started to introduce scope creep

ksmuczynski:
Sure thing, I'll get a Jira ticket opened for the observation_reason field and lexicon terms today.

Regarding the 0 default for the measuring point height: I’m hesitant to go that route because, in the field, a height of 0 actually describes a specific well type (flush mount). If we default to 0 during a bulk CSV upload, we'll lose the ability to distinguish between a flush mount well and a forgotten measurement. This risks introducing a small elevation error that would be invisible to anyone reviewing the data later.

To keep the logic simple for reporting without losing data integrity, what if we keep the database field (Observation.measuring_point_height) Nullable but handle the 'fallback' logic in a Calculated View or a SQLAlchemy hybrid property? This creates a 'virtual' non-nullable column for the app to query, and we don't have to deal with NULL math or complex joins in the code.

For our current CLI tool, we could add a Pydantic validator. The script will expect a height, but if that's missing, it will require an `observation_reason` to proceed. This fails fast during the import if the data is incomplete, which is safer than letting a default 0 slip through and silently corrupt our elevation models.

When we eventually move to web forms, we can have the UI auto-populate the height from the `MeasuringPointHistory` table. This keeps the raw data honest for the scientists, but the API and CLI stay 'easy' for developers. What do you think?

PR #545, opened 2026-02-24
Replace custom OGC API implementation with pygeoapi (mounted in FastAPI)

jirhiker:
## Summary
Replaced the custom OGC implementation with a pygeoapi-based OGC API Features stack, mounted into FastAPI, and moved pygeoapi support objects to DB-managed views/migrations.

### FastAPI + pygeoapi integration
- Mounted pygeoapi as a sub-application under `PYGEOAPI_MOUNT_PATH` (now `/ogcapi`).
- Moved pygeoapi server/resource configuration into a dedicated YAML template:
  - `core/pygeoapi-config.yml`
- Runtime config/openapi files are generated in a runtime-writable directory (`/tmp/pygeoapi` by default).
- Route registration pattern consolidated (no duplicate registration paths).

### OGC data model changes
- Removed custom hand-rolled OGC route behavior in favor of pygeoapi collections backed by SQL views/materialized views.
- Added/updated pygeoapi collections for thing types and derived well datasets.
- Set collection bbox extents to New Mexico.

### DB/Alembic
- Added pygeoapi supporting views in Alembic (`d5e6f7a8b9c0...`), including:
  - `ogc_*` thing-type views
  - `ogc_latest_depth_to_water_wells` (materialized)
  - `ogc_avg_tds_wells` (materialized; sourced from `NMA_MajorChemistry`)
- Removed fallback `WHERE FALSE` stub behavior; migration now fails clearly when required source tables are missing.
- Improved required-table error messages with explicit missing table names.
- Updated latest-depth logic:
  - `measuring_point_height` nulls are now treated as `0` (`COALESCE`) for depth-to-water BGS.
- Added pg_cron scheduling in the same migration:
  - creates refresh function
  - schedules nightly refresh job
  - unschedules/drops on downgrade
  - fixed DO-block quoting bug (`$do$` / `$cmd$`).

### CLI
- Added command to refresh pygeoapi materialized views on demand:
  - `refresh-pygeoapi-materialized-views`
  - supports `--view` and `--concurrently`
- CLI dotenv behavior changed to `load_dotenv(override=True)` so `.env` values win.

### Infra/workflows/env
- Added/standardized pygeoapi DB env vars in CI/CD/test workflows:
  - `PYGEOAPI_POSTGRES_HOST`
  - `PYGEOAPI_POSTGRES_PORT`
  - `PYGEOAPI_POSTGRES_DB`
  - `PYGEOAPI_POSTGRES_USER`
  - `PYGEOAPI_POSTGRES_PASSWORD`
- Added `pygeoapi` to default `app_read` grant membership.
- Docker DB setup updated for pg_cron support:
  - new `docker/db/Dockerfile` installs `postgresql-17-cron`
  - compose DB command preloads `pg_cron` and sets `cron.database_name`.

### Cleanup
- Removed custom OGC references/tests/routes where applicable while preserving pygeoapi-based OGC API behavior.
- Unstaged/ignored local `.pygeoapi` runtime artifacts from versioned code paths.

## Why
- Align OGC API implementation with a standard server (pygeoapi) instead of maintaining custom OGC logic.
- Make OGC collection generation and DB support deterministic via migrations.
- Improve operability with explicit refresh mechanisms (manual CLI + nightly pg_cron).
- Reduce routing/config drift and deployment surprises across local/CI/staging/prod.

## Notes
- `pg_cron` must be available on the target PostgreSQL server/instance for scheduled refresh setup.


core/pygeoapi-config.yml
```diff
@@ -0,0 +1,104 @@
...
+    url: https://geoinfo.nmt.edu
+  contact:
+    name: API Support
+    email: support@example.com
+
+resources:
+  locations:
+    type: collection
+    title: Locations
+    description: Sample locations
```
jacob-a-brown: Samples aren't taken at all locations. Since using `thing` can be confusing, should this be changed to something liked `Surveyed locations` or `Inventory locations`?

core/pygeoapi-config.yml
```diff
@@ -0,0 +1,104 @@
...
+          search_path: [public]
+        id_field: id
+        table: location
+        geom_field: point
+
+  wells:
+    type: collection
+    title: Wells
+    description: Things filtered to water wells
+    keywords: [wells]
```
jacob-a-brown: Since these are `water wells` should this be 

```
water-wells:
    type: collection
    title: Water Wells
    description: Things filtered to water wells
    keywords: [wells, water-wells]
```

?

core/pygeoapi.py
```diff
@@ -0,0 +1,197 @@
...
+
+def _project_root() -> Path:
+    return Path(__file__).resolve().parent.parent
+
+
+def _template_path() -> Path:
+    return Path(__file__).resolve().parent / "pygeoapi-config.yml"
+
+
+def _mount_path() -> str:
```
jacob-a-brown, line 171: This can use and return pathlib's Path instead of strings

jacob-a-brown: Unless a string is required for when it is invoked (in that case, if you want to use Path you can `str(<Path object>)`)

jirhiker: _mount_path is the URL path, not a file path, so a str is expected

alembic/versions/d5e6f7a8b9c0_create_pygeoapi_supporting_views.py
```diff
@@ -0,0 +1,261 @@
...
+from sqlalchemy import inspect, text
+
+# revision identifiers, used by Alembic.
+revision: str = "d5e6f7a8b9c0"
+down_revision: Union[str, Sequence[str], None] = "c4d5e6f7a8b9"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+THING_COLLECTIONS = [
+    ("wells", "water well"),
```
jacob-a-brown: Should this be `("water_wells", "water well")`? Subsequent tuples have the well descriptor prepended in the first item of the tuple

alembic/versions/d5e6f7a8b9c0_create_pygeoapi_supporting_views.py
```diff
@@ -0,0 +1,261 @@
...
+        CREATE VIEW ogc_{view_id} AS
+        WITH latest_location AS (
+            SELECT DISTINCT ON (lta.thing_id)
+                lta.thing_id,
+                lta.location_id,
+                lta.effective_start
+            FROM location_thing_association AS lta
+            WHERE lta.effective_end IS NULL
+            ORDER BY lta.thing_id, lta.effective_start DESC
+        )
```
jacob-a-brown, line 68: This SQL statement to get the `latest_location` is used in a number of functions. It may be worth putting into a constant at the beginning of the script and inserting it into the f-strings throughout.

alembic/versions/d5e6f7a8b9c0_create_pygeoapi_supporting_views.py
```diff
@@ -0,0 +1,261 @@
...
+            JOIN thing AS t ON t.id = csi.thing_id
+            WHERE
+                t.thing_type = 'water well'
+                AND mc."SampleValue" IS NOT NULL
+                AND (
+                    lower(coalesce(mc."Analyte", '')) IN (
+                        'tds',
+                        'total dissolved solids'
+                    )
+                    OR lower(coalesce(mc."Symbol", '')) = 'tds'
```
jacob-a-brown, line 180: Are we storing total dissolved solids in two different ways? If so, @ksmuczynski, I think that this should be standardized. Would I/we make a GitHub issue for this?

alembic/versions/d5e6f7a8b9c0_create_pygeoapi_supporting_views.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    inspector = inspect(bind)
+
+    required_core = {"thing", "location", "location_thing_association"}
+    if not required_core.issubset(set(inspector.get_table_names(schema="public"))):
+        raise RuntimeError(
+            "Cannot create pygeoapi supporting views: required core tables are missing"
```
jacob-a-brown: It may be worth enumerating the required core tables in this error message so that people don't have to hunt through the code to determine what they are:

```python
required_core = {"thing", "location", "location_thing_association"}
    if not required_core.issubset(set(inspector.get_table_names(schema="public"))):
        missing_tables = [t for t in required_core if t not in set(inspector.get_table_names(schema="public"))]
        missing_tables_str = ", ".join(missing_tables)
        raise RuntimeError(
            "Cannot create pygeoapi supporting views. The following required core tables are missing: {missing_tables_str}"
        )
```

core/pygeoapi-config.yml
```diff
@@ -0,0 +1,106 @@
...
+          search_path: [public]
+        id_field: id
+        table: location
+        geom_field: point
+
+  latest_depth_to_water_wells:
+    type: collection
+    title: Latest Depth to Water (Wells)
+    description: Most recent depth-to-water observation for each water well.
+    keywords: [wells, groundwater-level, depth-to-water, latest]
```
jacob-a-brown: How about adding `depth-to-water-bgs` here since that's also included in the observation?

core/pygeoapi.py
```diff
@@ -0,0 +1,311 @@
+import os
+from importlib.util import find_spec
+from pathlib import Path
+
+from fastapi import FastAPI
+
+THING_COLLECTIONS = [
+    {
+        "id": "wells",
+        "title": "Wells",
```
jacob-a-brown: Should this be "water wells" since that is the type of well being queried?

core/pygeoapi.py
```diff
@@ -0,0 +1,311 @@
...
+        "thing_type": "water well",
+        "description": "Groundwater wells used for monitoring, production, and hydrogeologic investigations.",
+        "keywords": ["wells", "groundwater", "water-well"],
```
jacob-a-brown: Should all different types of wells, like abandoned and artesian, also have "wells" as a keyword?

PR #548, opened 2026-02-26
feat(tests): relax validation rules and expand enum coverage in well-inventory-csv feature

marissafichera:
```
BDMS-608: Relax well inventory CSV validation rules and expand enum coverage
```
@BDMS-608
---

### **Why**
- Validation rules in the feature file were stricter than necessary, causing real user-submitted CSVs to fail ingestion
- Required fields did not reflect actual field usage in the field — elevation and measuring point data is not always available at time of entry
- Contact and water level rules were all-or-nothing, blocking valid partial submissions
- Several enum fields (address_type, state, well_hole_status, monitoring_status, well_pump_type) lacked explicit validation coverage

### **How**
- Moved `site_name`, `elevation_ft`, `elevation_method`, and `measuring_point_height_ft` from required to optional
- Replaced "both contact_name and contact_organization required" rule with "at least one of the two required"
- Removed all-or-nothing water level validation; `water_level_date_time` is now required only when `depth_to_water_ft` is provided, all other water level fields are independent and optional
- Added negative scenarios for invalid `address_type`, state abbreviation, `well_hole_status`, `monitoring_status`, and `well_pump_type` with allowed values specified in each Given step
- Added `well_notes`, `well_measuring_notes`, `water_notes`, `historical_notes`, `well_hole_status`, and `monitoring_status` to optional fields

### **Notes**
- Corresponding changes to the Pydantic schema (`WellInventoryRow`) are required for these rules to take effect in the backend — this PR covers the feature spec only
- Allowed values for new enum scenarios are based on the current lexicon and should be kept in sync if the lexicon changes

### DIFF
--- a/tests/features/water-level-csv.feature
+++ b/tests/features/water-level-csv.feature
@@ -27,19 +27,20 @@ Feature: Bulk upload water level entries from CSV via CLI
       | field_staff           |
       | well_name_point_id    |
       | field_event_date_time |
-      | water_level_date_time |
       | measuring_person      |
-      | sample_method         |
-      | mp_height             |
-      | level_status          |
-      | depth_to_water_ft     |
-      | data_quality          |
     And each "well_name_point_id" value matches an existing well
+    And "field_event_date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T08:00:00")
     And "water_level_date_time" values are valid ISO 8601 timezone-naive datetime strings (e.g. "2025-02-15T10:30:00")
     And the CSV includes optional fields when available:
       | optional field name |
       | field_staff_2       |
       | field_staff_3       |
+      | water_level_date_time |
+      | sample_method         |
+      | mp_height             |
+      | level_status          |
+      | depth_to_water_ft     |
+      | data_quality          |
       | water_level_notes   |
     When I run the CLI command:
       """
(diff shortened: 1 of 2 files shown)


ksmuczynski:
@marissafichera Two comments below related to potential future work, but I agree overall with the updates. Approved!

Regarding point no. 1 in the HOW section of the PR:
>1. Moved `site_name`, `elevation_ft`, `elevation_method`, and `measuring_point_height_ft` from required to optional

- `site_name`: Yes
- `elevation_ft`: Yes **for now**. 
	- The Ocotillo codebase includes functionality to retrieve elevation data from the USGS National Map Elevation Point Query Service (EPQS). While this feature exists, it is **not currently used** during the ingestion of the well inventory CSV.
	- The ingestion logic uses the user-provided values and does not automatically query or verify them against the USGS DEM.
	- In the interest of getting new data into Ocotillo, let's relax the requirement. In the future, we can implement functionality to retrieve elevation data from the USGS National Map Elevation Point Query Service (EPQS), and assign `elevation_method` appropriately.
- `elevation_method`: Yes **for now**
	- In the interest of getting new data into Ocotillo, let's relax the requirement. In the future, if we implement functionality to retrieve elevation data from the USGS National Map Elevation Point Query Service (EPQS), we can have this be assigned automatically
- `measuring_point_height_ft`: Yes

ksmuczynski (approved):

jirhiker (changes requested):
Looks good. Remove instances of "  And the system should return a response in JSON format" and this should be ready to proceed to implementation.  

Since this updates the spec without updating the implementation, there are (at least) two options for keeping the tests passing on staging. 1) do not merge into staging.  Whoever is working on the implementation should create a new branch from `marissa`, make the tests pass, then open a PR to staging 2) remove the `@production` tag from this file so that CI skips this test. once implementation is complete `@production` tag can be re-added

PR #557, opened 2026-02-27
BDMS-610: Add `Windmill` to lexicon `well_pump_type` category

ksmuczynski:
### **Why**
- Windmill is a common pump type in NM and should be part of the `well_pump_type` lexicon.

### **How**
- Added `Windmill` as a `well_pump_type` in the lexicon.

### **Notes**
- None

### DIFF
--- a/core/lexicon.json
+++ b/core/lexicon.json
@@ -8243,6 +8243,13 @@
     },
     {
       "categories": [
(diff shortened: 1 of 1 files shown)


marissafichera (approved):
*insert dancing baby here*

PR #583, opened 2026-03-04
`test(features): enforce lexicon values for water-level descriptor fields

marissafichera:
### **Why**
- The `water-level-csv.feature` still allowed free-text behavior where `sample_method`, `level_status`, and `data_quality` should be enum/lexicon-constrained.
- We need feature coverage that explicitly requires lexicon-valid values for those three fields.

### **How**
- Updated `tests/features/water-level-csv.feature` only.
- Added a happy-path requirement that, when provided, `sample_method`, `level_status`, and `data_quality` must be valid lexicon values.
- Replaced the free-text acceptance scenario with a negative validation scenario that fails upload when those fields contain invalid lexicon values.
- Kept expectations consistent with existing validation style:
  - non-zero exit code
  - validation errors identify row/field
  - no rows imported

### **Notes**
- This is a spec-only PR (no Python implementation changes).
- If current step definitions still model the old free-text behavior, they may need a small follow-up to match this updated feature.


ksmuczynski (approved):
Looks great, approved!

PR #596, opened 2026-03-11
BDMS-626: Improve validation, schema alignment, and well inventory handling

ksmuczynski:
### **Why**
- Database mapping was incomplete: `public_availability_acknowledgement`, `monitoring_status`, `well/water notes`, and water level observations were not being persisted correctly.
- Several schema fields were overly strict, requiring values that should be optional, using plain strings where lexicon enums were expected, and rejecting valid CSVs too aggressively.
- WellInventoryRow schema now integrates directly with lexicon-based enums to improve consistency between CSV ingestion and database persistence.
- Real-world imports exposed database failures that were hard to diagnose, especially around contact organization foreign keys and empty-string values being written into lexicon-backed columns.
- Real-world source files also included `monitoring_frequency` values like `Complete` that do not map cleanly to the current lexicon and needed importer-side normalization.
- Valid CSVs saved with a UTF-8 BOM could be misread at the header level, causing required fields like project to appear missing.
- BDD tests suffered from primary key conflicts and outdated test data that no longer aligned with the importer’s validation and best-effort behavior.
- Previously, a single malformed row in a CSV would abort the entire import process. Users need "best-effort" logic where valid data is saved while invalid rows are flagged. 

### **How**
- **Database mapping**:
    - `public_availability_acknowledgement` now maps to `Location.release_status` (True → public, False → private, unset → draft).
    - `monitoring_status` is now written to the `StatusHistory` table.
    - `well_notes` and `water_notes` are now stored as polymorphic notes on the `Thing`.
    - Providing water-level data now creates the expected Sample and Observation records and attaches them to the correct field activity.
    - Water-level observation field names were aligned with the actual database model.

- **`WellInventoryRow` schema:** 
    - Made `site_name`, `elevation_ft`, `elevation_method`, `monitoring_point_height_ft`, and `depth_to_water_ft` optional.
    - Replaced plain `str` types with lexicon-based enums for `elevation_method`, `depth_source`, `well_pump_type`, `monitoring_status`, `sample_method`, and `data_quality`.
    - Added a `flexible_lexicon_validator` for case-insensitive, whitespace-tolerant matching.
    - Relaxed contact validation to require only one of `contact_name` or `contact_organization`.
    - Made water_level_date_time required only when depth_to_water_ft is provided.
    - Normalized blank values like depth to water, contact organization, and well status to `None` instead of persisting invalid empty strings.
    - Added regression coverage for blank depth_to_water_ft and blank lexicon-backed contact/status fields.

- **Best-effort logic:**
    - Implemented row-level atomic savepoints so individual row failures no longer abort the full import. 
    - Failed rows are logged with 1-based row numbers and well IDs in a `validation_errors` list.
    - Fixed an `UnboundLocalError` caused by auto-generated well IDs.
    - Updated error reporting to surface actual database-level failures rather than collapsing them to a generic database error.
    - Added `commit=False` support so `services/thing_helper.py` and `services/contact_helper.py` participate correctly in the outer best-effort transaction without prematurely committing.
    - Made reruns idempotent by detecting previously imported rows and skipping duplicate record creation instead of failing on unique constraints.

- **Lexicon and import compatibility:**
    - Added missing organization terms required by real-user imports so contact organization foreign keys can resolve correctly.
    - Updated monitoring_status enum construction to use the correct lexicon category.
    - Added BOM-safe CSV decoding so UTF-8 BOM headers are parsed correctly during CLI import.
    - Normalized `monitoring_frequency = Complete` source values so they do not create a monitoring frequency record and instead set `monitoring_status` to `Not currently monitored.`
  - Added unit and BDD coverage for the Complete monitoring frequency normalization behavior.

- **BDD test suite:**
    - Updated test CSVs in `tests/features/data/` to use valid lexicon terms and properly quoted comma-containing values.
    - Added scenario-based unique ID suffixing in `Given` steps to isolate tests and prevent primary key conflicts.
    - Updated `well-inventory-csv.feature` to assert partial success (e.g., "1 well is imported") in negative scenarios where best-effort behavior is expected.
    - Aligned validation expectations with the importer’s current detailed database error reporting.
    -   All 44 well inventory BDD scenarios now pass when the test database is rebuilt against the current branch state.

### **Notes**
- Further enhancements may be required for complete schema validation coverage. 
- Additional real-user CSV validation coverage may be scoped in a follow-up PR focused on ingestion of real user-entered data.
- Lexicon matching is intentionally lenient at ingestion time (case-insensitive, whitespace-stripped), but values must still resolve to a known lexicon term before persistence.
-  Because contact organizations remain lexicon-backed, new real-world organizations still require lexicon seeding in the target database before imports will succeed.
- Test isolation is achieved via per-scenario ID suffixing in the BDD setup; if the shared test database strategy changes in the future, this workaround may no longer be necessary.
- BDD runs on this branch may require rebuilding `ocotilloapi_test` after schema or lexicon changes to avoid stale test database state.


tests/features/steps/well-inventory-csv-given.py
```diff
@@ -29,17 +29,41 @@ def _set_file_content(context: Context, name):
...
+    context.file_name = name or path.name
+
+    if path.suffix == ".csv" and path.exists() and path.stat().st_size > 0:
+        suffix = hashlib.md5(context.scenario.name.encode()).hexdigest()[:6]
+        df = pd.read_csv(path, dtype=str)
+        if "well_name_point_id" in df.columns:
+            df["well_name_point_id"] = df["well_name_point_id"].apply(
+                lambda x: (
+                    f"{x}_{suffix}"
+                    if x and not str(x).endswith("-xxxx") and not str(x).strip() == ""
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved.

tests/features/steps/well-inventory-csv-validation-error.py
```diff
@@ -21,14 +21,18 @@
...
+        found = False
+        for actual in validation_errors:
+            field_match = str(expected.get("field", "")) in str(actual.get("field", ""))
+            error_match = str(expected.get("error", "")) in str(actual.get("error", ""))
+            if field_match and error_match:
+                found = True
+                break
+        assert (
+            found
+        ), f"Expected validation error for field '{expected.get('field')}' with error containing '{expected.get('error')}' not found. Got: {validation_errors}"
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved.

tests/features/steps/well-inventory-csv-validation-error.py
```diff
@@ -280,26 +284,32 @@ def step_then_response_includes_invalid_well_pump_type_error(context: Context):
...
-            or "contact_1_organization is required" in str(err.get("error", ""))
+            "At least one of contact_1_name or contact_1_organization must be provided"
+            in str(err.get("error", ""))
         )
         for err in validation_errors
     )
+    if not found:
+        pass
+        # print(f"ACTUAL VALIDATION ERRORS: {validation_errors}")
+
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved.

schemas/well_inventory.py
```diff
@@ -122,28 +127,64 @@ def email_validator_function(email_str):
...
+MonitoringStatusField: TypeAlias = Annotated[
+    Optional[MonitoringStatus],
+    BeforeValidator(flexible_lexicon_validator(MonitoringStatus)),
+]
+SampleMethodField: TypeAlias = Annotated[
+    Optional[OriginType], BeforeValidator(flexible_lexicon_validator(OriginType))
+]
+DataQualityField: TypeAlias = Annotated[
+    Optional[OriginType], BeforeValidator(flexible_lexicon_validator(OriginType))
 ]
```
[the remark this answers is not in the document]
ksmuczynski: Don't know how I missed this. Addressed and resolved.

schemas/well_inventory.py
```diff
@@ -325,38 +367,48 @@ def validate_model(self):
...
                     raise ValueError(
-                        f"{key}_name is required when other contact fields are provided"
+                        f"{key}_role is required when contact fields are provided"
                     )
-                if not organization:
+                if not getattr(self, f"{key}_type"):
                     raise ValueError(
-                        f"{key}_organization is required when other contact fields are provided"
+                        f"{key}_type is required when contact fields are provided"
                     )
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved.

services/well_inventory_csv.py
```diff
@@ -264,12 +290,21 @@ def _make_location(model) -> Location:
     transformed_point = transform_srid(
         point, source_srid=source_srid, target_srid=SRID_WGS84
     )
-    elevation_ft = float(model.elevation_ft)
-    elevation_m = convert_ft_to_m(elevation_ft)
+    elevation_ft = model.elevation_ft
+    elevation_m = (
+        convert_ft_to_m(float(elevation_ft)) if elevation_ft is not None else 0.0
+    )
```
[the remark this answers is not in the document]
ksmuczynski: I agree this may not be the ideal long-term behavior, and that the better design may be to either persist NULL or fail the row explicitly. However, that change is beyond the scope of this PR because it would require revisiting Location.elevation nullability, related serializers, geospatial outputs, and derived products that currently assume a numeric elevation.

For this PR, I'm going to keep the existing 0.0 fallback so that inventory imports can proceed without expanding the scope into a broader location-model change. These cases should be easy to identify and correct manually if needed until we evaluate nullable location elevation and downstream impacts end-to-end.

services/well_inventory_csv.py
```diff
@@ -52,8 +55,10 @@
 from services.util import transform_srid, convert_ft_to_m
 
 AUTOGEN_DEFAULT_PREFIX = "NM-"
-AUTOGEN_PREFIX_REGEX = re.compile(r"^[A-Z]{2,3}-$")
-AUTOGEN_TOKEN_REGEX = re.compile(r"^(?P<prefix>[A-Z]{2,3})\s*-\s*(?:x{4}|X{4})$")
+AUTOGEN_PREFIX_REGEX = re.compile(r"^[A-Z]{2,3}-$", re.IGNORECASE)
+AUTOGEN_TOKEN_REGEX = re.compile(
+    r"^(?P<prefix>[A-Z]{2,3})\s*-\s*(?:x{4}|X{4})$", re.IGNORECASE
```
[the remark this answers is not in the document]
ksmuczynski: Test expectations have been updated. Resolved.

tests/test_well_inventory.py
```diff
@@ -907,6 +924,50 @@ def test_group_query_with_multiple_conditions(self):
...
+            {
+                "measuring_person": "Tech 1",
+                "sample_method": "Tape",
+                "water_level_date_time": "2025-02-15T10:30:00",
+                "mp_height_ft": 2.5,
+                "level_status": "Static",
+                "depth_to_water_ft": 11.2,
+                "data_quality": "Good",
+                "water_level_notes": "Initial reading",
+            }
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved.

ksmuczynski:
@jirhiker Recent updates to the CLI command tests are failing on my branch. Since these changes don't seem to impact the well inventory ingestion, should I resolve the test failures within this PR, or would you prefer I handle them in a separate one before re-opening for review?

alembic/versions/p9c1d2e3f4a5_make_contact_role_nullable.py
jacob-a-brown, line 1: I'm not very familiar with the actual data and how often a `contact` is added without a `role`. Are we relaxing our standards for a few edge cases, or is this common enough that role should be nullable?

alembic/versions/q0d1e2f3a4b5_make_contact_type_nullable.py
jacob-a-brown, line 1: Same question about a `contact`'s `role`: are we relaxing the standard for a few edge cases, or is the `contact_type` not known frequently enough that it warrants making the field nullable?

schemas/thing.py
jacob-a-brown, line 1: If `monitoring_status` is added in `CreateWell` then I think that it should also be a field in `WellResponse`

schemas/thing.py
```diff
@@ -143,6 +143,7 @@ class CreateWell(CreateBaseThing, ValidateWell):
     is_suitable_for_datalogger: bool | None = None
     is_open: bool | None = None
     well_status: str | None = None
+    monitoring_status: str | None = None
```
jacob-a-brown, line 146: This should be restricted to `MonitoringStatus` enum values

schemas/well_inventory.py
```diff
@@ -325,38 +363,41 @@ def validate_model(self):
                 f" Zone={self.utm_zone}"
             )
 
+        if self.depth_to_water_ft is not None:
```
jacob-a-brown, line 397: There should be a check here for `mp_height`, too, since it should be required for every observation.

services/contact_helper.py
```diff
@@ -114,16 +114,16 @@ def add_contact(
...
 
-        session.refresh(contact)
-
-        for note in contact.notes:
-            session.refresh(note)
-
     except Exception as e:
-        session.rollback()
+        if commit:
+            session.rollback()
```
jacob-a-brown, line 126: I think that the session should always be rolled back on an error, otherwise erroneous data may be accidentally added to the DB

services/thing_helper.py
```diff
@@ -417,15 +430,16 @@ def add_thing(
...
             session.flush()
-        session.refresh(thing)
-
-        for note in thing.notes:
-            session.refresh(note)
 
     except Exception as e:
-        session.rollback()
+        if commit:
+            session.rollback()
```
jacob-a-brown, line 442: I think that the session should always be rolled back on an error, otherwise erroneous data may be accidentally added to the DB

schemas/well_inventory.py
```diff
@@ -240,40 +280,55 @@ class WellInventoryRow(BaseModel):
...
 
     # water levels
-    sampler: Optional[str] = None
-    sample_method: Optional[str] = None
-    measurement_date_time: OptionalPastOrTodayDateTime = None
-    mp_height: Optional[float] = None
+    sampler: Optional[str] = Field(
+        default=None,
+        validation_alias=AliasChoices("sampler", "measuring_person"),
+    )
```
jacob-a-brown, line 330: `sampler` should be validated in `validate_model` to ensure that it is one of `field_staff`, `field_staff_2`, or `field_staff_3`. Since this is a custom validation it should be tested as either a unit test or added to the feature tests (I imagine the latter)

services/well_inventory_csv.py
```diff
@@ -235,7 +261,7 @@ def _import_well_inventory_csv(session: Session, text: str, user: str):
             "total_rows_imported": rows_imported,
             "validation_errors_or_warnings": rows_with_validation_errors_or_warnings,
         },
-        "wells": wells,
+        "wells": wells_imported,
```
jacob-a-brown, line 344: Without partial imports this should just evaluate to the variable `wells`

services/well_inventory_csv.py
```diff
@@ -159,69 +161,93 @@ def _import_well_inventory_csv(session: Session, text: str, user: str):
...
+                        )
+                        validation_errors.append(
+                            {
+                                "row": row_number,
+                                "well_id": current_row_id,
+                                "field": field,
+                                "error": error_text,
+                            }
+                        )
+            session.commit()
```
jacob-a-brown, line 311: Per [the comment from the initial implementation](https://github.com/DataIntegrationGroup/OcotilloAPI/pull/247#discussion_r2775297071) of the `well inventory csv` I think that partial imports should not be allowed. `session.commit()` should only be called at the end if there were no errors, otherwise `session.rollback()` should be called. Something like this after the for loop is finished.

```python
if validation_errors:
    wells = []
    rows_imported = 0
    session.rollback()
elif not validation_errors:
    rows_imported = len(wells_imported)
    session.commit()
else:
    # catchall if something goes awry during session.commit()
    wells = []
    rows_imported = 0
    session.rollback()
    return {"detail": "Something went wrong when committing the data to the DB"}
```

services/well_inventory_csv.py
```diff
@@ -264,12 +290,21 @@ def _make_location(model) -> Location:
     transformed_point = transform_srid(
         point, source_srid=source_srid, target_srid=SRID_WGS84
     )
-    elevation_ft = float(model.elevation_ft)
-    elevation_m = convert_ft_to_m(elevation_ft)
+    elevation_ft = model.elevation_ft
+    elevation_m = (
+        convert_ft_to_m(float(elevation_ft)) if elevation_ft is not None else 0.0
```
jacob-a-brown, line 375: If `elevation_ft` is `None` then I think that `elevation_m` should be set to `None`, otherwise we falsely set it to `0.0` which may lead to incorrect analyses

jacob-a-brown: I just saw your comment below in response to CoPilot's suggestion. Either we should change the schema to make `elevation_ft` non-nullable or we should change the db to make `elevation` nullable for a `Location`

services/well_inventory_csv.py
```diff
@@ -321,9 +357,17 @@ def _make_contact(model: WellInventoryRow, well: Thing, idx) -> dict:
...
             "thing_id": well.id,
             "name": name,
-            "organization": getattr(model, f"contact_{idx}_organization"),
-            "role": getattr(model, f"contact_{idx}_role"),
-            "contact_type": getattr(model, f"contact_{idx}_type"),
+            "organization": organization,
+            "role": (
+                getattr(model, f"contact_{idx}_role").value
+                if hasattr(getattr(model, f"contact_{idx}_role"), "value")
+                else getattr(model, f"contact_{idx}_role")
```
jacob-a-brown: If `role` is now nullable then I think that the else portion of this statement should be `else None`:

```python
"role": getattr(model, f"contact_{idx}_role").value if hasattr(getattr(model, f"contact_{idx}_role"), "value") else None
```

services/well_inventory_csv.py
```diff
@@ -321,9 +357,17 @@ def _make_contact(model: WellInventoryRow, well: Thing, idx) -> dict:
...
+            ),
+            "contact_type": (
+                getattr(model, f"contact_{idx}_type").value
+                if hasattr(getattr(model, f"contact_{idx}_type"), "value")
+                else getattr(model, f"contact_{idx}_type")
+            ),
```
jacob-a-brown: If `contact_type` is now nullable then I think that the else portion of this statement should be `else None`

```python
"role": getattr(model, f"contact_{idx}_role").value if hasattr(getattr(model, f"contact_{idx}_role"), "value") else None
```

services/well_inventory_csv.py
```diff
@@ -539,7 +602,17 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user)
...
+        (model.water_notes, "Water"),
         (historic_depth_note, "Historical"),
+        (
+            (
+                f"Sample possible: {model.sample_possible}"
+                if model.sample_possible is not None
+                else None
+            ),
+            "Sampling Procedure",
+        ),
```
jacob-a-brown, line 740: Should this be a status history rather than a note?

services/well_inventory_csv.py
```diff
@@ -661,6 +739,62 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user)
...
+        if not parameter:
+            # this shouldn't happen if initialized properly, but just in case
+            parameter = Parameter(
+                parameter_name="groundwater level",
+                matrix="groundwater",
+                parameter_type="Field Parameter",
+                default_unit="ft",
+            )
+            session.add(parameter)
+            session.flush()
```
jacob-a-brown, line 907: I feel that an error should be raised rather than add the parameter here

services/well_inventory_csv.py
```diff
@@ -661,6 +739,62 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user)
     )
     session.add(fa)
 
+    if model.measurement_date_time is not None:
```
jacob-a-brown, line 887: A `FieldActivity` record needs to be added where `field_event` is the event already created and `activity_type="groundwater level"`

services/well_inventory_csv.py
```diff
@@ -661,6 +739,62 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user)
...
+            session.flush()
+
+        # create Sample
+        sample_method = (
+            model.sample_method.value
+            if hasattr(model.sample_method, "value")
+            else (model.sample_method or "Unknown")
+        )
+        sample = Sample(
+            field_activity_id=fa.id,
```
jacob-a-brown: this should evaluate to the newly created `"groundwater level"` `FieldActivity` record, not the `"well inventory"` `FieldActivity`

tests/features/steps/well-inventory-csv.py
```diff
@@ -244,11 +244,20 @@ def step_then_the_response_identifies_the_row_and_field_for_each_error(
...
+@then("{count:d} wells are imported")
+@then("{count:d} well is imported")
+def step_then_count_wells_are_imported(context: Context, count: int):
     response_json = context.response.json()
     wells = response_json.get("wells", [])
-    assert len(wells) == 0, "Expected no wells to be imported"
+    validation_errors = response_json.get("validation_errors", [])
+    assert (
+        len(wells) == count
+    ), f"Expected {count} wells to be imported, but got {len(wells)}: {wells}. Errors: {validation_errors}"
```
jacob-a-brown, line 256: If partial steps are not allowed then I don't think that this step will be necessary anymore

tests/test_well_inventory.py
```diff
@@ -19,17 +19,36 @@
...
+        "site_name": "Test Site",
+        "date_time": "2025-02-15T10:30:00",
+        "field_staff": "Test Staff",
+        "utm_easting": 357000,
+        "utm_northing": 3784000,
+        "utm_zone": "13N",
+        "elevation_ft": 5000,
+        "elevation_method": "Global positioning system (GPS)",
+        "measuring_point_height_ft": 3.5,
+    }
```
jacob-a-brown, line 51: Since this is reusable I think that it should be a pytest fixture

tests/test_well_inventory.py
```diff
@@ -437,6 +457,43 @@ def test_well_inventory_db_contents():
...
+    row = _minimal_valid_well_inventory_row()
+    row.update(
+        {
+            "water_level_date_time": "2025-02-15T10:30:00",
+            "depth_to_water_ft": "",
+            "sample_method": "Steel-tape measurement",
+            "data_quality": "Water level accurate to within two hundreths of a foot",
+            "water_level_notes": "Attempted measurement",
+            "mp_height_ft": 2.5,
+        }
```
jacob-a-brown, line 674: This is a style comment, but if you put this dictionary in a variable you access the values for subsequent assert statements. Then if something changes in the future you only need to update the key-value pair and the assert should hold

tests/test_well_inventory.py
```diff
@@ -915,6 +1043,68 @@ def test_group_query_with_multiple_conditions(self):
...
+            {
+                "measuring_person": "Tech 1",
+                "sample_method": "Steel-tape measurement",
+                "water_level_date_time": "2025-02-15T10:30:00",
+                "mp_height_ft": 2.5,
+                "level_status": "Static",
+                "depth_to_water_ft": 11.2,
+                "data_quality": "Water level accurate to within two hundreths of a foot",
+                "water_level_notes": "Initial reading",
+            }
```
jacob-a-brown, line 1364: See above style comment about setting the dictionary to a variable

jacob-a-brown (commented):
Overall I think it's well done and well documented! I like your comments and how the code is organized

alembic/versions/p9c1d2e3f4a5_make_contact_role_nullable.py
jacob-a-brown: After going through the data @ksmuczynski and I determined that there are no edge cases that warrant making this non-nullable. Every `contact_1` and `contact_2` has a defined `role`.

alembic/versions/q0d1e2f3a4b5_make_contact_type_nullable.py
jacob-a-brown: After going through the data @ksmuczynski and I determined that there are no edge cases that warrant making this non-nullable. Every `contact_1` and `contact_2` has a defined `contact_type`.

services/well_inventory_csv.py
jacob-a-brown: role is no longer nullable

services/well_inventory_csv.py
jacob-a-brown: contact type is no longer nullable

services/contact_helper.py
jacob-a-brown: this if statement is used for well inventory purposes since everything gets rolled back on error at the same time, so this comment does not apply

services/thing_helper.py
jacob-a-brown: this if statement is used for well inventory purposes since everything gets rolled back on error at the same time, so this comment does not apply

services/well_inventory_csv.py
jacob-a-brown: addressed in PR #609

services/well_inventory_csv.py
jacob-a-brown: addressed in PR #609

schemas/well_inventory.py
jacob-a-brown: Addressed in PR #610

tests/features/environment.py
```diff
@@ -17,6 +17,11 @@
 import random
 from datetime import datetime, timedelta
 
+# Lock test database before any db module imports
+# Ensures BDD tests only use ocotilloapi_test, never ocotilloapi_dev
+os.environ["POSTGRES_DB"] = "ocotilloapi_test"
+os.environ["POSTGRES_PORT"] = "5432"
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved (commit 525a9fd)

tests/features/steps/well-inventory-csv-validation-error.py
```diff
@@ -253,7 +280,7 @@ def step_then_response_includes_invalid_state_error(context: Context):
 )
 def step_then_response_includes_invalid_well_hole_status_error(context: Context):
     _assert_any_validation_error_contains(
-        context, "Database error", "database error occurred"
+        context, "Database error", "status_history_status_value_fkey"
     )
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved (commit 2932721)

services/well_inventory_csv.py
```diff
@@ -159,69 +161,94 @@ def _import_well_inventory_csv(session: Session, text: str, user: str):
...
                 )
-                session.rollback()
-                wells = []
-            else:
-                session.commit()
+                group = session.scalars(sql).one_or_none()
+                if not group:
+                    group = Group(name=project, group_type="Monitoring Plan")
+                    session.add(group)
+                    session.flush()
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved (commit 1460d4f4). Existing groups are still reused immediately. New groups are now created lazily inside the per-row savepoint. A new group only persists if a row for that project actually imports successfully.

schemas/contact.py
```diff
@@ -151,7 +151,7 @@ class CreateContact(BaseCreateModel, ValidateContact):
     name: str | None = None
     organization: str | None = None
     role: Role
-    contact_type: ContactType = "Primary"
+    contact_type: ContactType
```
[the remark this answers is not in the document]
ksmuczynski: The "Primary" default was not the original behavior. It was changed from required, to "Primary", then rverted back. This can stay as is.

schemas/well_inventory.py
```diff
@@ -122,28 +123,69 @@ def email_validator_function(email_str):
...
+        if v is None or v == "":
+            return None
+        if isinstance(v, enum_cls):
+            return v
+
+        v_str = str(v).strip().lower()
+        for item in enum_cls:
+            if item.value.lower() == v_str:
+                return item
+        return v
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved (commit 3e93bb6).

jirhiker (approved):

PR #610, opened 2026-03-18
BDMS 626: mp height

jacob-a-brown:
…ft for non-null observations

Either of these should be required when a non-null observation is being added to the DB so that dtw bgs can be calculated

### **Why**
- either `measuring_point_height_ft` or `mp_height` should be required if there is a non-null observation
- if bother are provided they should correspond

### **How**
- raise errors if `mp_height != measuring_point_height_ft`
- default to `measuring_point_height_ft`, otherwise use `mp_height`, for `Observation` and `MeasuringPointHistory`
- add numerous unit tests

### **Notes**
- Use bullet points here

### DIFF
--- a/tests/test_well_inventory.py
+++ b/tests/test_well_inventory.py
@@ -530,7 +530,7 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
 
 
 def test_measuring_point_height_ft_used_for_thing_and_observation(tmp_path):
-    """When measuring_point_height_ft is provided it is used for the thing's and observation's measuring_point_height."""
+    """When measuring_point_height_ft is provided it is used for the thing's (MeasuringPointHistory) and observation's measuring_point_height values."""
     row = _minimal_valid_well_inventory_row()
     row.update(
         {
--- a/tests/test_well_inventory.py
+++ b/tests/test_well_inventory.py
@@ -530,7 +530,7 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
 
 
 def test_measuring_point_height_ft_used_for_thing_and_observation(tmp_path):
-    """When both mp_height and measuring_point_height_ft are provided, measuring_point_height_ft should be used for the thing's and observation's measuring_point_height."""
+    """When measuring_point_height_ft is provided it is used for the thing's and observation's measuring_point_height."""
     row = _minimal_valid_well_inventory_row()
     row.update(
         {
--- a/tests/test_well_inventory.py
+++ b/tests/test_well_inventory.py
@@ -562,7 +562,9 @@ def test_measuring_point_height_ft_used_for_thing_and_observation(tmp_path):
         assert observations[0].measuring_point_height == 3.5
 
 
-def test_mp_height_used_for_thing_and_observation_when_mp_height_ft_blank(tmp_path):
+def test_mp_height_used_for_thing_and_observation_when_measuring_point_height_ft_blank(
+    tmp_path,
+):
     """When depth to water is provided but measuring_point_height_ft is blank, the mp_height value should be used for the thing's and observation's measuring_point_height."""
     row = _minimal_valid_well_inventory_row()
     row.update(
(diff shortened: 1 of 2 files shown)


tests/test_well_inventory.py
```diff
@@ -529,6 +529,154 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
...
+def test_measuring_point_height_ft_used_for_thing_and_observation(tmp_path):
+    """When both mp_height and measuring_point_height_ft are provided, measuring_point_height_ft should be used for the thing's and observation's measuring_point_height."""
+    row = _minimal_valid_well_inventory_row()
+    row.update(
+        {
+            "measuring_point_height_ft": 3.5,
+            "water_level_date_time": "2025-02-15T10:30:00",
+            "depth_to_water_ft": "8",
+            "sample_method": "Steel-tape measurement",
+            "data_quality": "Water level accurate to within two hundreths of a foot",
```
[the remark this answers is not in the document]
jacob-a-brown: This is a typo that needs to be addressed in the lexicon

tests/test_well_inventory.py
```diff
@@ -529,6 +529,154 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
         assert observation.sample == sample
 
 
+def test_measuring_point_height_ft_used_for_thing_and_observation(tmp_path):
+    """When both mp_height and measuring_point_height_ft are provided, measuring_point_height_ft should be used for the thing's and observation's measuring_point_height."""
```
ksmuczynski: Thinking of ways to make this read a little clearer, especially for folks who may not be as intimately acquainted with the schema and database fields. Maybe we could clarify the `mp_height` and `measuring_point_height_ft` in the doc string. Something like:

"If values are present in both fields, use the measuring_point_height_ft value (from the MeasuringPointHistory table) instead of the mp_height value (from the Observations table)".

Actually, as I write this, are we sure this what we want? It may not be safe to assume this and instead flag as an error that can be reviewed by AMP...Thoughts?

tests/test_well_inventory.py
```diff
@@ -529,6 +529,154 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
...
+        observations = session.query(Observation).all()
+
+        assert len(things) == 1
+        assert things[0].measuring_point_height == 3.5
+        assert len(observations) == 1
+        assert observations[0].measuring_point_height == 3.5
+
+
+def test_mp_height_used_for_thing_and_observation_when_mp_height_ft_blank(tmp_path):
+    """When depth to water is provided but measuring_point_height_ft is blank, the mp_height value should be used for the thing's and observation's measuring_point_height."""
```
ksmuczynski: Consider clarifying the measuring_point_height_ft and mp_height terms in the doc string.

"When depth to water is provided but measuring_point_height_ft (MeasuringPointHistory)is blank, the mp_height (Observations) value should be used for the thing's and observation's measuring_point_height."

tests/test_well_inventory.py
```diff
@@ -529,6 +529,154 @@ def test_well_inventory_db_contents_with_waterlevels(tmp_path):
...
+        observations = session.query(Observation).all()
+
+        assert len(things) == 1
+        assert things[0].measuring_point_height == 4.0
+        assert len(observations) == 1
+        assert observations[0].measuring_point_height == 4.0
+
+
+def test_null_observation_allows_blank_mp_height(tmp_path):
+    """When depth to water is not provided, a blank measuring_point_height_ft should be allowed and result in a null measuring_point_height on the thing and observation and no associated measuring point height for the well."""
```
ksmuczynski: Similar to previous comments, consider clarifying the measuring_point_height_ft terminology

ksmuczynski: Should we state that a blank mp_height is also allowed in this case?

ksmuczynski (changes requested):
Very minor comments, none of which are required to be addressed. One thought I had was about the BDD tests and feature files - should they be updated to reflect these changes, too? This is the only reason I am requesting changes and not approving.

tests/test_well_inventory.py
jacob-a-brown: This is to test that `measuring_point_height_ft` is used for both the `Observation` and `MeasuringPointHistory`. I updated the doc string to indicate that since it was vague beforehand

ksmuczynski (approved):
APproved!

PR #615, opened 2026-03-20
BDMS-626: mp height fixes

jacob-a-brown:
### **Why**
- `measuring_point_height` for both the `MeasurementPointHistory` and `Observation` tables should be set to the same value for well inventory
- null mp heights should be allowed

### **How**
- remove the requirement for mp heights
- add test to ensure null values are allowed

### **Notes**
- Use bullet points here


ksmuczynski (approved):
Looks good!

PR #633, opened 2026-04-01
NO TICKET fix(importers): prevent duplicate well-name collisions in well inventory and water-level imports

ksmuczynski:
### **Summary**
This PR hardens the CSV importers around duplicate water well names.

It updates the well inventory importer so it:

- preserves the current re-import/idempotency behavior
- reuses an existing well when the row clearly matches a prior import
- fails the row if a `water well` with the same `Thing.name` already exists elsewhere in the database
- avoids silently creating duplicate `Thing` records

It also keeps the water-level importer defensive so ambiguous well lookups no longer crash the CLI and instead return a row-level validation error.

### **Why**

I found that duplicate `Thing.name` records for `water well` could exist in the database, and at least one duplicate appears to have been introduced by the well inventory import path.

That caused two issues:
- the well inventory importer created a second `Thing` with the same well name
- the water-level importer crashed with `MultipleResultsFound` when trying to resolve a well by name

### **How**

- Added a database-wide duplicate-name check in the well inventory import flow after the existing re-import detection logic
- Fail the row with a clear `well_name_point_id` validation error when the well already exists
- Kept the water-level duplicate-well guard so ambiguous lookups fail gracefully instead of aborting the command
- Added regression coverage for both behaviors

### **Why This Approach**
 The well inventory importer still needs to support safe reruns of the same import, so the existing `_find_existing_imported_well()` logic is evaluated first.

Only if the row does not match a prior import do we check whether a well with the same name already exists in the database and block creation.

That gives us:
- idempotent re-imports
- no silent duplicate `Thing` creation
- clearer operator-facing errors

### **Notes**
- This change does not clean up existing duplicate `Thing.name` rows already in the database. Any pre-existing duplicates should be identified and resolved separately.
- This PR intentionally treats well inventory as a create-oriented importer, not a general update/upsert workflow for existing wells. If updating existing wells via CSV is needed, that should be scoped as a separate feature with explicit merge rules.

### DIFF
--- a/services/well_inventory_csv.py
+++ b/services/well_inventory_csv.py
@@ -51,7 +51,7 @@ from schemas.thing import CreateWell
 from schemas.well_inventory import WellInventoryRow
 from services.contact_helper import add_contact
 from services.exceptions_helper import PydanticStyleException
-from services.thing_helper import add_thing
+from services.thing_helper import add_thing, find_water_wells_by_name
 from services.util import transform_srid, convert_ft_to_m
 
 AUTOGEN_DEFAULT_PREFIX = "NM-"
@@ -671,15 +671,8 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user) 
     if existing_well is not None:
         return existing_well.name
 
-    existing_named_well = session.scalars(
-        select(Thing)
-        .where(
-            Thing.name == model.well_name_point_id,
-            Thing.thing_type == "water well",
-        )
-        .order_by(Thing.id.asc())
-    ).first()
-    if existing_named_well is not None:
+    existing_named_wells = find_water_wells_by_name(session, model.well_name_point_id)
+    if existing_named_wells:
         raise ValueError(
             f"Well already exists in database for well_name_point_id '{model.well_name_point_id}'"
         )
--- a/services/well_inventory_csv.py
+++ b/services/well_inventory_csv.py
@@ -280,7 +280,12 @@ def _import_well_inventory_csv(
                             field = "Database error"
                         else:
                             error_text = str(e)
-                            field = _extract_field_from_value_error(error_text)
+                            if error_text.startswith(
+                                "Well already exists in database for well_name_point_id "
+                            ):
+                                field = "well_name_point_id"
+                            else:
+                                field = _extract_field_from_value_error(error_text)
 
                         logging.error(
                             f"Error while importing row {row_number} ('{current_row_id}'): {error_text}"
@@ -665,6 +670,19 @@ def _add_csv_row(session: Session, group: Group, model: WellInventoryRow, user) 
     existing_well = _find_existing_imported_well(session, model)
     if existing_well is not None:
         return existing_well.name
+
+    existing_named_well = session.scalars(
+        select(Thing)
+        .where(
+            Thing.name == model.well_name_point_id,
+            Thing.thing_type == "water well",
+        )
+        .order_by(Thing.id.asc())
+    ).first()
+    if existing_named_well is not None:
+        raise ValueError(
+            f"Well already exists in database for well_name_point_id '{model.well_name_point_id}'"
+        )
 
     # --------------------
     # Location and associated tables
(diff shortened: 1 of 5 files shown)


jirhiker (changes requested):
both of changes seem related to querying the db to check for an existing well. this should be a shared function

ksmuczynski:
There are two related checks here:

1. well-inventory-specific re-import detection via `_find_existing_imported_well()`
2. the generic database lookup for an existing `water well` by `Thing.name`

I think only the second one should be shared.

`_find_existing_imported_well()` is importer-specific idempotency logic. It is intentionally checking for evidence that the row matches a prior well-inventory import, which is different from a general “does a well with this
name already exist?” query.

Example for `_find_existing_imported_well()`:
- a prior well inventory import already created well `TEST-0001`
- that import also created the expected `well inventory` field event on the same inventory datetime, or the expected groundwater-level sample name for the same measurement datetime
- in that case, a rerun of the same inventory row should reuse the existing imported well and succeed

That is different from this case:
- a `Thing` named `TEST-0001` already exists in the database
- but it was created by some other workflow, or the expected inventory lineage does not match
- in that case, the row should not be treated as a safe rerun

The overlap between these changes is the generic `Thing.name == well_name_point_id` and `thing_type == "water well"` lookup. I agree that should be extracted into a shared helper and then each importer can apply its own
behavior to the result set.

Example for the generic lookup in water-level import:
- CSV row references `TEST-0001`
- if 0 wells match, report `Unknown well_name_point_id`
- if 1 well matches, attach the measurement to that well
- if 2+ wells match, report that multiple wells were found and fail the row

Example for the generic lookup in well-inventory import:
- after `_find_existing_imported_well()` does not find a prior import match
- if any `water well` already exists with `Thing.name = TEST-0001`
- fail the row with “well already exists in the database” rather than creating a duplicate `Thing`

So I agree the generic water-well lookup should be shared, but I would keep `_find_existing_imported_well()` separate because it encodes well-inventory-specific rerun behavior rather than just a name-based existence check.

I can make that refactor in a follow-up commit.

ksmuczynski:
@jirhiker Okay, I added a helper `find_water_wells_by_name(...)` in `services/thing_helper.py`. It is reused in the well inventory and water-level CSV importers in the following way:

- water-level importer (`services/water_level_csv.py`): uses it for 0 / 1 / many resolution
- well inventory importer (`services/well_inventory_csv.py`): uses it only for the generic “already exists in DB” check after `_find_existing_imported_well(...)`

So the shared query is centralized, but the two importer-specific behaviors remain separate.

services/well_inventory_csv.py
```diff
@@ -280,7 +280,12 @@ class dialect:
                             field = "Database error"
                         else:
                             error_text = str(e)
-                            field = _extract_field_from_value_error(error_text)
+                            if error_text.startswith(
+                                "Well already exists in database for well_name_point_id "
```
jacob-a-brown, line 284: Since this string is used in multiple places I would recommend creating a global constant at the beginning and then inserting it into f-strings wherever it is used. This way you don't have to worry about having the text correspond between its different invocations and can easily change it in the future. Something like

```python
WELL_EXISTS_SUBSTRING = "Well already in database for well_name_point_id"

...
...
...
if error_text.startswith(WELL_EXISTS_SUBSTRING)
...
...
raise ValueError(f"{WELL_EXISTS_SUBSTRING} `{model.well_name_point_id}`")
```

tests/test_water_level_csv_service.py
```diff
@@ -463,3 +463,65 @@ def test_bulk_upload_water_levels_imports_valid_rows_when_other_rows_fail(
         "Unknown well_name_point_id 'Unknown Well'" in message
         for message in result.payload["validation_errors"]
     )
+
+
+def test_bulk_upload_water_levels_reports_duplicate_well_name_matches():
```
jacob-a-brown, line 468: This test raises what appears to be an omission in the DB. Shouldn't names be unique? That would prevent this error from even occuring. There's even a `TODO` above the name column:

`# TODO: should name be unique?`

jirhiker (approved):

PR #642, opened 2026-04-08
NO TICKET fix(transfers): add scoped PointID transfer mode and rerun guards

ksmuczynski:
### **Summary**
This PR adds an opt-in scoped transfer mode for the legacy ETL pipeline using `TRANSFER_TEST_POINTIDS`.

When `TRANSFER_TEST_POINTIDS` is unset, the existing bulk transfer workflow remains unchanged.

When `TRANSFER_TEST_POINTIDS` is set, the transfer pipeline now scopes work to the requested legacy `PointID` values across applicable transferers instead of pulling unrelated records. This PR also hardens targeted reruns so scoped remediation runs are safer and more predictable.

### **Why**
`TRANSFER_TEST_POINTIDS` did not previously behave as a universal scoped transfer mode.

A targeted run for a single `PointID` could still:

- transfer unrelated non-well `Thing` records
- transfer unrelated assets and other downstream records
- perform global location cleanup
- create duplicate records on rerun for some create-only transfer paths

That made targeted remediation runs unsafe in staging/production and broke the expectation that scoped runs should only touch the requested legacy records.

This PR keeps the default bulk transfer path intact and only changes behavior when `TRANSFER_TEST_POINTIDS` is explicitly set.

That gives us:

- backwards-compatible bulk transfers
- targeted transfer runs for one or more requested legacy `PointID` values
 - clearer operator feedback when requested `PointID` values are missing
- safer reruns through transfer-specific duplicate guards
- faster scoped runs by avoiding global cleanup work

### **How**

#### Scoped transfer orchestration
- Normalized and threaded `TRANSFER_TEST_POINTIDS` through the transfer pipeline
- Added scoped preflight validation so missing requested `PointID` values fail fast before transfer work begins
- Preserved existing bulk transfer behavior when `TRANSFER_TEST_POINTIDS` is not set
- Added scoped geologic formation seeding instead of broad foundational lookup loading during scoped runs

#### Scoped transferer behavior

Updated transferers to honor scoped `PointID` runs where applicable, including:

- well and non-well `Thing` transfers
- assets
- associated data
- chemistry sample info and chemistry child transfers
- hydraulics
- sensors
- NGWMN views
- surface water / weather data and photos
- soil / rock results
- stratigraphy
- water levels and continuous water-level variants
- link IDs
- group associations
- permissions

#### Scoped cleanup behavior

- Scoped location cleanup to only `Location` rows associated with the requested `PointID` values instead of scanning the full table during targeted runs

#### Idempotency hardening for targeted reruns

Added duplicate guards for transfer paths that were still create-only during scoped reruns, including:

- water level field-event lineage
- asset associations
- permission history
- group-to-thing associations
- thing ID links

Also hardened water-level contact reuse so existing contacts are reused instead of causing a transaction-aborting uniqueness failure.

#### Validation

Validated locally and through targeted scoped transfer runs:

- `pytest tests/test_transfer_scoping.py tests/test_thing_transfer.py tests/test_well_transfer.py tests/transfers/test_contact_with_multiple_wells.py -q`
- additional targeted pytest coverage for transfer error handling and scoped helpers
- successful scoped transfer rerun for `SM-0001`
- verified scoped location cleanup against only affected `Location` rows
- verified rerun behavior after adding duplicate guards

### **Notes**
- Bulk transfer behavior is intended to remain unchanged unless `TRANSFER_TEST_POINTIDS` is set.
- The scoped mode was validated against a real targeted transfer scenario (`SM-0001`)

### DIFF
--- a/transfers/asset_transfer.py
+++ b/transfers/asset_transfer.py
@@ -18,7 +18,7 @@ import io
 from sqlalchemy.orm import Session
 from starlette.datastructures import UploadFile
 
-from db import Thing
+from db import Thing, Asset, AssetThingAssociation
 from services.asset_helper import upload_and_associate
 from services.gcs_helper import (
     get_storage_bucket,
@@ -40,7 +40,7 @@ class AssetTransferer(Transferer):
 
     def _get_dfs(self):
         input_df = read_csv(self.source_table)
-        cleaned_df = filter_to_valid_point_ids(input_df)
+        cleaned_df = filter_to_valid_point_ids(input_df, self.pointids)
         return input_df, cleaned_df
 
     def _transfer_hook(self, session: Session):
@@ -55,6 +55,13 @@ class AssetTransferer(Transferer):
                 .filter(Thing.name == row.PointID, Thing.thing_type == "water well")
                 .one_or_none()
             )
+            if well is None:
+                self._capture_error(
+                    row.PointID,
+                    "Thing not found",
+                    "PointID",
+                )
+                continue
             self._asset_step(session, i, well)
             session.commit()
 
@@ -68,6 +75,22 @@ class AssetTransferer(Transferer):
                 return
 
         n = len(photos)
+        existing_asset_names = {
+            name
+            for (name,) in session.query(Asset.name)
+            .join(AssetThingAssociation, AssetThingAssociation.asset_id == Asset.id)
+            .filter(AssetThingAssociation.thing_id == db_item.id)
+            .all()
+            if name
+        }
+        existing_asset_paths = {
+            storage_path
+            for (storage_path,) in session.query(Asset.storage_path)
+            .join(AssetThingAssociation, AssetThingAssociation.asset_id == Asset.id)
+            .filter(AssetThingAssociation.thing_id == db_item.id)
+            .all()
+            if storage_path
+        }
         for j, row in enumerate(photos.itertuples()):
             photo_path = row.OLEPath
             srcblob = self._bucket.get_blob(f"nma-photos/{photo_path}")
@@ -78,6 +101,16 @@ class AssetTransferer(Transferer):
                 continue
 
             head, filename = srcblob.name.split("/")
+            if filename in existing_asset_names or any(
+                storage_path.endswith(filename) for storage_path in existing_asset_paths
+            ):
+                logger.info(
+                    "Skipping existing asset %s for thing.id=%s thing=%s",
+                    filename,
+                    db_item.id,
+                    db_item.name,
+                )
+                continue
             f = srcblob.download_as_bytes()
             ff = UploadFile(file=io.BytesIO(f), filename=filename, size=len(f))
 
@@ -91,8 +124,16 @@ class AssetTransferer(Transferer):
             )
 
             logger.info(
-                f"Added asset {i}-{j}/{n} thing.id={db_item.id} thing={db_item.name} uri: {uri}"
+                "Added asset %s-%s/%s thing.id=%s thing=%s uri: %s",
+                i,
+                j,
+                n,
+                db_item.id,
+                db_item.name,
+                uri,
             )
+            existing_asset_names.add(filename)
+            existing_asset_paths.add(uri[1])
 
 
 # ============= EOF =============================================
--- a/transfers/chemistry_sampleinfo.py
+++ b/transfers/chemistry_sampleinfo.py
@@ -23,7 +23,7 @@ import pandas as pd
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session
 
-from db import NMA_Chemistry_SampleInfo, Location, LocationThingAssociation
+from db import NMA_Chemistry_SampleInfo, Location, LocationThingAssociation, Thing
 from db.engine import session_ctx
 from transfers.logger import logger
 from transfers.transferer import Transferer
@@ -75,8 +75,12 @@ class ChemistrySampleInfoTransferer(Transferer):
                     Location.id == LocationThingAssociation.location_id,
                 )
                 .filter(Location.nma_pk_location.isnot(None))
-                .all()
             )
+            if self.is_scoped_run():
+                results = results.join(
+                    Thing, Thing.id == LocationThingAssociation.thing_id
+                ).filter(Thing.name.in_(self.pointids))
+            results = results.all()
             location_to_thing = {}
             for nma_pk_location, thing_id in results:
                 if nma_pk_location is None:
--- a/transfers/group_transfer.py
+++ b/transfers/group_transfer.py
@@ -47,11 +47,16 @@ class ProjectGroupTransferer(Transferer):
             if prefix:
                 # get all PointIDs that start with prefix
                 sql = select(Thing).where(Thing.name.like(f"{prefix}%"))
+                if self.is_scoped_run():
+                    sql = sql.where(Thing.name.in_(self.pointids))
                 records = session.scalars(sql).unique().all()
                 if records:
                     logger.info(
                         f"Adding {len(records)} things to group {group.name}, prefix {prefix}"
                     )
+                    existing_thing_ids = {
+                        assoc.thing_id for assoc in group.thing_associations
+                    }
                     group_is_monitoring_plan = False
                     for record in records:
                         # set the group_type to Monitoring Plan if at least one well is currently monitored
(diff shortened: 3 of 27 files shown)


jirhiker (changes requested):
Lets change this into a cli function and not touch any of the existing transfer logic. 
This should alway you to greatly simplify the logic as no parallel processing really necessary. 

At this point we should never have to run a bulk wholesale transfer from NM_Aquifer again

transfers/asset_transfer.py
```diff
@@ -68,6 +75,22 @@ def _asset_step(self, session, i, db_item):
                 return
 
         n = len(photos)
+        existing_asset_names = {
+            name
+            for (name,) in session.query(Asset.name)
+            .join(AssetThingAssociation, AssetThingAssociation.asset_id == Asset.id)
+            .filter(AssetThingAssociation.thing_id == db_item.id)
+            .all()
+            if name
```
jacob-a-brown, line 84: This can be updated to `if name is not None` (if that's the intended behavior (would it be if `len(name) > 1?`). There may be some edge cases where name is `""` or `0`, which may be valid, but with falsiness/truthiness `if name` will evaluate to `False`

transfers/asset_transfer.py
```diff
@@ -68,6 +75,22 @@ def _asset_step(self, session, i, db_item):
...
+        }
+        existing_asset_paths = {
+            storage_path
+            for (storage_path,) in session.query(Asset.storage_path)
+            .join(AssetThingAssociation, AssetThingAssociation.asset_id == Asset.id)
+            .filter(AssetThingAssociation.thing_id == db_item.id)
+            .all()
+            if storage_path
```
jacob-a-brown, line 92: Same as above comment: use `if storage_path is not None` (if that's the intended behavior). I think it's better to be explicit, particularly when checking different conditions

transfers/chemistry_sampleinfo.py
```diff
@@ -75,8 +75,12 @@ def _build_thing_id_cache(self):
                     Location.id == LocationThingAssociation.location_id,
                 )
                 .filter(Location.nma_pk_location.isnot(None))
-                .all()
             )
+            if self.is_scoped_run():
+                results = results.join(
+                    Thing, Thing.id == LocationThingAssociation.thing_id
+                ).filter(Thing.name.in_(self.pointids))
```
jacob-a-brown, line 82: I'd recommend using an `if-else` block since `results` may be overwritten if `self.is_scoped_run()` is `True`. It's a bit harder to follow otherwise

```python
if self.is_scored_run():
    results_query = query
else:
    results_query = different query
results = results_query.all()
```

transfers/group_transfer.py
```diff
@@ -47,11 +47,16 @@ def _step(self, session: Session, df: pd.DataFrame, i: int, row: pd.Series):
             if prefix:
                 # get all PointIDs that start with prefix
                 sql = select(Thing).where(Thing.name.like(f"{prefix}%"))
+                if self.is_scoped_run():
+                    sql = sql.where(Thing.name.in_(self.pointids))
```
jacob-a-brown, line 51: Id' recommend using an `if-else` block to make it easier to follow since `sql` may be overwritten if `self.is_scoped_run()` is `True`

```python
if self.is_scoped_run():
    sql = query...
else:
    sql = a different query...
records = session.scalars(sql).unique().all()
```

transfers/group_transfer.py
```diff
@@ -78,9 +83,11 @@ def _step(self, session: Session, df: pd.DataFrame, i: int, row: pd.Series):
...
                                         )
 
-                        gta = GroupThingAssociation(group=group, thing=record)
-                        session.add(gta)
-                        group.thing_associations.append(gta)
+                        if record.id not in existing_thing_ids:
+                            gta = GroupThingAssociation(group=group, thing=record)
+                            session.add(gta)
+                            group.thing_associations.append(gta)
+                            existing_thing_ids.add(record.id)
```
jacob-a-brown, line 90: Should we log that another `GroupThingAssociation` is being created?

transfers/minor_trace_chemistry_transfer.py
```diff
@@ -65,14 +65,15 @@ def __init__(self, *args, batch_size: int = 1000, **kwargs):
...
-                .all()
-            )
+            query = session.query(
+                NMA_Chemistry_SampleInfo.nma_sample_pt_id,
+                NMA_Chemistry_SampleInfo.id,
+            ).filter(NMA_Chemistry_SampleInfo.nma_sample_pt_id.isnot(None))
+            if self.is_scoped_run():
+                query = query.join(
+                    Thing, Thing.id == NMA_Chemistry_SampleInfo.thing_id
+                ).filter(Thing.name.in_(self.pointids))
```
jacob-a-brown, line 75: same comment as above: consider using an `if-else` block here for clarity

transfers/permissions_transfer.py
```diff
@@ -78,16 +96,38 @@ def transfer_permissions(session: Session) -> None:
...
+                    contact.id,
+                    permission.permission_type,
+                )
+                not in existing_permissions
+            ):
                 objs.append(permission)
                 created_count += 1
+                existing_permissions.add(
+                    (well.id, contact.id, permission.permission_type)
+                )
```
jacob-a-brown, line 130: This is repeated code with different variables when calling `_make_permission`. Consider putting it in a for loop for clarity and ease of maintenance:

```python
for permission_tuple in (("MonitorOK", "Water Level Sample"), ("SampleOK", "Water Chemistry Sample")):
    nma_field = permission_tuple[0]
    permission_type = permission_tuple[1]

    permission = _make_permission(wdf, well, contact.id, nma_field, permission_type)
    if (
        permission
        and (
            well.id,
            contact.id,
            permission.permission_type,
         )
        not in existing_permissions
    ):
        objs.append(permission)
        created_count += 1
        existing_permissions.add((well.id, contact.id, permission.permission_type))

**Comment by ksmuczynski** [comment]
I like the idea of a separate CLI path for scoped transfers. It would definitely simplify the operator path for targeted runs.

That said, I don’t think a separate CLI alone would have been sufficient here. The main issues I found were not just entrypoint/orchestration issues, but transferer-level correctness issues. For example:
- some transferers ignored `pointids`
- some child transfers needed parent-based scoping, not just direct `PointID` filtering
- some create-only paths were not idempotent on rerun

So even with a separate CLI, we still would have needed the lower-level transfer logic changes to make scoped runs behave correctly.

What do you think about keeping this PR focused on the underlying transferer correctness fixes, and then opening a follow-up PR to add a dedicated scoped-transfer CLI/entrypoint on top of that? @jirhiker

File: transfers/transfer.py
Comment (jacob-a-brown) [inline transfers/transfer.py:1]: There are a bunch of new functions defined in this file. I'd recommend adding a docstring to each of them to make them easier to understand and use (for now and the future).

File: transfers/transferer.py [hunk 8]
```diff
@@ -45,7 +45,59 @@ def __init__(self, flags: dict = None, pointids: list = None):
...
+        if not text:
+            return None
+        return text.upper()
+
+    @classmethod
+    def _normalize_pointids(cls, pointids: list | None) -> list[str] | None:
+        if not pointids:
+            return None
+        normalized: list[str] = []
+        seen: set[str] = set()
```
jacob-a-brown, line 69: An aside, I really like that you are typing your variables 🔥 . I want to do that whenever possible moving forward, I think it's good practice

transfers/well_transfer_util.py
```diff
@@ -169,11 +170,32 @@ def dump_cached_elevations(lut: dict):
...
+    if normalized_pointids:
+        location_query = (
+            location_query.join(
+                LocationThingAssociation,
+                LocationThingAssociation.location_id == Location.id,
+            )
+            .join(Thing, Thing.id == LocationThingAssociation.thing_id)
+            .filter(Thing.name.in_(normalized_pointids))
+            .distinct()
+        )
```
jacob-a-brown, line 186: like above consider using an `if-else` block here for clarity

jacob-a-brown (commented):
Overall it looks quite good! I mostly made style comments that I think would make the code easier to read and maintain.

jirhiker:
@ksmuczynski i don't want any changes to the transferer logic. Lets consider it deprecated. 

The CLI will need to implement the correct transfer logic for scoped point id transfer

ksmuczynski:
Closed.  Do not build on `kas-transfer-select-pointids-and-data` and do not extend the deprecated transferer execution path in transfers/transfer.py.

A new PR will be opened to implement a new CLI command for targeted legacy PointID transfer. 

Use `kas-transfer-select-pointids-and-data` only as a behavioral reference for scoping rules, rerun/idempotency rules, and test scenarios.

PR #656, opened 2026-04-16
kas-bdms-749-thru-753-refactor-water-level-importer

ksmuczynski:
### **Summary**
 Update the water-level importer to create and reuse structured field-event participants, link imported samples to the measuring participant, and expose water-level staff through existing API payloads instead of relying on
  note text.

### **Why**
 Frontend and reporting consumers need reliable access to water-level measuring staff. Before this change, the importer stored staff and sampler information in notes, which made the data hard to query and easy to misuse. If left unchanged, downstream consumers would continue depending on note text instead of structured participant/contact data.

### **How**
**Backend**
- Create or reuse `FieldEventParticipant` rows for `field_staff`, `field_staff_2`, and `field_staff_3` during water-level import.
- Link imported groundwater-level samples to the measuring participant through `sample.field_event_participant_id`.
- Fail a row with a human-readable error if `measuring_person` does not resolve to exactly one participant for the field event.
  - Keep participant creation idempotent so rerunning the same import does not create duplicate participant rows.
- Stop storing staff/sampler boilerplate in `field_event.notes` and `field_activity.notes`; keep only freeform water-level notes there.

**API**
- Use the existing sample and well-details payloads as the authoritative path for measuring staff:
    - `latest_field_event_sample.contact`
    - `field_event_participants[]`
- Add regression coverage to verify imported water-level staff is exposed through `/thing/water-well/{thing_id}/details`.

**Testing**
- Added/updated importer service tests for:
    - participant creation
    - sample-to-participant linkage
    - rerun/idempotent participant reuse
    - ambiguous `measuring_person` `failure behavior
    - notes now containing only freeform text
- Added API regression coverage for imported water-level staff in well details.
- Local test runs passed:
    - `pytest tests/test_water_level_csv_service.py`
    - `pytest tests/test_thing.py`
    - `pytest tests/test_cli_commands.py`

### **Notes**
- No schema migration was needed.
- Existing API payload shapes were preserved; this change fills structured participant/contact fields and removes the old staff/sampler note boilerplate.

### DIFF
--- a/services/water_level_csv.py
+++ b/services/water_level_csv.py
@@ -537,11 +537,13 @@ def _get_or_create_field_staff_contact(session: Session, staff_name: str) -> Con
     """Resolve or create the contact record used by field event participants."""
     contact_type = "Field Event Participant"
     organization = "NMBGMR"
+    # Contact uniqueness is enforced on (name, organization), so the lookup
+    # must use the same key to avoid missing an existing row with a different
+    # contact_type and attempting a duplicate insert.
     contact = session.scalars(
         select(Contact)
         .where(Contact.name == staff_name)
         .where(Contact.organization == organization)
-        .where(Contact.contact_type == contact_type)
     ).first()
 
     if contact is None:
--- a/services/water_level_csv.py
+++ b/services/water_level_csv.py
@@ -65,9 +65,7 @@ class _ValidatedRow:
     row_index: int
     raw: dict[str, str]
     well: Thing
-    field_staff: str
-    field_staff_2: str | None
-    field_staff_3: str | None
+    field_staff_entries: tuple[tuple[str, str], ...]
     sampler: str
     sample_method_term: str
     field_event_dt: datetime
@@ -302,9 +300,7 @@ def _validate_rows(
                 row_index=idx,
                 raw={**normalized},
                 well=well,
-                field_staff=model.field_staff,
-                field_staff_2=model.field_staff_2,
-                field_staff_3=model.field_staff_3,
+                field_staff_entries=_normalize_field_staff_entries(model),
                 sampler=model.measuring_person,
                 sample_method_term=model.sample_method,
                 field_event_dt=model.field_event_date_time,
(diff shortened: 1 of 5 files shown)


services/water_level_csv.py
```diff
@@ -454,13 +481,111 @@ def _find_existing_imported_sample(
...
+
+def _get_or_create_field_staff_contact(session: Session, staff_name: str) -> Contact:
+    """Resolve or create the contact record used by field event participants."""
+    contact_type = "Field Event Participant"
+    organization = "NMBGMR"
+    contact = session.scalars(
+        select(Contact)
+        .where(Contact.name == staff_name)
+        .where(Contact.organization == organization)
+        .where(Contact.contact_type == contact_type)
```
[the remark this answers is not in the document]
ksmuczynski: Addressed and resolved. I updated the water-level importer to look up contacts by the actual unique key on contact (name + organization) instead of including contact_type, and added regression coverage for reusing an existing same-name/same-organization contact with a different contact_type.

I also noticed the well inventory importer uses the same lookup pattern. I left that out of this PR to keep the scope focused on the water-level importer, but it should get the same fix in a follow-up.

services/water_level_csv.py
```diff
@@ -291,6 +303,8 @@ def _validate_rows(
                 raw={**normalized},
                 well=well,
                 field_staff=model.field_staff,
+                field_staff_2=model.field_staff_2,
```
jirhiker: My only concern is that this is brittle and would not handle more than three participants. Thoughts?

ksmuczynski: That limit is mostly set by the current input csv/schema contract, which only supports three staff columns. That said, I agree the current shape is a bit brittle if we ever expand beyond three participants. It would prob be cleaner to normalize those fields into a list and iterate that, so we’re not hardcoding the slots in the importer. Let me know if you'd like me to implement something like that.

jirhiker: Cool. Yep that's what I was thinking, keep the input form unmodified but normalize staff_1,staff_2 etc into a list. Please implement that then it should be good to merge

ksmuczynski: Done. Re-requested your review.

jirhiker (approved):

PR #667, opened 2026-04-30
 BDMS-764: Add contact name & type to water well list search

TylerAdamMartinez:
### **Why**
- Search functionality was limited to `name_contains`, restricting users to exact/partial name matches only
- Users could not effectively search by other attributes, such as thing type or related contact names
- The existing search did not handle typos or approximate matches (e.g., “Aron” vs “Aaron”)
- This limited usability and made it difficult to locate relevant records

### **How**
- Replaced `name_contains` parameter with a unified `query` parameter across the frontend and backend
- Implemented a combined search strategy in `get_db_things()`:
   - Full-text search using `Thing.search_vector (tsvector)`
   - Fuzzy matching using PostgreSQL `pg_trgm (% operator)`
   - Added similarity-based ranking using similarity() to improve result ordering

### **Notes**
- Requires PostgreSQL `pg_trgm` extension to be enabled for fuzzy search
- Search behavior now depends on both `search_vector` configuration and trigram matching
- Contact search is only applied when `include_contacts=true` to avoid unnecessary joins

### DIFF
--- a/alembic/versions/0493ebb8237d_enable_pg_trgm_extension.py
+++ b/alembic/versions/0493ebb8237d_enable_pg_trgm_extension.py
@@ -1,0 +1,27 @@
+"""enable pg_trgm extension
+
+Revision ID: 0493ebb8237d
+Revises: t6u7v8w9x0y1
+Create Date: 2026-05-01 11:17:44.571959
+
+"""
+
+from typing import Sequence, Union
+
+from sqlalchemy import text
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "0493ebb8237d"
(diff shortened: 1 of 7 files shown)


jacob-a-brown (commented):
Should the `query` be sanitized to prevent malicious behavior?

TylerAdamMartinez:
> Should the `query` be sanitized to prevent malicious behavior?

Yes, good point. I double-checked that there is nowhere in this code change where a user query parameter is placed directly into an SQL statement.

TylerAdamMartinez:
https://github.com/user-attachments/assets/0574c87e-241c-4a24-bcd3-2171fe80e747

ksmuczynski:
Are these changes carried over into the Docker setup, too?

ksmuczynski (approved):

PR #678, opened 2026-05-26
feat(lexicon): add new organization terms in support of the 2026-05-26 well inventory ingestion

ksmuczynski:
### **Why**
- The May 26, 2026 ingestion requires these organizations to exist in the lexicon so incoming records can map cleanly without introducing invalid organization values.

### **How**
Adds three organization terms to core/lexicon.json in support of the May 26, 2026 data ingestion:

  - Santa Ana Pueblo Department of Natural Resources
  - WSP
  - Village of Hope

  These entries did not already exist in the lexicon, so they were added as new organization terms.

### **Notes**
- None

### DIFF
--- a/core/lexicon.json
+++ b/core/lexicon.json
@@ -4568,6 +4568,27 @@
       "categories": [
         "organization"
       ],
+      "term": "Santa Ana Pueblo Department of Natural Resources",
+      "definition": "Santa Ana Pueblo Department of Natural Resources"
+    },
+    {
+      "categories": [
+        "organization"
+      ],
+      "term": "Village of Hope",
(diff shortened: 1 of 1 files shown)


jacob-a-brown (commented):
What does `WSP` stand for? If we want to use that as the `term` then I think that it should be completely spelled out in the `definition` (otherwise I think that the full name should be used for both)

ksmuczynski:
WSP is the name of a global consulting firm. Google says, " "WSP" is simply the corporate brand name and is no longer officially used as an acronym."

jacob-a-brown (approved):

PR #683, opened 2026-06-03
feat: add POST /asset/upload-and-record endpoint for digital asset upload (BDMS-828)

likithabommasani21:
Adds a single endpoint that uploads a digital asset to GCS and records
it in the database in one step. Accepts multipart/form-data with the
file, a required thing_id, and optional label/name fields.

### **Why**
- The existing API required two separate calls to upload a photo:
  `POST /asset/upload` (GCS only) then `POST /asset` (DB only), forcing
  the UI to chain requests and handle intermediate state
- The ticket requires a single endpoint where the back-end handles both
  the GCS upload and database record creation atomically

### **How**
- Added `POST /asset/upload-and-record` to `api/asset.py` with multipart
  form fields: `file` (required), `thing_id` (required), `label` and
  `name` (optional)
- Validates file type before uploading — accepts JPEG, PNG, GIF, WebP,
  TIFF, PDF, and plain text; rejects all others with 400
- Enforces a 250 MB size limit with 400 on violation
- Guards against duplicate uploads — if the same file is re-uploaded for
  the same Thing, the existing asset record is returned instead of
  creating a duplicate
- Creates `Asset` DB record and `AssetThingAssociation` in one
  transaction using existing `gcs_upload`, `check_asset_exists`, and
  `audit_add` helpers
- Fixed a pre-existing bug in `audit_helper.py` where
  `AUTHENTIK_DISABLE_AUTHENTICATION=1` caused a `TypeError` by returning
  `True` instead of a user claims dict

### **Notes**
- The `audit_helper.py` fix affects all endpoints that call `audit_add`,
  not just this one — it is a defensive guard (`isinstance(user, dict)`)
  that is a no-op in production where auth is always enabled
- The existing `/asset/upload` and `POST /asset` two-step endpoints are
  unchanged and still available
- 649 tests passing, 0 failures
- Manually tested via Swagger (`/docs-auth`) — uploaded a JPEG (88,964
  bytes), file was stored in GCS as
  `test_49e37e503db40eeb09e1e4d0f7c76c8c.JPG`, DB record created with
  id=1, uri=https://storage.cloud.google.com/ocotillo/uploads/test_49e37e503db40eeb09e1e4d0f7c76c8c.JPG,
  release_status=draft, storage_service=gcs

### DIFF
--- a/api/asset.py
+++ b/api/asset.py
@@ -17,7 +17,7 @@
 import logging
 import time
 
-from fastapi import APIRouter, Depends, UploadFile, File
+from fastapi import APIRouter, Depends, Form, UploadFile, File
 from fastapi_pagination.ext.sqlalchemy import paginate
 from sqlalchemy import select
 from sqlalchemy.exc import ProgrammingError
@@ -25,6 +25,7 @@ from starlette.concurrency import run_in_threadpool
 from starlette.status import (
     HTTP_201_CREATED,
     HTTP_204_NO_CONTENT,
+    HTTP_400_BAD_REQUEST,
     HTTP_409_CONFLICT,
 )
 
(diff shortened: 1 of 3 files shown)


jirhiker (approved):

jirhiker:
looks great!

PR #701, opened 2026-06-08
fix: rollback and delete GCS blob if asset DB write fails

jirhiker:
### Why
- `POST /asset/upload-and-record` calls `gcs_upload` before `session.commit()`. If the commit (or `session.add`/`session.refresh`) raises, the file stays in the bucket with no `Asset` row pointing at it — same orphan class as the 409 case fixed in [#698](https://github.com/DataIntegrationGroup/OcotilloAPI/pull/698), but on the DB-failure path.

### How
- Wrap the DB write in `try/except` ([api/asset.py:283](api/asset.py:283)).
- On failure: `session.rollback()` + best-effort `bucket.blob(blob_name).delete()` in the threadpool, then re-raise the original exception.
- Blob-delete failures are logged with `logger.warning(..., exc_info=True)` and do not mask the original error.

### Notes
- Best-effort cleanup only — if the GCS delete also fails (e.g. transient outage), the blob stays. Logged so it can be reaped later.
- No new test added — would need to mock both `gcs_upload` and a failing commit.

### DIFF
--- a/api/asset.py
+++ b/api/asset.py
@@ -297,44 +297,51 @@ async def upload_and_record_asset(
         session.add(assoc)
         session.commit()
     except Exception:
-        # Best-effort rollback. If rollback itself raises, log and continue
-        # so the original DB exception still propagates as the request error.
+        # Entire cleanup path is wrapped in one outer try/except so NOTHING
+        # in here (rollback, reference query, bucket.blob(), delete) can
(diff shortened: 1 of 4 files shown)


jirhiker:
Changes migrated into #702. Closing in favor of that PR.

PR #739, opened 2026-06-23
feat(assets): add the ability to disassociate an uploaded asset from a well

TylerAdamMartinez:
### **Why**
- Add an endpoint to update an Asset's Thing association without modifying the Asset itself.
- Allow assets to be reassigned to a different Thing or disassociated entirely.

### **How**
- Add a new PATCH `/asset/{asset_id}/association` endpoint.
- Add `AssetAssociationUpdate` and `AssetAssociationResponse` schemas.
- Add audit logging when creating new associations

### **Notes**
- This endpoint only updates the Asset <-> Thing association; it does not modify Asset metadata.
- Existing associations are removed before creating a new one, ensuring an Asset is associated with at most one Thing.
- Passing `thing_id: null` removes all Thing associations while preserving the Asset record.

### DIFF
--- a/api/asset.py
+++ b/api/asset.py
@@ -19,7 +19,7 @@ import time
 
 from fastapi import APIRouter, Depends, Form, UploadFile, File
 from fastapi_pagination.ext.sqlalchemy import paginate
-from sqlalchemy import select
+from sqlalchemy import delete, select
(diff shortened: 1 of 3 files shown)


likithabommasani21 (approved):
Looks good to me

PR #742, opened 2026-06-25
fix(api/asset): patch delete asset notifications

TylerAdamMartinez:
### **Why**
- Asset deletions were not passing the authenticated user to the deletion helper.

### **How**
- Updated the asset delete endpoint to pass the authenticated user into `model_deleter()`

### **Notes**
- N/A


chasetmartin (approved):
looks good to me

likithabommasani21 (approved):

PR #743, opened 2026-06-26
fix: guard user dict access when auth dependency returns True

jeremyzilar:
## Summary

- Fixes a crash in `crud_helper.py` and `observation_helper.py` where `user["sub"]` was called on a boolean `True` value, raising a `TypeError`
- Adds a regression test to lock in the fix

## Background

When running locally with `AUTHENTIK_DISABLE_AUTHENTICATION=1`, the auth dependency returns `True` instead of a claims dict like `{"name": ..., "sub": ...}`. The existing `if user:` guard passed for `True`, and the code then tried to subscript it (`user["sub"]`), which raised a `TypeError`.

Because this error happened after FastAPI had already started building the response, it bypassed the exception handlers entirely and Uvicorn returned a raw 500 with no CORS headers. This meant the frontend saw a CORS error rather than a real API error, making it very hard to debug.

## Changes

- `services/crud_helper.py`: Changed `if user:` to `if isinstance(user, dict):` in `model_adder`, `model_patcher`, and the `notify_edit_event` block
- `services/observation_helper.py`: Same fix in `observation_model_patcher`
- `tests/test_contact.py`: Regression test that overrides the auth dependency with `True` (matching what local dev does) and confirms `PATCH /contact/:id` returns 200

## Test plan

- [ ] Run `pytest tests/test_contact.py` and confirm the new test passes
- [ ] Confirm existing contact tests still pass

### DIFF
--- a/services/crud_helper.py
+++ b/services/crud_helper.py
@@ -43,7 +43,7 @@ def model_adder(session, table, model, user=None, **kwargs):
     if kwargs:
         md.update(kwargs)
 
-    if user:
+    if isinstance(user, dict):
         # TODO: see note in "AuditMixin"
         md["created_by_id"] = user["sub"]
         md["created_by_name"] = user["name"]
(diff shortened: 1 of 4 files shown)


services/crud_helper.py
chasetmartin: Probably worth updating the other `if user` instances in this file.

chasetmartin: and here?

chasetmartin (approved):
This looks ok to me, just two spots in the crud_helper.py file where the same fix could be useful.

PR #749, opened 2026-06-30
test(ogc): bdms-923 add Sprint 1 cleanup feature spec for OGC

ksmuczynski:
### **Why**
- BDMS-923 calls for a behavior spec capturing Sprint 1 of the OGC feature layer cleanup so expected behavior is documented before implementation lands.
- The OGC layers need a single source of truth for what Sprint 1 covers (exposure filters, naming, configuration) versus what's deferred, so reviewers, testers, and follow-up sprints can plan against the same scope. 
- Helps ensure scenarios are properly categorized for efficient tracking and execution.

### **How**
- Introduced a new BDD feature file capturing the Sprint 1 cleanup scenarios.
- Tagged scenarios with relevant metadata like concern, scope, priority, and action item.

### **Notes**
- Spec-only PR — no step definitions or production code changes yet. Scenarios are intentionally pending until the Sprint 1 implementation work lands. 
- A2 is `@wip`: `terms_of_service` and `contact_name` still need confirming with the technical lead before that scenario can run in CI
- A22 ADR: needs to be written before the two remaining A22 scenarios are meaningful — see comment in the file.
- `other_things` ordering: A1 S3 and all three A13 scenarios include `other_things` because A1 still needs to patch its view; comments in the file flag that these should run before A18 is applied.
- A4 count may drift: 322 is the audit snapshot — if the data changes before step definitions are written, this will need updating.
- A11 role name TBD: the Authentik role for `/ogcapi-internal` access is part of A11 itself and doesn't exist yet; `"public-viewer"` and "required role" references in A11 S2–S4 should be updated once it's provisioned.

### DIFF
--- a/tests/features/ogc-cleanup-sprint1.feature
+++ b/tests/features/ogc-cleanup-sprint1.feature
@@ -1,0 +1,397 @@
+Feature: OGC Feature Layer Cleanup — Sprint 1
+  As an OGC API consumer
+  I want the Ocotillo API feature layers to be accurately filtered, correctly named, and reliably configured
+  So that I can depend on the API for scientific and operational use
+
+  # Sprint 1 scope: A1, A2, A3, A4, A6, A11, A13, A16, A17, A18, A22, A23
+  #
+  # Deferred to Sprint 2:
+  #   A5  — int(None) runtime warning in pygeoapi/api/itemtypes.py
+  #   A7  — Level 1 display title pass (non-breaking naming update)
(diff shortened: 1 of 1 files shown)


chasetmartin (approved):
Looks good!  If you run into issues testing against exact feature or entity counts you may need to adjust a few clauses, but I would just roll with it as you work through them.

TylerAdamMartinez (approved):

PR #750, opened 2026-07-01
feat(api/asset): add new list unassociated assets endpoint

TylerAdamMartinez:
### **Why**
- Adds an endpoint to retrieve assets that are not currently associated with any Thing.
- Supports the upcoming frontend page at `(url)/ocotillo/asset/unassociated`. Enables users to review unassociated assets and decide whether to delete them or reassociate them at a later time.

### **How**
- Added a new `GET /asset/unassociated` endpoint that returns assets without an associated Thing.

### **Notes**
- This endpoint is intended for use by the `(url)/ocotillo/asset/unassociated` frontend list page.

### DIFF
--- a/api/asset.py
+++ b/api/asset.py
@@ -483,6 +483,9 @@ async def list_unassociated_assets(
         .order_by(Asset.id)
     )
 
+    # Signed URLs are generated for thumbnail display on the frontend.
+    # The frontend paginates this endpoint and requests only 10 assets at a time,
+    # which limits GCP IAM calls and keeps signed URL generation manageable.
     def transformer(records: list[Asset]):
         from services.gcs_helper import add_signed_url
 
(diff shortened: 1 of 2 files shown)


api/asset.py
```diff
@@ -468,6 +468,30 @@ def transformer(records: list[Asset]):
...
+    """
+    sql = (
+        select(Asset)
+        .outerjoin(AssetThingAssociation)
+        .where(AssetThingAssociation.asset_id.is_(None))
+        .order_by(Asset.id)
+    )
+
+    def transformer(records: list[Asset]):
+        from services.gcs_helper import add_signed_url
```
chasetmartin, line 490: I would add a code comment given the developer note on line 435 for how this list endpoints deals with signed urls in a reasonable way that does not overload the system or perform unnecessary GCP IAM calls. Or just a note about how the signed urls are for the thumbnails and could be removed later if necessary?

chasetmartin (approved):
looks good, just one note about adding a comment for future changes/dev given the previous documentation about signed urls.

ksmuczynski (approved):

likithabommasani21 (approved):

PR #761, opened 2026-07-06
hotfix: stop per-request OOM on location export endpoints

jirhiker:
## Problem

App Engine terminated the process mid-request: *"the process that handled this request was found to be using too much memory and was terminated."* Already on **F4 (1GB), 1 worker** — so it's a per-request memory spike, not an undersized instance.

Two endpoints loaded the **entire** `SampleLocation` table into RAM on a single request:

- `GET /location/feature_collection` — `.all()` the whole table, then built a **second** list of dicts (double memory).
- `GET /location/shapefile` — `.all()` the whole table **and** wrote `locations.shp/.zip` to CWD. On App Engine the app dir is read-only; only `/tmp` is writable and it's RAM-backed (counts against instance memory).

Both were `async def` doing blocking sync DB/file work, stalling the event loop.

## Fix

**`api/base.py`**
- `feature_collection` → `StreamingResponse`, emits GeoJSON row-by-row via `yield_per(1000)`. Whole table never buffered. Output shape unchanged.
- `shapefile` → writes to `tempfile.mkdtemp()` (not read-only CWD), streams rows with `yield_per`, cleans up the temp dir via `BackgroundTask` after send.
- Both `async def` → `def` (blocking work runs in threadpool).

**`db/__init__.py`**
- Production async engine `echo=True` → `False` (was logging every statement).
- Bounded pools on both engines: `pool_size=5, max_overflow=2, pool_recycle=1800, pool_pre_ping=True`.

## Not in this PR (ops / follow-up)
- `app.yaml` entrypoint updated locally with `--max-requests 200 --max-requests-jitter 40 --timeout 120` (worker recycling as a residual-leak safety net). Not committed — `app.yaml` is untracked and contains secrets.
- ⚠️ `app.yaml` holds a plaintext DB password + GCP service-account private key. Move to Secret Manager and **rotate** those credentials.

## Testing
- Syntax verified. Not load-tested here — verify a `feature_collection` / `shapefile` call against the full table post-deploy.


jirhiker:
Superseded by #762 — clean hotfix cherry-picked onto production (production has a refactored codebase; the original branch targeted the geoserver-iac line and produced a 100-commit diff against production).

PR #764, opened 2026-07-06
feat(api/search): add groups/projects to search

TylerAdamMartinez:
### **Why**
- Adds projects/groups to the global search results.
- Improves search performance for projects by introducing a PostgreSQL full-text search vector.

### **How**
- Added a `search_vector` field to the Group model using `TSVectorType`.
- Added a database migration to create a `search_vector` column for the `group` table, backfill existing records, create a GIN index, and register a trigger to keep the search vector up to date.

### **Notes**
- The PostgreSQL trigger automatically maintains the search vector on future inserts and updates.

### DIFF
--- a/alembic/versions/y3z4a5b6c7d8_add_group_search_vector.py
+++ b/alembic/versions/y3z4a5b6c7d8_add_group_search_vector.py
@@ -12,7 +12,6 @@ import sqlalchemy as sa
 import sqlalchemy_utils
 from alembic import op
 
-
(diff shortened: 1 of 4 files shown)


chasetmartin (approved):
This looks good to me, follows the same patterns as other search indices in the DB. @ksmuczynski would like your confirmation as well.

ksmuczynski (approved):
Nice work, Tyler. This looks great. Approved!

PR #772, opened 2026-07-07
docs: add API monitoring options writeup BDMS-1023 

jirhiker:
Adds `docs/api-monitoring-options.md` — a reference survey of production API monitoring options for OcotilloAPI.

### DIFF
--- a/docs/api-monitoring-options.md
+++ b/docs/api-monitoring-options.md
@@ -1,212 +1,167 @@
-# API Monitoring Options for OcotilloAPI Production
+# API Monitoring & Public Status Page — Options and Recommendation
 
-**Purpose:** Evaluate open-source uptime and health-check monitoring tools for watching production deployments of OcotilloAPI (FastAPI + PostgreSQL/PostGIS), and — added in the 2026-07-07 revision — tools for a **public status page** so users and developers can see live service health.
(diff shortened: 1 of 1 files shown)


jirhiker:
https://ocotillo.openstatus.dev/

chasetmartin (approved):
Great start to monitoring, I think the OpenStatus recommendation is well-reasoned and the probe independence requirement is spot-on and having this status page is great.

One note: the `SELECT 1` health check is a solid addition, and if all workers are down it'll absolutely catch that. The trickier case might be what we were seeing yesterday where some workers were running hot on memory while others were fine. The health check could land on a healthy instance and come back clean, even though users are getting errors from the struggling ones. I'm guessing this health check would eventually see an issue even in this case, but might be worth also checking a heavier common endpoint or query, like thing(well) details.

For future improvement on this as the app grows within the Bureau and the API usage increases: We could implement a common open source monitoring stack such as Prometheus and Grafana. There is a `prometheus-fastapi-instrumentator ` library, and storing the metrics provided + visualizing with Grafana would give us further insight into error rates across all endpoints, latency, etc. We could then use `K6` load testing to formally test for regressions or improvement as needed.

likithabommasani21 (approved):
This looks good overall, nice comparison and clear reasoning for the recommendation. One thought: since Ocotillo  depends on external services like the auth provider and any other services, it might be worth adding those as their own components on the status page too. They're a separate failure domain from the DB/API, so a green health check wouldn't necessarily catch an issue there ,  but it's still something users would notice directly (login not working). Just a thought for completeness, not a blocker.

PR #803, opened 2026-07-29
fix(api/asset): update access from admin to editor

TylerAdamMartinez:
### **Why**
-Asset management endpoints were restricted to admin users.
- Editors also need permission to upload, create, delete, and remove assets.
- Type annotations in the asset API needed minor cleanup for improved accuracy.

### **How**
- Replaced `admin_dependency` with `editor_dependency` for asset upload endpoints.
- Removed the unused `admin_dependency` import.
- Updated the upload documentation to refer to the authenticated editor user.
- Changed the `add_asset` return type to `AssetResponse | None`.
- Changed the optional `thing_id` query parameter type to `int | None`.

### **Notes**
- Editors can now perform asset-management operations that were previously limited to admins.
- Viewer access for listing assets remains unchanged.


chasetmartin (approved):
looks good, I think [this PR](https://github.com/DataIntegrationGroup/OcotilloAPI/pull/802) is the one that will cut the release and deploy, so should be merged before this production code update.

likithabommasani21 (approved):
Looks good to me. 
Not a blocker for this PR - just curious: do we have tests in place that verify editors can upload/create/delete/remove assets, and that viewers cannot?

PR #810, opened 2026-08-06
feat(ogc): ogc A2 replace server metadata placeholders (BDMS-972)

ksmuczynski:
### Summary
Replaces the four `example.com` placeholder values in the OGC server metadata across both the public and internal pygeoapi configs, and adds a `GET /disclaimer` page for `terms_of_service` to point at, so nothing the API advertises is a placeholder or a dead link.

### Why
`core/pygeoapi-config.yml` shipped `terms_of_service: https://example.com/terms`, `identification.url: https://example.com`, `contact.name: API Support`, and `contact.email: support@example.com`. These reach every consumer of the OpenAPI document and the HTML landing page, and `identification.url` is the `rel=about` href on the JSON landing page. They need resolving before the API URL goes to any external partner. `core/pygeoapi-config-internal.yml` carries the same four values, having postdated the audit, which names only the public file.

### Changes
- **Server metadata**: both configs get `identification.url: https://ocotillo.newmexicowaterdata.org`, `contact.name: Ocotillo Support, NMBGMR`, `contact.email: ocotillo-nmbg@nmt.edu`, and a `terms_of_service` pointing at the new disclaimer page. `provider.url` needed no change, since it was already `https://geoinfo.nmt.edu`. The second `example.com` value the audit attributes to the provider URL is actually `identification.url`, a different field.
- **Disclaimer page**: new `GET /disclaimer` (HTML by default, JSON on `?f=json`), text held as constants in `core/disclaimer.py`. `terms_of_service` has to resolve to something and no NMBGMR disclaimer page exists: the `geoinfo.nmt.edu` disclaimer path 404s and the Ocotillo site has no legal page. Rather than block on publishing one in the UI repo, it is served from here, versioned in git and resolving in every environment. Inlining the text in `terms_of_service` instead would make OpenAPI `info.termsOfService` a non-URI. Unauthenticated by design, since an OGC client following the advertised link has no credentials to present.
- **`terms_of_service` URL derivation**: computed from `PYGEOAPI_SERVER_URL` by stripping the mount path, rather than a new env var. That variable is already set in `app.template.yaml` and all three CD workflows; a second base-URL variable would be a fourth place to get a deploy wrong.
- **OpenAPI contact completeness**: adds `provider.email`. pygeoapi builds `info.contact` from `metadata.provider`, not `metadata.contact` (`gen_contact`), so `info.contact.email` was empty regardless of what `contact.email` said.
- **`contact.role` deliberately omitted**: pygeoapi 0.23.5 writes `contact.role` into `x-ogc-serviceContact.hoursOfService` (that line in `gen_contact` is a copy-paste of the `hours` branch above it), so setting it publishes the role string as the service's hours of operation. Caught against a live server, not by the test suite. Relevant to A7.
- **A2 behave scenarios**: the two `@A2` scenarios have existed since the feature file was written but had no step definitions and carried no tag CI filters on, so nothing ran them. Steps added, `@wip` dropped, `@production` added. The second scenario is retargeted from the landing page to the OpenAPI document, because in pygeoapi 0.23.5 the JSON landing page returns only title, description and links, so the provider and contact values it named could never have been asserted there. `terms_of_service` is checked by resolving the URL and looking for the disclaimer text, not by matching a table value: the value is environment-dependent, and a 404 is no better than a placeholder.
- **pytest coverage**: `tests/test_disclaimer.py` for the route, plus assertions in `tests/test_ogc.py` that no OGC surface contains `example.com` and that the OAS `info` contact and terms values are correct. `test_ogc_landing` previously asserted only that `title` was truthy.

### Verification
- Full `pytest` suite: 736 passed, 81 skipped, 6 xpassed.
- `behave --tags=@A2`: 2/2 scenarios, 8/8 steps passing.
- `behave --tags="@backend and @production and not @skip" --dry-run` confirms both A2 scenarios are now selected by the CI filter.
- Live server on `:8011`: zero `example.com` hits across `/ogcapi?f=json`, `?f=html`, `/ogcapi/openapi`, `/ogcapi/collections`; `info.contact` and `info.termsOfService` populated; `/disclaimer` returns 200; `/ogcapi-internal/openapi` advertises the same disclaimer URL.
- Both config templates re-rendered and YAML-parsed after the edits, to confirm no `str.format` brace breakage.

### Notes
- **`run_bdd.sh` clobbers `tests/features/`.** The script still rsyncs `../OcotilloBDD/features/backend/` over the local feature files, cloning the repo if absent. Until that is removed as part of the OcotilloBDD deprecation, running it reverts this PR's `ogc-cleanup-sprint1.feature` changes with no warning. CI is unaffected: `tests.yml` invokes `behave` directly.
- A3 (README example URLs) and A23 (spatial extent) also touch this config and remain open. A23 in particular is a 30-site edit across three files, including two hardcoded literals in `core/pygeoapi.py`, not the single-file change the audit implies.

### DIFF
--- a/api/disclaimer.py
+++ b/api/disclaimer.py
@@ -1,0 +1,113 @@
+# ===============================================================================
+# Copyright 2026
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ===============================================================================
+"""Public data disclaimer page.
+
+Both pygeoapi mounts advertise this URL as
+`metadata.identification.terms_of_service`, so it is deliberately
+unauthenticated -- an OGC client following the advertised link has no
+credentials to present.
+
+HTML is the default because the pygeoapi landing page renders
+terms_of_service as a link a human clicks; JSON is offered for catalog
+harvesters that want the text as data rather than markup.
+"""
+
+import html
+from typing import Annotated
+
+from fastapi import APIRouter, Query, Request
+from fastapi.responses import HTMLResponse, JSONResponse
+
+from core.disclaimer import (
+    DISCLAIMER_CONTACT_EMAIL,
+    DISCLAIMER_PARAGRAPHS,
+    DISCLAIMER_TITLE,
+)
+
+router = APIRouter(tags=["disclaimer"])
+
+_STYLE = (
+    "max-width:44rem;margin:3rem auto;padding:0 1.25rem;"
+    "font-family:system-ui,-apple-system,'Segoe UI',sans-serif;"
+    "line-height:1.6;color:#1a1a1a"
+)
+
+
+def _wants_json(request: Request, f: str | None) -> bool:
+    # An explicit ?f= wins over content negotiation, matching pygeoapi's own
+    # precedence so the two surfaces behave the same way.
+    if f is not None:
+        return f.lower() == "json"
+    accept = request.headers.get("accept", "")
+    return "application/json" in accept and "text/html" not in accept
+
+
+def _render_html() -> str:
+    paragraphs = []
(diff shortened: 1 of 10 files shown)


jirhiker (approved):

jirhiker (changes requested):
Fix pr title, and failing BDD tests. then should be g2g

ksmuczynski:
@jirhiker Failing BDD was a regression on staging after merging A1 and A11. I created a separate PR #812 to resolve the issue and stacked this A2 work on top.

jirhiker (approved):

PR #822, opened 2026-08-11
feat(well-inventory): allow wells outside New Mexico (BDMS-1109)

ksmuczynski:
### **Summary**
Drops the New Mexico-only restriction on well inventory CSV imports and expands supported UTM zones from just 12N/13N to the full continental US range (10N–19N, NAD83). Along the way I also fixed two bugs I ran into: a lowercase zone value like "13n" was slipping past schema validation but blowing up later at persist time, and the location GeoJSON endpoint was hardcoding UTM zone "13N" no matter where a well actually was.

### **Why**
`schemas/well_inventory.py`'s `validate_model` had two hardcoded, coupled checks: an allowlist of exactly `{"12N", "13N"}`, and a lat/lon bounding box matching New Mexico's borders. Anything outside either one got rejected at validation. NMBGMR needs to bring in wells from neighboring states now, and this check was really just an input gate — nothing downstream actually cares whether the well is in NM. `transform_srid` is just generic pyproj, `Location.county`/`.state` are nullable free text pulled from Census TIGERWeb rather than an NM-specific lookup, and there's no `CheckConstraint` on the geometry column enforcing this either. `domain/wells.py` even has a `# TODO: this needs to be more sophisticated in the future. Likely more than 13N and 12N will be used` sitting in it, which is basically calling out the exact limitation this PR removes.
 
I also found a pre-existing bug in the same function: `validate_model` upper-cases `utm_zone` into a local variable but never actually writes it back to `self.utm_zone`. So a row with `"13n"` would pass schema validation fine, then fail case-sensitively down in `services/well_inventory_csv.py` at persist time with a not-very-helpful "Unsupported UTM zone: 13n" error. That's unrelated to the NM restriction, so it's split into its own commit rather than bundled with the feature change.
 
A second bug turned up during review as well: `schemas/location.py`'s `GeoJSONUTMCoordinates` hardcodes `utm_zone = "13N"` and always reprojects the stored point to EPSG 26913, regardless of where the point actually is. This was already wrong for 12N wells out in western NM, but it was a quiet bug since it never really surfaced. It stops being quiet the moment this PR starts allowing wells outside 12N/13N — those wells would import fine and then read back with a confidently wrong easting/northing labeled "13N." Same root cause as the import restriction (assuming every well is in NM), so I fixed it here rather than punting it to a follow-up — shipping the import-side fix without this would mean new imports go in fine but come back wrong.
 
Worldwide support came up too, but we're deferring it on purpose: UTM isn't one global grid, so supporting every zone means picking a datum per hemisphere (WGS84 32601–32660/32701–32760) instead of just reusing NAD83, and there's no real need for that outside North America right now.

### **Changes**
- **Casing bug fix**: added a `field_validator` that strips and upper-cases `utm_zone` once, so every downstream consumer actually sees the canonical value the schema already validated (instead of validating one thing and persisting another).
- **Widened zone support in the domain layer**: `domain/wells.py`'s fixed `{"13N": 26913, "12N": 26912}` dict is gone, replaced by `utm_zone_number()`, a regex parser that accepts any northern-hemisphere zone from 10N-19N. `srid_for_utm_zone()` now just derives the EPSG code arithmetically as `SRID_NAD83_UTM_BASE + zone`, since NAD83 covers the whole range on one consistent datum — no more per-zone lookup table needed. Added `SRID_NAD83_UTM_BASE` and `UTM_ZONE_MIN`/`MAX` as new constants in `core/constants.py`.
- **Dropped the NM restriction: **: `validate_model` no longer checks against NM's lat/lon box or the old two-zone allowlist. Instead it calls `domain.wells.utm_zone_number` and checks against a generic coordinate sanity range (`COORD_LAT_MIN`/`MAX`, `COORD_LON_MIN`/`MAX` in `core/constants.py`) — a plausibility guard against things like transposed easting/northing or feet-vs-meters entry mistakes, explicitly commented as not being a border check.
- **BDD/fixture updates**: renamed the "not within New Mexico" scenario and its Given step to "outside the expected coordinate range," swapped out the invalid-UTM fixture's coordinate (it decoded to lat 37.58, which is southern Colorado — legal now under the new rule) for one that's genuinely outside the sanity range, and updated the expected error strings in the validation-error step to match the new messages.
- **Stale comment/doc corrected**: the feature file said naive datetimes are assumed to be Mountain Time "because location is restricted to New Mexico" — that justification no longer holds, so it's reworded to describe the Mountain Time assumption as a fixed convention and point submitters outside Mountain Time to the (already-supported) option of including a UTC offset.
- **GeoJSON UTM zone/coordinates fixed**: added d`omain/geospatial.py` with `utm_zone_for_longitude()`, which calculates the correct UTM zone from a point's longitude. The result is capped to the same 10N-19N range the import side supports — if the math ever produced a zone outside that range, the code forces it back to the nearest valid zone rather than letting it through. That cap matters because EPSG defines codes only for UTM zones 1 through 23. Beyond that range, the numbering scheme starts reusing codes for a completely different coordinate system (NAD83 state-plane zones), not UTM. Without the cap, an out-of-range longitude could quietly get labeled with one of those unrelated codes — technically a valid EPSG code, just the wrong system entirely — instead of throwing a clear error. I checked this against the actual EPSG data in our pyproj library before writing the fix, rather than assuming it worked as described. `schemas/location.py` now calculates both the numeric coordinate system code and the zone label directly from the point's location, rather than assuming them.

### **Verification**
- `pytest tests/test_domain_wells.py` `tests/test_well_inventory.py` `tests/test_cli_commands.py` `tests/test_domain_geospatial.py` `tests/test_location.py`: 143 passed, including the new `TestWellInventoryRowUtmValidation` class (lowercase normalization, CONUS-boundary rejection, generic-range rejection, the utm library's own `OutOfRangeError` surfacing correctly as a per-row error rather than an aborted run, and acceptance of an 11N coordinate previously rejected only for being outside NM), , the new `test_domain_geospatial.py` (zone-from-longitude math including the CONUS clamping edge cases), and a new `test_location.py` case proving a zone-11N well's GeoJSON response reports `"11N"` with correct coordinates rather than `"13N"`.
- `behave tests/features/well-inventory-csv.feature`: 45/45 scenarios passing, including the renamed scenario end-to-end through the real CLI path.
- Full pytest suite: 820 passed, 81 skipped, 6 passed, 0 failed (excluding two pre-existing, unrelated collection errors from a missing `openpyxl` dependency in the local venv).

### **Notes**
- One related issue found but deliberately left out of scope, since fixing it means touching shared code well beyond this PR's footprint: `services/util.py`'s `normalize_datetime_to_utc` still assumes timezone-naive CSV timestamps are Mountain Time. The original justification for that ("location is restricted to New Mexico") is gone now, but the conversion logic is shared with the water-level CSV importer too, so I didn't want to change behavior there as a side effect of this PR. Covered for now with the doc/comment update above; the real fix — deriving timezone from the coordinate, or just requiring an explicit offset — deserves its own ticket


jirhiker (commented):
**Amended after clarifying scope with Jake** — the original version of this review argued for expanding UTM support worldwide across the board. That was the wrong frame. The actual constraint is a split:

| layer | scope | code |
|---|---|---|
| Location storage + GeoJSON read path | **anywhere on earth** | `schemas/location.py`, `domain/geospatial.py` |
| AMP water-well CSV ingestion | **CONUS** | `schemas/well_inventory.py`, `domain/wells.py` |

That split resolves most of what I originally flagged, and sharpens the rest. Retracting and strengthening below; the two bug fixes in this PR — normalizing `utm_zone` casing, and unhardcoding `13N` in the GeoJSON response — are still clearly right.

## What I'm retracting

These were flagged as limitations. Under the split they're correct as written, and should stay:

- **`domain/wells.py`'s 10N–19N range.** This is the AMP ingestion path. CONUS is the intended scope. Keep it.
- **The `N`-only zone regex.** No southern-hemisphere zone should ever come through the well-inventory importer. Keep it.
- **`northern = True` in `validate_model`.** Permanently correct on this path, not a latent assumption.
- **NAD83 269xx on the import side.** Right datum for the right region. The 10N–19N range sits comfortably inside EPSG's 1–23 NAD83 UTM series, so the arithmetic is safe *here*.
- **`COORD_LAT_MIN/MAX`, `COORD_LON_MIN/MAX`.** A CONUS-shaped gate on AMP ingestion is exactly what's wanted. See the inline comment for the one change I'd still make (naming and placement, not values).

## What this sharpens

**The clamp in `domain/geospatial.py` is now unambiguously a defect rather than a scope decision.** That module is on the worldwide side — it serves the GeoJSON read path for every stored Location. Clamping a worldwide input to a CONUS range means a Berlin point reads back as `E=4961478.4 N=9352242.0 "19N" NAD83`, HTTP 200, no log line. Details in the inline comment.

**The real bug generator is that both layers share constants.** `UTM_ZONE_MIN`/`UTM_ZONE_MAX` live in `core/constants.py` and are imported by *both* `domain/wells.py` (correctly, as an ingestion policy) and `domain/geospatial.py` (incorrectly, as a projection limit). Two modules sitting adjacent in `domain/`, both doing UTM zone work, with opposite scopes and a shared constant between them — that coupling is precisely how a CONUS bound ended up applied to the worldwide read path. Worth splitting the constants so the boundary is visible:

```python
# core/constants.py

# AMP water-well ingestion policy: submissions are limited to the
# continental US. Not a projection limit -- see domain/geospatial.py,
# which serves points anywhere on earth.
AMP_UTM_ZONE_MIN = 10
AMP_UTM_ZONE_MAX = 19
AMP_COORD_LAT_MIN, AMP_COORD_LAT_MAX = 18.0, 72.0
AMP_COORD_LON_MIN, AMP_COORD_LON_MAX = -180.0, -66.0
```

Then `domain/geospatial.py` imports none of them, and the next person can't reach for the wrong bound by accident.

## What the worldwide read path needs

`domain/geospatial.py` has to handle any point we can store. Concretely:

**Use WGS 84 UTM, not NAD83.** `32600 + zone` north, `32700 + zone` south, zones 1–60. NAD83 is a North American datum — a NAD83 UTM lookup at Berlin returns **zero** matching CRS from pyproj, and the 269xx series has real holes: 26924–26928 don't exist, and 26929 is NAD83 / Alabama East (state plane). The WGS 84 series has no holes across 1–60, which means the clamp deletes outright rather than needing a replacement guard.

Datum choice costs nothing numerically — measured the same easting/northing through 26913 vs. the `utm` library (WGS84 ellipsoid) at **0.17 mm** apart, since pyproj applies a ballpark NAD83↔WGS84 transform with no grid shift. NAD83 stays where it belongs, on the AMP import side.

**Take the point, not just the longitude.** Hemisphere can't be derived from longitude, and UTM is undefined outside 80°S–84°N.

**Would pyproj be more robust?** On this side, yes — and more than I originally credited. Not for the arithmetic: `(lon + 180) // 6 + 1` is exact and matches EPSG, and the Norway/Svalbard exceptions don't apply (checked — pyproj returns `32631` for lon 5° / lat 60°, same as plain arithmetic, because the 32V/31X widening is an MGRS grid-zone convention, not an EPSG CRS definition). What it buys is:

1. **The EPSG code is looked up, never invented** — no possibility of landing on a real-but-unrelated CRS.
2. **Out-of-domain input fails loudly for free.** `query_utm_crs_info(datum_name="WGS 84", area_of_interest=AreaOfInterest(0, 85, 0, 85))` returns `[]`. On a worldwide read path that receives arbitrary stored points, that's the difference between a clear error and a plausible wrong number.
3. **Hemisphere derived from the point.** `(-58.4, -34.6)` → `32721`; `(151.2, -33.9)` → `32756`.

~33 ms per uncached query (measured 200 at 6.5 s), so memoize on `(zone, northern)` — not raw lon/lat, or every point misses.

```python
UTM_LAT_MIN, UTM_LAT_MAX = -80.0, 84.0


class OutsideUtmDomain(ValueError):
    """Point falls outside the latitude band UTM is defined for."""


def utm_zone_for_longitude(longitude: float) -> int:
    """Zone number 1-60, with longitude normalized onto [-180, 180)."""
    return int(((longitude + 180) % 360) // 6) + 1


@lru_cache(maxsize=128)
def _utm_epsg(zone: int, northern: bool) -> int:
    # Resolve against the EPSG database rather than computing a code, so an
    # undefined zone raises instead of landing on an unrelated CRS.
    lon = (zone - 1) * 6 - 177  # zone central meridian
    lat = 1.0 if northern else -1.0
    suffix = f"{zone}{'N' if northern else 'S'}"
    for info in query_utm_crs_info(
        datum_name="WGS 84",
        area_of_interest=AreaOfInterest(lon, lat, lon, lat),
    ):
        if info.name.endswith(suffix):
            return int(info.code)
    raise OutsideUtmDomain(f"No WGS 84 UTM CRS for zone {suffix}")


def utm_crs_for_point(longitude: float, latitude: float) -> tuple[int, str]:
    """Return the (EPSG code, zone label) of the UTM zone containing a point."""
    if not UTM_LAT_MIN <= latitude <= UTM_LAT_MAX:
        raise OutsideUtmDomain(
            f"Latitude {latitude} is outside the UTM domain "
            f"({UTM_LAT_MIN} to {UTM_LAT_MAX}); polar points need UPS."
        )
    zone = utm_zone_for_longitude(longitude)
    northern = latitude >= 0
    return _utm_epsg(zone, northern), f"{zone}{'N' if northern else 'S'}"
```

Note the zone math also needs `% 360` independent of everything else — `int((longitude + 180) // 6) + 1` returns **61** at `longitude == 180.0`, and 62 / −2 for unnormalized input like `185` / `-190`. The clamp masks all of it today, so removing the clamp surfaces a crash without this.

## Antarctica

We have no Antarctic data yet but expect to store Antarctic samples. Under the split this lands squarely on the worldwide read path, and not at all on the AMP importer — those wells are CONUS by policy, so nothing there needs to change.

Two bands:

| band | status | sites |
|---|---|---|
| 60°S – 80°S | UTM valid; southern zones exist (`32720`, `32758`, `32712`, `32748`) | Palmer Station, McMurdo, WAIS Divide, Vostok, Dome C |
| below 80°S | **UTM undefined.** `utm.from_latlon` raises `OutOfRangeError`; pyproj returns no UTM CRS | South Pole, Ross Ice Shelf interior, most of the plateau |

Below 80°S needs UPS South (EPSG `32761`) or, more usefully for polar science, Antarctic Polar Stereographic (EPSG `3031`). Fine to defer the implementation — but the read path should *raise* on those points rather than clamp, so the gap is visible the day the data arrives instead of six months later.

## Suggested scope for this PR

Small, in scope, and they turn the eventual polar work into a normal ticket rather than a breaking change:

1. **Replace the clamp with a raise**, and rename the shared constants so the CONUS bound can't be reached from the worldwide side.
2. **Add an explicit `-80.0 / 84.0` latitude gate** in `domain/geospatial.py`, with a comment naming EPSG 3031/32761 as the future path.
3. **Make `utm_coordinates` optional in the GeoJSON response.** Turning a required response object optional *later*, after clients depend on it, is a breaking API change. Doing it now — while every stored point is CONUS and the field is populated on every response anyway — is invisible to every consumer.

The full worldwide WGS 84 UTM switch can be its own ticket. Items 1–3 are what keep this PR from making that ticket harder.

domain/geospatial.py
```diff
@@ -0,0 +1,49 @@
...
+
+    UTM zones are 6 degrees wide starting at -180 (zone 1 covers -180..-174).
+    Clamping keeps the result inside the range EPSG defines as "NAD83 / UTM
+    zone nN" (n = 1..23); above that, 269xx codes belong to unrelated NAD83
+    state-plane systems, so an uncapped zone number could resolve to a valid
+    but wrong CRS instead of failing loudly. Worldwide support is out of
+    scope -- see domain/wells.py.
+    """
+    zone = int((longitude + 180) // 6) + 1
+    return max(UTM_ZONE_MIN, min(UTM_ZONE_MAX, zone))
```
jirhiker, line 41: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. This module is on the worldwide side, which makes the clamp a defect rather than a scope decision.

The hardcoded `13N` was wrong-by-one-zone within New Mexico. This is wrong-by-a-continent, and equally silent. Ran the PR's read path on points outside 10N–19N:

| point | GeoJSON `utm_coordinates` |
|---|---|
| Berlin (13.4, 52.5) | `E=4961478.4  N=9352242.0  "19N"  NAD83` |
| Tokyo (139.7, 35.7) | `E=-2128591.4  N=15641211.8  "19N"  NAD83` |
| McMurdo (166.7, -77.8) | `E=-623450.0  N=-10771051.7  "19N"  NAD83` |

Negative eastings, northings past the pole distance, HTTP 200, no log line. Every one of those is a point Ocotillo is supposed to be able to store.

The docstring frames clamping as avoiding "a valid but wrong CRS instead of failing loudly" — but the uncapped version doesn't silently succeed either. `CRS.from_epsg(26924)` raises `crs not found` (26924–26928 are simply absent; the state-plane collision only starts at 26929). So the choice was never "clamp or silent garbage" — it was "clamp, which *is* silent garbage" versus "raise, which is loud."

Two ways out, either fine:

- **Raise** an `OutsideUtmDomain` here for anything outside the supported range. Cheapest, and makes the gap visible.
- **Switch this module to WGS 84 UTM** (`32600/32700 + zone`), which has no holes across zones 1–60 — the clamp then deletes outright and only a latitude guard is needed.

Worth noting the CONUS bound arrives here via `UTM_ZONE_MIN`/`UTM_ZONE_MAX`, imported from `core/constants.py` — the same constants `domain/wells.py` uses correctly as an AMP ingestion policy. Sharing them across that boundary is what let a CONUS limit land on the worldwide path.

domain/geospatial.py
```diff
@@ -0,0 +1,49 @@
...
+``domain/wells.py`` parses a UTM zone label supplied on import. This module is
+the read-side counterpart: given a point already stored as WGS84, find the UTM
+zone it actually falls in, so the GeoJSON response can report real coordinates
+instead of a fixed zone. See ``schemas/location.py``.
+"""
+
+from core.constants import SRID_NAD83_UTM_BASE, UTM_ZONE_MAX, UTM_ZONE_MIN
+
+
+def utm_zone_for_longitude(longitude: float) -> int:
```
jirhiker, line 28: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. Taking longitude only forecloses two things this path now needs, since it has to serve any stored point:

- **Hemisphere.** N vs S can't be derived from longitude, so this can never label a southern-hemisphere point. (The mirror assumption in `schemas/well_inventory.py` is fine — that path is CONUS by policy. This one isn't.)
- **Latitude bounds.** UTM is undefined outside 80°S–84°N, and nothing in the codebase checks that today.

Suggest `(longitude, latitude)` — the call site in `schemas/location.py` already has the full point in hand, so it costs one argument and lets the polar branch slot in later without touching callers.

Separate from all of the above, the zone math has an edge case the clamp is currently hiding: `int((longitude + 180) // 6) + 1` returns **61** at `longitude == 180.0`, and unnormalized input (`185`, `-190`) yields 62 and −2. So removing the clamp surfaces a crash unless this changes too. `int(((longitude + 180) % 360) // 6) + 1` holds 1–60 across the range I tested (−190 … 185).

domain/geospatial.py
```diff
@@ -0,0 +1,49 @@
...
+    but wrong CRS instead of failing loudly. Worldwide support is out of
+    scope -- see domain/wells.py.
+    """
+    zone = int((longitude + 180) // 6) + 1
+    return max(UTM_ZONE_MIN, min(UTM_ZONE_MAX, zone))
+
+
+def srid_for_longitude(longitude: float) -> int:
+    """Return the NAD83 UTM EPSG code for the zone a longitude falls in."""
+    return SRID_NAD83_UTM_BASE + utm_zone_for_longitude(longitude)
```
jirhiker, line 46: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. NAD83 is the wrong datum for this module specifically. It's correct in `domain/wells.py` — AMP wells are CONUS, and 10N–19N sits comfortably inside EPSG's 1–23 NAD83 UTM series. But this path serves points anywhere on earth, and a NAD83 UTM lookup outside North America returns **zero** matching CRS from pyproj.

`SRID_NAD83_UTM_BASE + zone` also computes a code rather than looking one up, which is the root reason the clamp above has to exist: 26924–26928 don't exist and 26929+ are state plane, so unguarded arithmetic can land on a real-but-unrelated CRS.

WGS 84 UTM (`32600 + zone` north, `32700 + zone` south) has no holes across 1–60, so switching this module removes both problems at once. The datum change is numerically free — measured 0.17 mm difference through 26913 vs. the WGS84-ellipsoid `utm` library, since pyproj applies a ballpark NAD83↔WGS84 transform with no grid shift.

core/constants.py
```diff
@@ -18,6 +18,17 @@
...
+# EPSG 269xx == NAD83 / UTM zone xxN. Zones 10N-19N span the continental US.
+SRID_NAD83_UTM_BASE = 26900
+UTM_ZONE_MIN = 10
+UTM_ZONE_MAX = 19
+
+# A coarse sanity range, not a national border: it catches transposed
+# easting/northing and feet-vs-meters entry mistakes. It is wider than the US
+# (e.g. it admits Mexico City) -- do not use it to decide "is this the US".
+COORD_LAT_MIN, COORD_LAT_MAX = 18.0, 72.0
+COORD_LON_MIN, COORD_LON_MAX = -180.0, -66.0
```
jirhiker, line 30: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. **Retracting my earlier comment here** — I originally flagged this range as blocking worldwide support. It isn't. This gates AMP water-well ingestion, which is CONUS by policy, so a CONUS-shaped bound is exactly right. Values stay.

One change I'd still suggest, though: name and place them so they read as ingestion policy rather than as a general-purpose coordinate limit.

```python
# AMP water-well ingestion policy: submissions are limited to the
# continental US. Not a projection limit -- see domain/geospatial.py,
# which serves points anywhere on earth.
AMP_COORD_LAT_MIN, AMP_COORD_LAT_MAX = 18.0, 72.0
AMP_COORD_LON_MIN, AMP_COORD_LON_MAX = -180.0, -66.0
```

Same argument applies more urgently to `UTM_ZONE_MIN`/`UTM_ZONE_MAX` two lines up: those are imported by both `domain/wells.py` (correct — ingestion policy) and `domain/geospatial.py` (incorrect — applied as a projection limit on the worldwide read path). Generic names on a shared constant are what made that mistake available. An `AMP_` prefix, and `domain/geospatial.py` importing none of them, closes it off.

Minor accuracy note while here: the existing comment calls this "a coarse sanity range, not a national border." Accurate in spirit, but it *is* CONUS-shaped — `-180.0 … -66.0` excludes the entire eastern hemisphere. Saying so plainly makes it obvious why the read path must not borrow it.

core/constants.py
```diff
@@ -18,6 +18,17 @@
 SRID_UTM_ZONE_13N = 26913
 SRID_UTM_ZONE_12N = 26912
 
+# EPSG 269xx == NAD83 / UTM zone xxN. Zones 10N-19N span the continental US.
+SRID_NAD83_UTM_BASE = 26900
```
jirhiker, line 22: Small doc accuracy point: the comment says "EPSG 269xx == NAD83 / UTM zone xxN," which holds for 26901–26923 but not past it. 26924–26928 don't exist, and 26929+ are NAD83 state plane (26929 = Alabama East). Worth stating the 1–23 bound here since this constant is what makes the arithmetic possible, and the bound is the reason the range check downstream isn't optional.

schemas/location.py
```diff
@@ -172,14 +173,16 @@ def populate_fields(cls, data: Any) -> Any:
...
+        # with a confidently wrong easting/northing labeled "13N".
+        zone = utm_zone_for_longitude(point_wgs84_wkt.x)
+        point_utm_wkt = transform_srid(
+            point_wgs84_wkt, SRID_WGS84, srid_for_longitude(point_wgs84_wkt.x)
         )
-        data_dict["properties"]["utm_coordinates"]["easting"] = point_utm_zone_13n_wkt.x
-        data_dict["properties"]["utm_coordinates"][
-            "northing"
-        ] = point_utm_zone_13n_wkt.y
+        data_dict["properties"]["utm_coordinates"]["easting"] = point_utm_wkt.x
```
jirhiker, line 183: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. This is the right fix for the hardcoded `13N` — good catch on it being a quiet bug that stops being quiet under this PR.

One ask while this block is already being touched: make `utm_coordinates` optional on `GeoJSONProperties`.

```python
utm_coordinates: GeoJSONUTMCoordinates | None = None
```

Today it's `Field(default_factory=GeoJSONUTMCoordinates)` — always present, never null. Since Ocotillo stores points anywhere on earth, this response has to cope with points that have no UTM coordinates at all: anything below 80°S or above 84°N, where UTM is undefined. The right answer for those is to *omit* the block, not invent a label. Nothing is lost — the `geometry` member already carries WGS84 lon/lat.

The reason to do it in this PR rather than later is purely API compatibility: changing a required response object to optional after clients depend on it is breaking. Right now every stored point is CONUS and this field is populated on every response, so the change is invisible to every consumer. That window closes as soon as the first non-CONUS location lands.

Related: `GeoJSONUTMCoordinates.horizontal_datum` is hardcoded `"NAD83"` (line 92, outside this diff). Correct for everything stored today, but it becomes wrong for any point outside North America, so it should be derived alongside `utm_zone` whenever this is revisited.

tests/test_domain_geospatial.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@pytest.mark.parametrize(
+    "longitude, expected_zone",
+    [
+        (-170.0, 10),  # far west of CONUS -- clamps rather than picking zone 3
+        (170.0, 19),  # far east of CONUS -- clamps rather than picking zone 51
+    ],
+)
+def test_utm_zone_for_longitude_clamps_outside_conus(longitude, expected_zone):
```
jirhiker, line 46: This test pins the clamp as intended behavior, which is what makes the clamp expensive to remove later — the failing test reads as a regression rather than the fix.

If the clamp becomes a raise, this inverts to something like:

```python
@pytest.mark.parametrize("longitude", [-170.0, 170.0])
def test_utm_zone_for_longitude_rejects_outside_conus(longitude):
    with pytest.raises(UnsupportedUtmZone):
        utm_zone_for_longitude(longitude)
```

Worth adding a `180.0` case either way — the current implementation returns zone 61 for it, which only looks fine because the clamp swallows it.

domain/wells.py
```diff
@@ -37,12 +37,7 @@
     r"^(?P<prefix>[A-Z]{2,3})\s*-\s*(?:x{4}|X{4})$", re.IGNORECASE
 )
 
-# TODO: this needs to be more sophisticated in the future. Likely more than 13N
-# and 12N will be used.
-UTM_ZONE_SRIDS = {
-    "13N": SRID_UTM_ZONE_13N,
-    "12N": SRID_UTM_ZONE_12N,
-}
+UTM_ZONE_REGEX = re.compile(r"^\s*(\d{1,2})\s*N\s*$", re.IGNORECASE)
```
jirhiker, line 40: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. **Retracting the main point I made here.** I'd suggested widening this to `[NS]` for southern-hemisphere support. That's wrong for this module — `domain/wells.py` serves AMP water-well ingestion, which is CONUS by policy, so rejecting an `S` suffix outright is correct and should stay. Same for the 10N–19N bound below.

The one thing I'd still keep from the original comment is much smaller: the rejection message. `Unsupported UTM zone: 13S` gives the submitter no signal about whether the problem is the `13` or the `S`. Since the range is now a deliberate policy rather than an implementation limit, it's worth saying so — something like "southern-hemisphere zones are not accepted; AMP well submissions are limited to CONUS zones 10N-19N."

`\d{1,2}` admitting up to 99 is fine given the range check downstream. Not worth changing.

schemas/well_inventory.py
```diff
@@ -375,22 +382,33 @@ def normalize_measurement_date_time(cls, value: datetime | None) -> datetime | N
...
-        if utm_zone_value not in ("12N", "13N"):
-            raise ValueError("Invalid utm zone. Must be one of: 12N, 13N")
-
-        zone = int(utm_zone_value[:-1])
-        northern = True  # only northern hemisphere zones (12N, 13N) are supported
+        # utm_zone_number raises UnsupportedUtmZone (a ValueError) for anything
+        # outside the supported northern-hemisphere CONUS range; Pydantic wraps it
+        # with its own "Value error, " prefix, so no message is composed here.
+        zone = utm_zone_number(self.utm_zone)
+        northern = True  # utm_zone_number rejects anything but an "N" suffix
```
jirhiker, line 399: > **Amended after clarifying scope with Jake.** Ocotillo should store locations anywhere on earth; AMP water-well ingestion should stay CONUS-only. **Softening this one.** `northern = True` is not a latent assumption on this path — AMP ingestion is CONUS by policy, so it's permanently correct. The comment explaining it is accurate. Nothing to change.

The observation about `test_badly_scaled_coordinates_raise_out_of_range_error` still holds, though, as a note on what's actually guaranteeing correctness here. It's true that `utm.to_latlon` rejects an out-of-domain easting, but the library validates **northing range and zone number, never latitude**:

```
utm.to_latlon(500000, 500000, 58, northern=False) -> lat -85.540   no error
utm.to_latlon(500000,      0, 58, northern=False) -> lat -90.018   no error
```

That last return value isn't a latitude. Not reachable through this importer — the CONUS coordinate range gates it well before that — so this is informational rather than a request. Just worth knowing that the bounding box, not the library, is doing the real work, in case the box is ever loosened.

ksmuczynski:
@jirhiker I pushed two follow-up commits addressing the review:

- `cfcfa70e` refactor(constants): scope UTM/coord bounds to AMP ingestion
- `4d3f23fa` fix(location): use pyproj for worldwide UTM zones

| # | Comment | Resolution |
|---|---|---|
| 1 | `domain/geospatial.py:41` — clamp is "wrong-by-a-continent" | Fixed: switched to pyproj WGS84 lookup, no more clamp |
| 2 | `domain/geospatial.py:28` — longitude-only can't derive hemisphere; zone formula breaks at 180.0/185/-190 | Fixed: `utm_crs_for_point(lon, lat)` takes both; zone formula normalized via `% 360` |
| 3 | `domain/geospatial.py:46` — NAD83 is the wrong datum here | Fixed: module no longer touches NAD83 at all |
| 4 | `core/constants.py:30` — rename/reframe as AMP policy, note the range is CONUS-shaped | Fixed: `AMP_COORD_LAT/LON_MIN/MAX`, comment states the CONUS shape plainly |
| 5 | `core/constants.py:22` — EPSG 269xx comment overclaims | Fixed: comment now states the real 1-23 bound |
| 6 | `schemas/location.py:183` — make `utm_coordinates` optional; fix hardcoded `horizontal_datum` | Fixed: field is `Optional`, datum derived dynamically as `"WGS84"` |
| 7 | `tests/test_domain_geospatial.py:46` — clamp test pins wrong behavior; add a 180.0 case | Fixed: rewrote the test file, clamp tests replaced with raise tests, 180.0 covered |
| 8 | `domain/wells.py:40` — improve rejection message wording | Fixed: message now states the AMP CONUS policy explicitly |
| 9 | `schemas/well_inventory.py:399` — informational only, explicitly "nothing to change" | Left as-is, per the review's own note |

The table above also covers all three "suggested scope for this PR" items from the review summary: raise instead of clamp plus rename the shared constants, add the -80°/84° latitude gate with a comment naming the future EPSG path, and make `utm_coordinates` optional.

Full pytest suite: 834 passed, 0 failed. Both directly-affected feature files (`well-inventory-csv.feature`, `well-core-information.feature`) came back green under `behave`.

Antarctic and polar support (EPSG 3031/32761, below -80°) is intentionally deferred. The read path raises there now instead of silently mislabeling, so the gap stays visible until that's implemented.

PR #825, opened 2026-08-13
fix(ogc): publish project areas imported from maps.nmt.edu

jirhiker:
## Problem

The `project_areas` OGC collection is advertised but serves nothing:

```
GET /ogcapi/collections/project_areas/items  ->  numberMatched: 0
```

`cli/project_area_import.py` writes `group.project_area` from the maps.nmt.edu `Water_Resources` layer but never set `release_status`, so every imported row kept `ReleaseMixin`'s `"draft"` default. `ogc_project_areas` (added in `t6u7v8w9x0y1`, re-created by `f4a5b6c7d8e9`) filters on `release_status = 'public'`, so the layer is published, advertised, and empty.

## Changes

One migration for the rows already imported, one importer change so it does not recur:

- **`c8d9e0f1a2b3`** promotes existing groups carrying a `project_area` boundary from `draft` to `public`.
- **The importer** now publishes on create, and promotes a matched group on *every* run rather than only when the boundary changes — a group whose geometry is already current still has to reach the OGC layer.
- The CLI reports a `Published N group(s) to the OGC layer.` count alongside the existing created/updated/skipped tallies.

## Scoping decisions

Two deliberate calls, both easy to reverse if you disagree:

**Only `draft` is promoted.** `private` and `archived` are deliberate curation decisions and survive untouched; `provisional` and `public` need no change. The boundaries are already published on maps.nmt.edu, so publishing the untouched default carries no disclosure risk.

**Rows are selected by `project_area IS NOT NULL`, not by `group_type`.** There is no provenance column on `Group`, and `group_type` is not provenance — a "Geographic Area" may itself be a legacy project — so filtering on it would both miss imported areas and catch rows this change has no claim on. Geometry is the closest true proxy for "came from the importer". Groups without geometry are left alone entirely.

`downgrade()` is a no-op: once applied, a published area is indistinguishable from one that was already public, so reverting would demote rows the migration never touched.

## Verification

Against `ocotilloapi_test` with seeded rows (removed afterwards):

| Row | Before | After |
|---|---|---|
| geometry + draft | `draft` | **`public`** |
| geometry + private | `private` | `private` |
| geometry + archived | `archived` | `archived` |
| geometry + public | `public` | `public` |
| no geometry + draft | `draft` | `draft` |

`ogc_project_areas` row count went 1 → 2. Migration reported `Published 1 project area group(s) to the OGC layer.`

`tests/test_cli_commands.py` + `tests/test_group.py` — 38 passed, 1 skipped. New test covers a curated `private` group keeping its status while still taking the boundary update.

## Notes

Alembic revision `c8d9e0f1a2b3` chains off `b7c8d9e0f1a2` (#824, merged), keeping a single head.

After deploy, the layer should return features:

```
curl -s "https://ocotillo-api-staging.newmexicowaterdata.org/ogcapi/collections/project_areas/items?limit=1&f=json"
```

### DIFF
--- a/alembic/versions/c8d9e0f1a2b3_publish_imported_project_areas.py
+++ b/alembic/versions/c8d9e0f1a2b3_publish_imported_project_areas.py
@@ -1,0 +1,68 @@
+"""publish imported project areas
+
+Promotes every group carrying a project_area boundary from the untouched
+"draft" default to release_status = 'public', so the boundaries imported from
+the maps.nmt.edu Water_Resources layer reach the ogc_project_areas OGC layer.
+
+Why this is needed: cli/project_area_import.py writes group.project_area but
+never set release_status, so imported rows kept ReleaseMixin's "draft" default.
+ogc_project_areas (see t6u7v8w9x0y1, re-created by f4a5b6c7d8e9) filters on
+release_status = 'public', so the collection is published and advertised while
+serving zero features.
+
+Scope is deliberately narrow on two axes:
+
+* Only rows with a non-null project_area. A group without geometry has nothing
+  to contribute to the layer, and its release status is none of this
+  migration's business.
+* Only rows still at "draft". "private" and "archived" are deliberate curation
+  decisions and must survive this migration; "provisional" and "public" need no
+  change. Project areas are boundary polygons already published on
+  maps.nmt.edu, so publishing the untouched default carries no disclosure risk.
+
+group_type is deliberately NOT used to select rows. It is not provenance -- a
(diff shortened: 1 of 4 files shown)


ksmuczynski:
@jirhiker I only read through this quickly, but if the goal here is to change the `release_status` of all group/project records with a geographic boundary from `draft` to `public`, the `data_migrations/migrations/20260714_0001_publish_project_areas.py` from #782  already does this. You (or someone) just needs to run `oco data-migrations run 20260714_0001_publish_project_areas`.

